PR #3, opened 2024-12-26
corral initial implementation

n0w0f:
## Summary by Sourcery

Disable documentation workflow and update project metadata.

Enhancements:
- Update project name, repository URL, and dependencies.

CI:
- Disable documentation workflow.

Documentation:
- Update documentation URLs and site name.

Tests:
- Update Python version for tests.


agents/agent.py
```diff
@@ -0,0 +1,219 @@
...
+
+class ModelName(Enum):
+    CLAUDE_3_SONNET = "anthropic/claude-3-5-sonnet-latest"
+
+
+class Temperature(Enum):
+    ZERO = 0.0
+    LOW = 0.3
+    MEDIUM = 0.7
+    HIGH = 1.0
```
MrtinoRG, line 34: This is a proposal, maybe it is needed to discuss if we want to do it

agents/agent.py
```diff
@@ -0,0 +1,219 @@
...
+
+if TYPE_CHECKING:
+    from corral.base import Tool
+    from corral.evaluate import BenchmarkInterface
+
+litellm.cache = Cache()
+
+
+class ModelName(Enum):
+    CLAUDE_3_SONNET = "anthropic/claude-3-5-sonnet-latest"
```
MrtinoRG, line 27: I am not sure if this can be beneficial in some way (obviously after adding more models)

agents/agent.py
```diff
@@ -0,0 +1,219 @@
...
+                        + "\n"
+                    )
+
+            calls += 1
+            if calls >= self.max_calls:
+                raise Exception("Max calls reached")
+        return None
+
+
+class ConstrainedAgent(ABC, Agent):
```
MrtinoRG, line 133: I think by adding constrained agents can be a nice ablation

agents/react_agent.py
```diff
@@ -0,0 +1,64 @@
...
+        self.model = model
+        self.temperature = temperature
+        self.timeout = timeout
+        self.caching = caching
+        self.max_calls = max_calls
+        self.model_kwargs = model_kwargs
+        self.sys_prompt = REACTSYSTEMPROMPT
+        self.user_prompt = REACTUSERPROMPT
+        self.toolstore = toolstore or ToolStore()
+        self._scratchpad = ""
```
MrtinoRG: This probably can be shorter by using a config class

agents/agent.py
```diff
@@ -0,0 +1,219 @@
...
+                "content": self.user_prompt,
+            },
+        ]
+
+    def _call(self):
+        return completion(
+            model=self.llm["model"],
+            temperature=self.llm["temperature"],
+            messages=self.format_messages(),
+            max_tokens=8192,
```
MrtinoRG, line 82: This can be a problem and something to sell. This is, this is here because it is the Anthropic limit, but if we keep it for all the models, we are really evaluating all the models in the same conditions

agents/agent.py
MrtinoRG: And I think it is interesting because I think it is not clear the influence of the temperature in the agents, i.e. what temperature yields best results

n0w0f:
@MrtinoRG, Could you run the `samplemath` task ?
What do you think about the current initial implementation of base classes ? 

I have now also implemented a decorator `tool` , that will take function and give `Tool` (corral Tool object) which then can be added to `Environment`.

I saw you added an example of function running in modal environment.
could you maybe also try implementing a decorator for tools running in modal environment. It should take function and return `Tool`
I am thinking of having decorators like `tool` (runs in same `task`'s environment ), `modal_tool`	(tool run in modal environment), maybe also a different decorator that would run in container/slurm. What do you think about this ?

MrtinoRG:
> @MrtinoRG, Could you run the `samplemath` task ? What do you think about the current initial implementation of base classes ?

I think we can try to merge what we have now, so we have initial signals, and then we can improve/enhance the code in different PRs and add the different tasks and tools for each of the tasks. But maybe we should discuss it first with Kevin as well.

> I saw you added an example of function running in modal environment. could you maybe also try implementing a decorator for tools running in modal environment. It should take function and return `Tool` I am thinking of having decorators like `tool` (runs in same `task`'s environment ), `modal_tool` (tool run in modal environment), maybe also a different decorator that would run in container/slurm. What do you think about this ?

Sure, what do you think if such decorator takes all the info (i.e. descriptions, args, returns and examples) from the docstring of the function as well instead of need to add them by hand?

agents/llamp_agent.py
MrtinoRG, line 1: The code that they use is through langchain, so I guess the ReAct loop is the classical one:

```
    agent_executor = initialize_agent(
        agent=AgentType.STRUCTURED_CHAT_ZERO_SHOT_REACT_DESCRIPTION,
        tools=tools,
        llm=llm,
        verbose=True,
        max_iterations=5,
        handle_parsing_errors=True,
        callback_manager=BaseCallbackManager(
            handlers=[top_level_cb]),
        memory=conversation_redis_memory,
        agent_kwargs={
            'prefix': PREFIX,
            'suffix': SUFFIX,
        },
    )
    pubsub = redis_client.pubsub()
    pubsub.subscribe(chat_id)

    ainvoke_task = asyncio.create_task(
        agent_executor.ainvoke({"input": input_data}))
```

n0w0f:
> what do you think if such decorator takes all the info (i.e. descriptions, args, returns and examples) from the docstring of the function as well instead of need to add them by hand?


So currently i parse from the doc string and build `ToolArgument` (corral\utils\parse_docstring). This should be reusable.
This parsing idea is inspired from `https://huggingface.co/docs/transformers/main/en/chat_templating`. 

```
@dataclass
class ToolArgument:
    """Specification for a tool argument"""

    name: str
    type: str
    description: str
    required: bool = True
    default: Any = None
    choices: Optional[List[Any]] = None  # from transformers
  ```

**Comment by n0w0f** [comment]
I suggest we keep the directory structure for agents also directory wise
```
agents/
├── baseline
 │   └── baseline.py
├── ReactAgent
├── LlampAgent
```

**Comment by MrtinoRG** [comment]
> I suggest we keep the directory structure for agents also directory wise
> 
> ```
> agents/
> ├── baseline
>  │   └── baseline.py
> ├── ReactAgent
> ├── LlampAgent
> ```

I followed what they have in [ldp](https://github.com/Future-House/ldp/tree/main/ldp/agent)

**Comment by n0w0f** [comment]
@MrtinoRG This PR is getting biggger.
Let us build parallely, I guess you are working on ReactAgent and i will build more complex task. We can iterate and improve then we take review from Kevin ?

File: src/corral/base.py [hunk 6]
```diff
@@ -0,0 +1,273 @@
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from dataclasses import dataclass, field
+from datetime import datetime
+from enum import Enum, StrEnum
+from typing import Any
```
MrtinoRG:
```suggestion
from typing import Any, Optional, List
```

MrtinoRG (approved):

PR #21, opened 2025-01-13
Improved agent logic

chandan21gupta:
Have included following changes : 
1. Agent logic for handling complex tasks involving multiple subtasks.
2. Improved system prompts.
3. Gemini and GPT based llm agents.


n0w0f (commented):
@chandan21gupta,
Is the objective here to try multiple models in the baseline agent setting ?
I suggest we use `litellm`. See the implemetation of react agent [here](https://github.com/lamalab-org/mat-agent-bench/blob/main/src/corral/agents/react.py)

We can have a BaselineAgent that takes model name and then use it like [here](https://github.com/lamalab-org/mat-agent-bench/blob/main/scripts/run_react_agent.py)

If required you can rewqrite https://github.com/lamalab-org/mat-agent-bench/blob/main/agents/baseline.py to be the BaselineAgent

pyproject.toml
```diff
@@ -37,6 +37,7 @@ dependencies = [
   "requests>=2.32.3",
   "loguru>=0.7.3",
   "litellm>=1.56.4",
+  "mistralai"
```
n0w0f, line 40:
```suggestion
  "litellm>=1.56.4",
```
this is not being used right ?

tasks/lammps/lammps/env.py
```diff
@@ -0,0 +1,79 @@
+import uvicorn
+# from tools import  read_lammps_to_string,ase_lammps, pickle_to_lammps,brave_search,wikipedia_search
```
n0w0f, line 2:
```suggestion
```

tasks/lammps/lammps/env.py
```diff
@@ -0,0 +1,79 @@
...
+    def __init__(self, task_id: str, question: str):
+        self.question = question
+        super().__init__(task_id)
+
+        # Add multiple tools
+        for tool in LAMMPS_TOOLS:
+            self.add_tool(tool)
+
+    def get_task_prompt(self) -> str:
+        # return f"Solve this  problem, you can use the path {self._path} to read and write if you want : {self.question}"
```
n0w0f, line 25:
```suggestion
        # return f"Solve this  problem, you can use the path {self._path} to read and write if you want : {self.question}"
```

tasks/lammps/lammps/env.py
```diff
@@ -0,0 +1,79 @@
...
+    # prompt = '''
+
+    # Generate a face-centered cubic (FCC) lattice structure for Aluminum (Al) using the Atomic Simulation Environment (ASE) package in python with a lattice constant of 4.05 angstrom. The simulation should define a 5x5x5 simulation box, set periodic boundary conditions to True, and dump the structure in XYZ format to the specified location. Create a directory /Users/chandan21gupta/Desktop/iit_delhi/agent_llms_3/mat-agent-bench/tasks/lammps/lammps/task_7/ and save the following inside it: 
+
+    # 1. The Python script used for generating the lattice. 
+
+    # 2. The generated XYZ file containing the lattice structure. 
+
+    # ASE library is already installed in the computing environment.
+    # '''
```
n0w0f, line 55:
```suggestion
    # prompt = """

    # Generate a face-centered cubic (FCC) lattice structure for Aluminum (Al) using LAMMPS with a lattice constant of 4.05 angstrom. The simulation should use metal units, define a 5x5x5 simulation box, set periodic boundary conditions to True, and dump the structure in XYZ format to the specified location. Create a directory /Users/chandan21gupta/Desktop/iit_delhi/agent_llms_3/mat-agent-bench/tasks/lammps/lammps/task_2/ and save the following inside it: 

    # 1. The input file used for the simulation. 

    # 2. The output log file generated by LAMMPS. 
    
    # 3. The generated XYZ file containing the lattice structure. 

    # """

    # prompt = '''

    # Generate a face-centered cubic (FCC) lattice structure for Aluminum (Al) using the Atomic Simulation Environment (ASE) package in python with a lattice constant of 4.05 angstrom. The simulation should define a 5x5x5 simulation box, set periodic boundary conditions to True, and dump the structure in XYZ format to the specified location. Create a directory /Users/chandan21gupta/Desktop/iit_delhi/agent_llms_3/mat-agent-bench/tasks/lammps/lammps/task_7/ and save the following inside it: 

    # 1. The Python script used for generating the lattice. 

    # 2. The generated XYZ file containing the lattice structure. 

    # ASE library is already installed in the computing environment.
    # '''
```

tasks/lammps/lammps/tools.py
```diff
@@ -0,0 +1,72 @@
...
+
+    Args:
+        input_file: Path to the LAMMPS input script.
+        log_file: Path to the log file where LAMMPS will write its output.
+    """
+    if not os.path.exists(input_file):
+        return f"Error: Input file not found: {input_file}"
+
+    # Construct the LAMMPS command with the -log option
+    command = f"lmp -in {input_file} -log {log_file}"
```
n0w0f, line 23: This assumes LAMMPS is compiled and could be used in this environment. 
would be nice to have this in the README for this environment.

tasks/lammps_input/lammps_input/env.py
```diff
@@ -12,7 +12,7 @@
 class LammpsEnvironment(Environment):
     def __init__(self, task_id: str, question: str):
         self.question = question
-        self._path = "/Users/n0w0f/git/n0w0f/mat-agent-bench/tasks/lammps_input/temp"
```
n0w0f, line 15: we can remove the hardcoded path here i guess

tasks/lammps_input/pyproject.toml
```diff
@@ -11,6 +11,9 @@ dependencies = [
     "pydantic>=2.10.4",
     "python-dotenv>=1.0.1",
     "uvicorn>=0.34.0",
+    "mistralai==1.2.6",
+    "google-generativeai==0.8.3",
+    "openai=1.59.5"
```
n0w0f, line 16: this would go away as we move to litellm. Either ways agent dependency shouldn't be part of the environment dependency.

n0w0f (changes requested):

agents/gemini_baseline.py
```diff
@@ -0,0 +1,101 @@
+import json
+
+import google.generativeai as genai
+
+from dotenv import load_dotenv
+
+from corral.evaluate import BenchmarkInterface, MatAgentBenchmark
+
+from prompts import BASELINESYSTEMPROMPT, BASELINEUSERPROMPT, BASELINESYSTEMPROMPT2
```
kjappelbaum, line 9: I feel we need to find a better naming convention - `BASELINESYSTEMPROMPTX` will be very hard to maintain and understand

agents/gemini_baseline.py
```diff
@@ -0,0 +1,101 @@
...
+
+import re
+
+load_dotenv("../.env", override=True)
+
+
+class GeminiAgent:
+    """Claude-based agent implementation"""
```
kjappelbaum, line 17:
```suggestion
    """Gemini-based agent implementation"""
```

do we really need a full class for every model if it is "just" prompt-based? Can't we use `litellm`?

agents/gemini_baseline.py
```diff
@@ -0,0 +1,101 @@
...
+    def __init__(self, api_key: str):
+        # self.client = anthropic.Anthropic(api_key=api_key)
+        self.client = None
+
+    def solve_task(self, interface: BenchmarkInterface, task_id: str) -> str:
+        guide = interface.get_task_guide(task_id)
+
+        system_prompt = BASELINESYSTEMPROMPT2.format(guide=guide)
+
+        # print("system_prompt", system_prompt)
```
kjappelbaum, line 28: in lieu of having commented out code I would use logging for debugging and remove unused code completely

kjappelbaum: (this also holds true for all the print statements below)

agents/gemini_baseline.py
```diff
@@ -0,0 +1,101 @@
...
+        messages = [{"role": "user", "parts": BASELINEUSERPROMPT}]
+
+
+        while True:
+
+            
+            response = self.client.generate_content(
+                messages, generation_config = genai.GenerationConfig(
+                                            max_output_tokens=1000,
+                                            temperature=0.1)
```
kjappelbaum, line 44: i would consider moving those important `kwargs` to the constructor of the class as we will probably want to change such things easily without touching the implementation of the classes themselves

agents/openai_baseline.py
kjappelbaum, line 1: same comment as above: there seems a lot of redundancy that we should be able to remove with something like `litellm` (or at least better inheritance)

agents/openai_baseline.py
```diff
@@ -0,0 +1,98 @@
...
+
+        messages = [{"role" : "developer", "content" : system_prompt}]
+
+        messages.append({"role": "user", "content": BASELINEUSERPROMPT})
+
+
+        while True:
+
+            response = self.client.chat.completions.create(
+                model="gpt-4o-mini",
```
kjappelbaum, line 40: this should also not be hard-coded but exposed in the constructor of the class

agents/openai_baseline.py
```diff
@@ -0,0 +1,98 @@
...
+                messages=messages,
+
+
```
kjappelbaum, line 43: I think it is not documented clearly, but we should run `ruff` for formatting and linting

tasks/lammps/lammps/env.py
```diff
@@ -0,0 +1,79 @@
...
+# from tools import  read_lammps_to_string,ase_lammps, pickle_to_lammps,brave_search,wikipedia_search
+from tools import run_lammps, bash_command
+
+
+from corral.base import Environment
+from corral.server import create_benchmark_server
+import ase
+
+
+# LAMMPS_TOOLS = [read_lammps_to_string,ase_lammps, pickle_to_lammps,brave_search,wikipedia_search]
```
kjappelbaum, line 11:
```suggestion
```

tasks/lammps/lammps/env.py
```diff
@@ -0,0 +1,79 @@
...
+
+LAMMPS_TOOLS = [run_lammps, bash_command]
```
kjappelbaum, line 13: it seems that we need to have a registry/store for tools. While we can now define such variables, it would be way cooler to just find the right list of tools in a store given an environment name or some other constraint/search

kjappelbaum: /cc @n0w0f

tasks/lammps/lammps/env.py
```diff
@@ -0,0 +1,79 @@
...
+
+    The EAM potential file is not available, you can download it from https://www.ctcms.nist.gov/potentials/ .
+
+    '''
+
+
+    environments = {
+        "lammps_1" : LammpsEnvironment("input_1", prompt)
+    }
+
```
kjappelbaum, line 75: for such long prompts it might be almost more convenient to store the prompts in some form of external file (e.g. a `json`)

tasks/lammps/lammps/tools.py
kjappelbaum: a nice-to-have in the long term could be that we implement some kind of "healthcheck". That is, we could implement a method for each tool that checks if it is ready to be run and then can fail before any real action is tried to be executed (e.g., when the environment is initialized).

tasks/lammps/lammps/tools.py
```diff
@@ -0,0 +1,72 @@
...
+    try:
+        print(f"Executing LAMMPS command: {command}")
```
kjappelbaum, line 25: we should use some logging framework. As we add more complexity to the system, we want to be able to set up handlers that redirect the output from different subsystems to different files/consoles

kjappelbaum:
what should we do with this PR? Is here still any activity planned, or is this already deprecated?

chandan21gupta:
this PR should be deprecated for now. 

This PR was mainly dedicated to make weaker agents work, but the react based agent works better and serves as a better baseline. I have noted the suggestions made and can incorporate them in later PRs.

PR #35, opened 2025-01-21
Lammps environment

chandan21gupta:
This is a draft PR, as it contains the eval set which others can see and use. Implementations are currently messy and will be updated over time.


kjappelbaum:
could you perhaps push your latest state? I would like to get an overview of where we are

chandan21gupta:
> could you perhaps push your latest state? I would like to get an overview of where we are

Have pushed - but the code is extremely messy. I am currently working to make it more modular.

tasks/lammps/lammps/data/energy_minimization/task_1/task_1.json
```diff
@@ -0,0 +1,97 @@
...
+  "task_prompt": "Your task is to generate a fcc lattice structure for Aluminum with a starting lattice constant 4.05 angstrom using lammps and calculate its cohesive_energy. The simulation should use metal units, define a 5x5x5 simulation box, set periodic boundary conditions to true, use eam potential, save all coordinates and lattice parameters in structure.xyz file and print the final cohesive_energy of the structure after energy_minimisation. Use conjugate_gradient with energy tolerance upto 1.0e-4 and force_tolerance upto 1.0e-6 having maximum iterations as 10000, and maximum evaluations as 10000. All the potentials are saved in the directory /root/potentials, which in turn contains subdirectories by the material names, and each subdirectory contains various potentials for that material. For example, if you want to access potential for Cu, you can access it using /root/potentials/Cu/ which will contain various potentials, exact file name can vary and should be figured out. Save all the related files at /Users/chandan21gupta/Desktop/iit_delhi/agent_llms_3/mat-agent-bench/tasks/lammps/lammps/results/energy_minimization/task_1.",
```
kjappelbaum, line 4: I think this is related to the I/O tools we still have to implement, but we should not have filepaths that do not exist on other systems in the prompt.

kjappelbaum: (this, ofc, also applies to other files)

PR #39, opened 2025-02-21
feat: add pass k metrics

n0w0f:
## Summary by Sourcery

Adds pass@k metrics to the benchmarking process, enabling a more robust evaluation of agent performance by running multiple trials per task and reporting detailed results.

New Features:
- Introduces the calculation of pass@k and pass^k metrics to evaluate the performance of agents across multiple trials, enhancing the robustness of benchmark results.

Enhancements:
- Refactors the benchmarking process to support multiple trials per task, allowing for a more comprehensive evaluation of agent performance.
- Improves the reporting functionality to display detailed benchmark results, including overall metrics and per-task statistics, with an option to save the report as a JSON file.

Tests:
- The changes include new metrics and reporting, but do not include any new tests.


agents/agent.py
```diff
@@ -1,22 +1,19 @@
 import re
-from abc import ABC
+from abc import ABC, abstractmethod
 from enum import Enum
-from typing import TYPE_CHECKING
 
 import litellm
 from litellm import completion
-from litellm.caching import Cache
+from litellm.caching.caching import Cache
```
MrtinoRG:
```suggestion
from litellm.caching import Cache
```

agents/agent.py
MrtinoRG, line 1: Can we remove these files? Like the ones in the agents directory?

src/corral/report.py
MrtinoRG, line 1: I am having a really hard time to understand this chunk of code

src/corral/report.py
```diff
@@ -0,0 +1,240 @@
...
+    tool_statistics: dict[str, Any]  # TODO replace Any with specific types
+
+    @property
+    def success(self) -> bool:
+        """Whether the trial was successful"""
+        return self.score > 0
+
+
+@dataclass
+class TaskTrials:
```
MrtinoRG: The docstring claims that is the collection of trials for a single task. Then why you have to make those loops with k steeping?

MrtinoRG: Like I think the simpler thing would be that this class calculate the metrics for each task, so you loop for range(k) in each of the functions so the code is much simpler

MrtinoRG: Like, I do not understand why you have to do such lists comprehensions if in `BenchmarkResult` you are already looping over all the different tasks. If you do that there, which I think is correct, here it should be as simple as looping over the different trials

MrtinoRG (commented):
We should merge this first because right now the main branch is not working correctly if one tries to run the MathEnvironment

kjappelbaum:
(I planned in some time this afternoon to go over the PRs there - sorry for the delay)

src/corral/report.py
n0w0f: TaskTrials has methods to compute score over trails for a single task.
Benchmark result class compute the score over tasks. 

Are you suggesting to ditch TaskTrials all together and adjust the data structure and methods in Benchmark results ?

agents/agent.py
n0w0f: I agree.

n0w0f: do you think we should add the baseline agent or use the react one as baseline ?

src/corral/evaluate.py
```diff
@@ -63,70 +47,70 @@ def submit_answer(self, task_id: str, answer: str) -> TaskResult:
...
+    def bench(
+        self, task_ids: list[str] | None = None, trials_per_task: int | None = None
+    ) -> BenchmarkResult:
+        """Run benchmark on specified tasks or all available tasks
 
-    def bench(self, task_ids: List[str] | None = None) -> BenchmarkResult:
-        """Run benchmark on specified tasks or all available tasks"""
+        Args:
+            task_ids: list of task IDs to run, or None for all tasks
+            trials_per_task: Number of trials per task, defaults to k if not specified
```
kjappelbaum: why do you name it sometimes `trials_per_task` and sometimes `k` and why can we set it in two different places - what is the benefit of that compared to having it only in one place?

src/corral/report.py
kjappelbaum: We may have too many abstractions here. 

Conceptually, it seems that one run should give us some kind of result (`TaskResult`). Then, if we have a list of those, we can compute more fancy statistics over them: but for the code it does not matter if it is a list of lenght 1 or `k`. That is, I feel we can return the result of a `bench` run always as a collection of `TaskResult` and the class that stores this collection would then compute all the metrics and potentially statistics (mean/std/pass @ k) of the scores of individual runs.

kjappelbaum: that is, I would remove this class here or choose names that are very clear. I see the following abstractions. 

- Result for one task: There should be almost no logic here
- Result for one task being run more than once: This logic computes summary statistics over one task 
- Result for multiple tasks, being run more than once in general: This logic computes summary statistics over multiple tasks

src/corral/report.py
```diff
@@ -0,0 +1,240 @@
...
+
+    trials: list[TaskResult] = field(default_factory=list)
+
+    @property
+    def success_rate(self) -> float:
+        """Calculate success rate across all trials"""
+        return mean(1 if trial.success else 0 for trial in self.trials)
+
+    def calculate_pass_at_k(self, k: int) -> float:
+        """Calculate pass@k - probability of at least one success in k trials"""
```
kjappelbaum:
```suggestion
        """ Calculate pass@k - probability of at least one success in k trials
        
        pass @ k is defined as
        
      	.. math: P(pass@k) = 1 - (1 - p)^k
      	
      	In practice, pass@k is estimated by:

	- Taking n samples for each problem (where n ≥ k)
	- Calculating the number c of correct solutions
	- Estimating pass@k as 
	.. math: 1 - (1 - c/n)^k 
	
	when c < n, or 1 when c = n
        """
```

I also wonder if we should allow `k` to be a list (often, in papers, you report this metric for multiple k)

kjappelbaum: and the math syntax I used might be wrong (I didn't check how it renders) ... but it used to work like this

kjappelbaum: in any case, I'd add the equations we implement

src/corral/report.py
```diff
@@ -0,0 +1,240 @@
...
+        if len(self.trials) < k:
+            return 0.0
```
kjappelbaum: why 0? should it then simply not be defined?

src/corral/report.py
```diff
@@ -0,0 +1,240 @@
...
+        # For each group of k trials, check if at least one succeeded
+        successes = sum(
+            1 if any(t.success for t in self.trials[i : i + k]) else 0
+            for i in range(0, len(self.trials), k)
+        )
```
kjappelbaum: why do you need this grouping here?

src/corral/report.py
```diff
@@ -0,0 +1,240 @@
...
+        return successes / (len(self.trials) // k)
+
+    def calculate_pass_hat_k(self, k: int) -> float:
+        """Calculate pass^k - probability of all k trials succeeding"""
```
kjappelbaum: I would add the equation here, too

src/corral/report.py
```diff
@@ -0,0 +1,240 @@
...
+        )
+        return successes / (len(self.trials) // k)
+
+
+@dataclass
+class BenchmarkResult:
+    """Results from running benchmark with multiple trials per task"""
+
+    task_results: dict[str, TaskTrials]
+    k: int  # Number of trials to consider for pass@k metrics
```
kjappelbaum: i wonder if this should be here. It is already an argument for the method that computes it

src/corral/report.py
```diff
@@ -0,0 +1,240 @@
...
+            title="Overall Metrics", show_header=False, header_style="bold magenta"
+        )
+        summary_table.add_column("Metric", justify="left", style="cyan")
+        summary_table.add_column("Value", justify="right", style="white")
+
+        summary_table.add_row("Total Tasks", str(self.total_tasks))
+        summary_table.add_row("Trials per Task (k)", str(self.k))
+        summary_table.add_row("Average Score", f"{self.average_score:.3f}")
+        summary_table.add_row(f"Pass@{self.k}", f"{self.pass_at_k:.3f}")
+        summary_table.add_row(f"Pass^{self.k}", f"{self.pass_hat_k:.3f}")
```
kjappelbaum: i wonder how this scales as we add more metrics :)

src/corral/report.py
n0w0f: I am a bit confused 🤔 because your first comment suggest removing abstraction for `Result for n trails for a task`, and in the second comment you suggest having it. Am I understanding this correctly ?

So what i have done now is `TaskTrailResult` store result of one trail, `TaskTrialResults` store results of all trails for a task and `BenchmarkResult` takes care of computing Task results and overall benchmark results

src/corral/evaluate.py
n0w0f: Removed this, now only trials_per_task only through `.bench`

src/corral/report.py
n0w0f: Added this to the docstring

src/corral/report.py
n0w0f: removed returning 0. now we raise errors for all these cases

src/corral/report.py
n0w0f: this is now fixed. initially i was running the benchmark for n trails and computing score for k , which could be different

src/corral/report.py
n0w0f: removed it

src/corral/report.py
n0w0f: Eventually we can somehow make it scalable 🥲. For now it looks nice

src/corral/report.py
n0w0f: Also added list of `k` now

src/corral/report.py
```diff
@@ -0,0 +1,415 @@
...
+
+        Args:
+            k: The k value to use
+
+        Returns:
+            The average pass^k score across all tasks
+        """
+        return mean(self.task_pass_hat_k(task_id, k) for task_id in self.all_task_ids)
+
+    def generate_report(self, report_path: str | None = None) -> None:
```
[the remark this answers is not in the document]
n0w0f: this is done now.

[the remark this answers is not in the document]
n0w0f: this is now addressed

n0w0f:
@kjappelbaum @MrtinoRG, I have now addressed the comments can you review again.

src/corral/evaluate.py
```diff
@@ -105,28 +84,53 @@ def __init__(self, interface: BenchmarkInterface, agent: Agent):
...
+        task_ids: list[str] | None = None,
+        trials_per_task: int = 1,
+        k_values: int | list[int] | None = None,
+    ) -> BenchmarkResult:
+        """Run benchmark on specified tasks or all available tasks
+
+        Args:
+            task_ids: list of task_ids to run, or None for all tasks
+            trials_per_task: Number of trials per task, Default to k=1 to number of trials
+
```
MrtinoRG: `k_values` seems to be missing here. And I do not understand exactly what it is, especially because it can be a list of integers...

MrtinoRG: What is the purpose of this `k_values` variable because I see from the tests that it is just a list of integers

src/corral/report.py
```diff
@@ -0,0 +1,417 @@
...
+
+        pass@k is defined as:
+        .. math: P(pass@k) = 1 - (1 - p)^k
+
+        In practice, pass@k is estimated by:
+        - Taking n samples for each problem (where n ≥ k)
+        - Calculating the number c of correct solutions
+        - Estimating pass@k as
+        .. math: 1 - (1 - c/n)^k when c < n, or 1 when c = n
+
```
MrtinoRG: Perhaps you need to specify what it is p, c and n

MrtinoRG: Perhaps you can say that basically it is the accuracy over k passes

MrtinoRG (approved):
Nice, it looks much more comprehensive, thanks

src/corral/evaluate.py
n0w0f: thanks for spotting this, Added docstring for this in [8bcb7f5](https://github.com/lamalab-org/mat-agent-bench/pull/39/commits/8bcb7f5eee7f0e956e78403741b683f15cd55281)

src/corral/report.py
n0w0f: update the docs for this to 
```
Calculate pass@k - probability that at least one out of k trials succeeds for a given k value.

        pass@k is defined as:
        .. math: P(pass@k) = 1 - (1 - p)^k

        Where p is the probability of a single trial succeeding.

        In practice, pass@k is estimated by:
        - Taking n samples. (the number of total trials for a given task). (where n ≥ k)
        - Calculating the number c, the number of correct or successful solutions
        - Estimating pass@k as
        .. math: 1 - (1 - c/n)^k when c < n, or 1 when c = n
```

n0w0f: > What is the purpose of these k_values variable because I see from the tests that it is just a list of integers

It is to compute the pass score for a range of values, and not only for the number of trails we run

PR #40, opened 2025-02-26
Intel labs

### LINKED ISSUE #24: implement clearer way to store and share prompts
I feel we need to find a better naming convention - `BASELINESYSTEMPROMPTX` will be very hard to maintain and understand

_Originally posted by @kjappelbaum in https://github.com/lamalab-org/mat-agent-bench/pull/21#discussion_r1913808304_

MrtinoRG:
## Summary by Sourcery

This PR introduces new agent implementations, including a VLLM-based agent and a tool-calling agent, and refactors the ReAct agent to use a prompt store. It also adds API specifications and updates the benchmark interface to support tool retrieval.

New Features:
- Adds a VLLM-based agent implementation.
- Introduces a tool-calling agent.
- Adds an LLM planner agent that can use tool calling or React agent


src/corral/agents/llm_planner.py
```diff
@@ -0,0 +1,126 @@
...
+            final_answer, low_planner_messages = agent.run_agent(
+                interface=interface,
+                task_id=task_id,
+                task_prompt=plan,
+            )
+            if final_answer.split(" ")[0] != "Error":
+                messages.append(
+                    LiteLLMMessage(
+                        role="tool",
+                        content=f"Final Answer: {final_answer}.\nIteration by the agent:\n{low_planner_messages}",
```
MrtinoRG: I do not know how to better disentangle this, so the low-level planner messages are differentiated somehow :\

src/corral/agents/llm_planner.py
```diff
@@ -0,0 +1,126 @@
...
+                max_iterations=self.max_iterations,
+                api_endpoint=self.api_endpoint,
+                temperature=self.temperature,
+                **self.kwargs,
+            )
+        else:
+            agent = ReactAgent(
+                model=self.model,
+                max_iterations=10,
+            )
```
MrtinoRG: Maybe we even want the planner to define the `max_iterations` number, right?

src/corral/agents/tool_calling.py
```diff
@@ -0,0 +1,139 @@
...
+
+        if history is None:
+            history = []
+
+        tools = interface.get_available_tools(task_id)
+        # I think this task prompt is without the tools descriptions
+        if task_prompt is None:
+            task_guide = interface.get_task_prompt(task_id)
+        else:
+            task_guide = task_prompt
```
MrtinoRG, line 180: This is because when the call is made from the LLM Planner I would say that the task guide should only be the plan provided from High-level

src/corral/agents/prompts/index.json
MrtinoRG, line 1: This is so nice

src/corral/agents/llm_planner.py
```diff
@@ -0,0 +1,126 @@
...
+                    LiteLLMMessage(
+                        role="tool",
+                        content=f"Final Answer: {final_answer}.\nIteration by the agent:\n{low_planner_messages}",
+                    )
+                )
+                return final_answer, messages
+
+            messages.append(
+                LiteLLMMessage(
+                    role="tool",
```
MrtinoRG: Here it is treated as tool, perhaps it would be better to call it assistant 1 or something else different than tool right?

src/corral/agents/llm_planner.py
MrtinoRG, line 1: Right now it is a very simple version. Perhaps if we want to improve it we can play with the subtasks, and only focus on the next subtask that needs to be solved in the workflow

MrtinoRG:
The conflict can be solved by simply removing that file

src/corral/agents/llm_planner.py
MrtinoRG: Also I do not know if we want to pass these other agent messages to the main agent. And this is something that applies to other agents :)

src/corral/agents/utils.py
```diff
@@ -0,0 +1,111 @@
...
+    Returns:
+        List[LiteLLMMessage]: The messages returned by the model.
+    """
+    try:
+        response = litellm.completion(
+            model=model,
+            messages=messages,
+            temperature=temperature,
+            **kwargs,
+        )
```
MrtinoRG: Perhaps it is needed to add an `stop=["Observation:"]` parameter to the call

docs/API_specs.md
```diff
@@ -0,0 +1,217 @@
...
+
+
+@tool
+def tool1(arg1: [str, float, int], arg2: [str, float, int]) -> [str, float, int]:
+    """Perform basic math operations.
+
+    Args:
+        Describe the arguments here
+    """
+
```
kjappelbaum:
```suggestion
```

docs/API_specs.md
```diff
@@ -0,0 +1,217 @@
...
+def tool1(arg1: [str, float, int], arg2: [str, float, int]) -> [str, float, int]:
+    """Perform basic math operations.
+
+    Args:
+        Describe the arguments here
+    """
+    
+
+    result = function(arg1, arg2)
+
```
kjappelbaum, line 84:
```suggestion
```

docs/API_specs.md
```diff
@@ -0,0 +1,217 @@
...
+from dotenv import load_dotenv
+from prompts import BASELINESYSTEMPROMPT, BASELINEUSERPROMPT
+
+from corral.evaluate import BenchmarkInterface, MatAgentBenchmark
+
+from prompts import BASELINESYSTEMPROMPT, BASELINEUSERPROMPT
+from openai import OpenAI
+
+
+class Base_Agent:
```
kjappelbaum:
```suggestion
class BaseAgent:
```

docs/API_specs.md
```diff
@@ -0,0 +1,217 @@
...
+
+    def __init__(self, base_url: str, tools: list(str), args, **kwargs):
+
+        # Base Agent API implementation
+
+        self.client = OpenAI(base_url=base_url, api_key="dummy")
+
+        self.tools = tools
+
+    """ Solve Task function to solve task in the environment """
```
kjappelbaum:
```suggestion
```

docs/API_specs.md
```diff
@@ -0,0 +1,217 @@
...
+
+    def solve_task(self, interface: BenchmarkInterface, task_id: str) -> str:
```
kjappelbaum, line 135:
```suggestion
    def solve_task(self, interface: BenchmarkInterface, task_id: str) -> str:
    	    """ Solve Task function to solve task in the environment """
```

docs/API_specs.md
```diff
@@ -0,0 +1,217 @@
...
+
+    # Create components
+    interface = BenchmarkInterface()
+
+    # Load Tools from Environment
+
+    
+
+    # Load Agent and Runner
+    agent = Base_Agent(base_url='agent_url')
```
kjappelbaum:
```suggestion
    agent = BaseAgent(base_url='agent_url')
```

src/corral/agents/llm_planner.py
```diff
@@ -0,0 +1,133 @@
...
+        self.system_prompt = system_prompt
+        self.temperature = temperature
+        self.store = PromptStore("./prompts")
+        self.kwargs = kwargs
+
+    def run_agent(
+        self,
+        interface: BenchmarkInterface,
+        task_id: str,
+        user_prompt_uuid: str | None = "d77a15a2-4ded-4ceb-9b33-e84ab899c899",
```
kjappelbaum: I wonder if we should make this keyword argument here the prompt or the `uuid`. 
Benefit of `uuid` is that we handle the lookup in the promptstore for the user. 
Drawback of the `uuid` is that the user anyhow might need to use the prompstore to find the right `uuid`

src/corral/agents/llm_planner.py
```diff
@@ -0,0 +1,133 @@
...
+        system_prompt (str, optional): The system prompt to use
+        temperature (float): The temperature to use for sampling
+    """
+
+    def __init__(
+        self,
+        model: str = "gpt-4",
+        max_iterations: int = 5,
+        api_endpoint: str | None = None,
+        system_prompt: str | None = None,
```
kjappelbaum, line 41: Why do you use a `uuid` for the user prompt but not for the system prompt?

src/corral/agents/llm_planner.py
kjappelbaum: I also realize that the promptstore perhaps could also return some information about the fields the different prompt templates have

kjappelbaum: and I almost wonder if there would be value in having a "lamalab promptstore" that is reusable in different projects ... (a bit like this prompt sharing thing langchain has ...)

src/corral/agents/llm_planner.py
kjappelbaum: is this something people do in general? (I do not think so?) 

If this is not commonly done, I'd perhaps make it optional to let the planner define this

src/corral/agents/utils.py
kjappelbaum: I would not hard-code the stop sequence as it will depend on the agent you use

src/corral/agents/utils.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+@retry(
+    stop=stop_after_attempt(3),
+    wait=wait_exponential(multiplier=2, min=1),
+    retry=retry_if_exception_type(RETRY_EXCEPTIONS),
+    before_sleep=before_sleep_log(logger, log_level=20),
+    reraise=True,
+)
+def llm_tool_call(
```
kjappelbaum: Would there be a benefit in just having one `llm_call` function? 
I guess you could handle the case without toolcall by simply setting the `tools` to `None` if you do not do a tool call?

src/corral/agents/tool_calling.py
```diff
@@ -0,0 +1,145 @@
...
+
+class ToolCallingAgent:
+    """Agent that uses tool calling
+    This agent uses the LiteLLM API to solve tasks by calling tools.
+
+    Args:
+        model (str): The LiteLLM model to use. Defaults to "gpt-4".
+        max_iterations (int): The maximum number of iterations to run. Defaults to 10.
+        api_endpoint (str): The API endpoint to use when using VLLM. Defaults to None.
+        system_prompt (str): The system prompt to use. Defaults to "You are a helpful AI assistant that solves tasks step by step."
```
kjappelbaum: i wonder if the system prompt should not also be in the prompt store for consistency

src/corral/agents/tool_calling.py
```diff
@@ -0,0 +1,145 @@
...
+        self.api_endpoint = api_endpoint
+        self.system_prompt = system_prompt
+        self.temperature = temperature
+        self.store = PromptStore("./prompts")
+        self.kwargs = kwargs
+
+    def create_prompt(
+        self, prompt: PromptStore, task_guide: str, history: list[dict[str, Any]]
+    ) -> list[LiteLLMMessage]:
+        """Create the initial prompt messages for the agent
```
kjappelbaum, line 80:
```suggestion
        """Create the initial prompt messages for the agent
        
```

src/corral/agents/tool_calling.py
```diff
@@ -0,0 +1,145 @@
...
+        Args:
+            task_guide (str): The task guide to use.
+            history (List[Dict[str, Any]]): The history items to include.
+        Returns:
```
kjappelbaum:
```suggestion
            history (List[Dict[str, Any]]): The history items to include.
            
        Returns:
```

src/corral/agents/tool_calling.py
```diff
@@ -0,0 +1,145 @@
...
+        self.model = model
+        self.max_iterations = max_iterations
+        self.api_endpoint = api_endpoint
+        self.system_prompt = system_prompt
+        self.temperature = temperature
+        self.store = PromptStore("./prompts")
+        self.kwargs = kwargs
+
+    def create_prompt(
+        self, prompt: PromptStore, task_guide: str, history: list[dict[str, Any]]
```
kjappelbaum: is `prompt` really of type `PromptStore`?

src/corral/agents/tool_calling.py
```diff
@@ -0,0 +1,145 @@
...
+    def run_agent(
+        self,
+        interface: BenchmarkInterface,
+        task_id: str,
+        history: list[dict[str, Any]] | None = None,
+        prompt_uuid: str | None = "fe04453b-5469-4611-bba6-6d81487df787",
+        task_prompt: str | None = None,
+    ) -> tuple[str, list[LiteLLMMessage]]:
+        """Run the agent to solve the task
+        Args:
```
kjappelbaum:
```suggestion
        """Run the agent to solve the task
        
        Args:
```

src/corral/agents/tool_calling.py
```diff
@@ -0,0 +1,145 @@
...
+
+            content = llm_response.content
+            if content:
+                messages.append(LiteLLMMessage(role="assistant", content=content))
+                if "Final Answer:" in content:
+                    return content, messages
+
+            tool_calls = llm_response.tool_calls
+            if tool_calls:
+                called_tool = tool_calls[0]
```
kjappelbaum: will there sometimes be more than one tool called in the responses of the model API? Or is taking the first always the right thing to do?

src/corral/agents/llm_planner.py
```diff
@@ -0,0 +1,133 @@
...
+                **self.kwargs,
+            )
+        else:
+            agent = ReactAgent(
+                model=self.model,
+                max_iterations=10,
+            )
+
+        for _i in range(self.max_iterations):
+            final_answer, low_planner_messages = agent.run_agent(
```
kjappelbaum: perhaps I would name `low_planner_` even more explicit into `low_level_planner_`

src/corral/agents/llm_planner.py
kjappelbaum: what would you like to disentangle?

src/corral/agents/llm_planner.py
kjappelbaum: yes, perhaps the `role` should be the `type` and we could then have an additional name attribute (or a convention that we use `-` in the `role` to separate `type` and `name`

src/corral/agents/llm_planner.py
kjappelbaum: i think this is more than enough for first tests :)

src/corral/agents/tool_calling.py
MrtinoRG: Yes, I guess I did it in this way to do some tests.

Now it should be capturing all the tool calls

src/corral/agents/tool_calling.py
MrtinoRG: When it is filled should be `str`, right?

src/corral/agents/tool_calling.py
MrtinoRG: Now, I moved the default system prompt into the prompt store. Additionally, I moved the user prompt as a class argument that is initialized with the class if no other prompt is provided.

src/corral/agents/utils.py
MrtinoRG: Yes, certainly is a good way of keeping the code simpler

src/corral/agents/llm_planner.py
MrtinoRG: Done

MrtinoRG: Done, also now I used that (`low_level_planner`) as the name to differentiate the messages for this agent

src/corral/agents/llm_planner.py
MrtinoRG: Right now, for all the agents the prompts go into the class arguments and are filled through an external function

MrtinoRG: But yeah, I agree that we should not force the users to use the prompt store. 

About the info from the prompt store about the arguments, yeah definitely would be useful.

Finally, about the lab promptstore should not be very difficult once `pystow` is implemented

src/corral/agents/llm_planner.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+class LLMPlanner:
+    """Agent that uses the LLM planner to generate plans
+
+    Args:
+        model (str): The model to use for planning
+        max_iterations (int): The maximum number of iterations to plan
+        api_endpoint (str, optional): The API endpoint to use for tool calls
+        system_prompt (str, optional): The system prompt to use
```
kjappelbaum: perhaps I'd add here a note about the default

src/corral/agents/react.py
```diff
@@ -1,46 +1,91 @@
...
 @dataclass
 class Action:
     """Represents an action to be taken"""
+
     tool_name: str
-    arguments: Dict[str, Any]
+    arguments: dict[str, Any]
+
 
 class ReActAgent:
```
kjappelbaum, line 35: should we add a docstring for the constructor of the class?

src/corral/agents/react.py
```diff
@@ -1,46 +1,91 @@
...
-                {
-                    "role": "system",
-                    "content": "You are a helpful AI assistant that solves tasks step by step."
-                },
-                {"role": "user", "content": prompt}
-            ],
+            messages=messages,
             temperature=0.7,
-            max_tokens=1000
+            max_tokens=1000,
```
kjappelbaum: I'm wondering if we should make it possible to override those. What do you think?

kjappelbaum: in some of the other agents you have temperature as argument for the class constructor

src/corral/agents/tool_calling.py
```diff
@@ -0,0 +1,142 @@
...
+import json
+from typing import Any
+
+from promptstore import PromptStore
+
+from corral.agents.prompt_utils import get_prompt
+from corral.agents.utils import LiteLLMMessage, llm_call
+
+
+class ToolCallingAgent:
```
kjappelbaum, line 26: perhaps I'd ask copilot for a docstring here :)

src/corral/agents/utils.py
kjappelbaum, line 1: cool, this looks cleaner now!

src/corral/agents/tool_calling.py
```diff
@@ -0,0 +1,142 @@
...
+    ) -> tuple[str, list[LiteLLMMessage]]:
+        """Run the agent to solve the task
+
+        Args:
+            interface (BenchmarkInterface): The interface to use.
+            task_id (str): The task ID to solve.
+            history (List[Dict[str, Any]]): The history items to include. Defaults to None.
+
+        Returns:
+            str: The final answer from the agent.
```
kjappelbaum: in the type annotation and code you return a tuple

src/corral/agents/tool_calling.py
```diff
@@ -0,0 +1,142 @@
...
+        history: list[dict[str, Any]] | None = None,
+        task_prompt: str | None = None,
+    ) -> tuple[str, list[LiteLLMMessage]]:
+        """Run the agent to solve the task
+
+        Args:
+            interface (BenchmarkInterface): The interface to use.
+            task_id (str): The task ID to solve.
+            history (List[Dict[str, Any]]): The history items to include. Defaults to None.
+
```
kjappelbaum, line 167: task prompt seems to be missing in the docstring

kjappelbaum (approved):
very cool - looking forward what these agents can do!

src/corral/agents/react.py
MrtinoRG, line 1: So, know I deeply changed this agent to fill and return this list of `LiteLLMMessages` so we keep all the agents with the same convention.

src/corral/agents/react.py
```diff
@@ -69,66 +124,100 @@ def parse_llm_response(self, response: str) -> tuple[Thought | None, Action | No
...
                 )
-                history.append(observation)
+
+                messages.append(
+                    LiteLLMMessage(
+                        role="user",
+                        content=observation,
+                        name=action.tool_name,
+                    )
+                )
```
MrtinoRG, line 222: This is a bit strange, I know. Ideally, I would like to have a role "tool". The problem is that "tool" role is only allowed in the tool-calling mode after a tool call was requested in the previous call.

Like the question is what is the most appropriate role? I am giving that importance to the roles because if we do the graph thing I guess the roles can be important, if it is not the case, just drop this conversation

scripts/run_llm_planner_agent.py
```diff
@@ -0,0 +1,39 @@
...
+from loguru import logger
+
+from corral.agents.llm_planner import LLMPlanner
+from corral.evaluate import BenchmarkInterface, MatAgentBenchmark
+
+
+def setup_litellm():
+    """Setup LiteLLM with appropriate configuration"""
+    litellm.set_verbose = True
+    # os.environ["OPENAI_API_KEY"] = os.getenv("OPENAI_API_KEY")
```
kjappelbaum:
```suggestion
```

scripts/run_react_agent.py
```diff
@@ -11,10 +9,10 @@
 def setup_litellm():
     """Setup LiteLLM with appropriate configuration"""
     litellm.set_verbose = True
-    os.environ["OPENAI_API_KEY"] = os.getenv("OPENAI_API_KEY")
+    # os.environ["OPENAI_API_KEY"] = os.getenv("OPENAI_API_KEY")
```
kjappelbaum:
```suggestion
```

src/corral/agents/llm_planner.py
```diff
@@ -0,0 +1,167 @@
...
+                max_iterations=self.max_iterations,
+                api_endpoint=self.api_endpoint,
+                temperature=self.temperature,
+                prompt_store=self.store,
+                **self.kwargs,
+            )
+        else:
+            agent = ReActAgent(
+                model=self.model,
+                max_iterations=10,
```
kjappelbaum:
```suggestion
                max_iterations=10,
```

do we want to hard-code this?

src/corral/agents/llm_planner.py
```diff
@@ -0,0 +1,167 @@
...
+            examples (str): The examples to use for planning
+            tool_usage (bool): Whether to use tool calling or not
+
+        Returns:
+            Tuple[str, List[LiteLLMMessage]]: The final answer and messages
+        """
+        tools = interface.get_available_tools_for_task(task_id)
+
+        task_guide = interface.get_task_prompt(task_id)
+        if examples is not None:
```
kjappelbaum: is this what we expect? Why do we fill the `examples`  key if `examples` is `None`?

src/corral/agents/llm_planner.py
```diff
@@ -0,0 +1,167 @@
...
+    Args:
+        model (str): The model to use for planning
+        max_iterations (int): The maximum number of iterations to plan
+        api_endpoint (str, optional): The API endpoint to use for tool calls
+        system_prompt (str, optional): The system prompt to use.
+            Defaults to "You are a helpful AI assistant that solves tasks step by step."
+        user_prompt (str, optional): The user prompt to use
+        temperature (float): The temperature to use for sampling
+        prompt_store (PromptStore, optional): The prompt store to use. Defaults to None.
+        kwargs: Additional keyword arguments to pass to the LiteLLM API
```
kjappelbaum:
```suggestion
        kwargs: Additional keyword arguments to pass to the LiteLLM API for all LLM calls
```

src/corral/agents/react.py
```diff
@@ -27,25 +33,74 @@ class Action:
...
 
 class ReActAgent:
-    def __init__(self, model: str = "gpt-4", max_iterations: int = 10):
+    """
+    Agent that uses the ReAct framework to solve tasks
+    Based on https://arxiv.org/abs/2210.03629
+
+    Args:
+        model (str): The model to use for planning
+        max_iterations (int): The maximum number of iterations to plan
```
kjappelbaum: perhaps I would use a verb other than "planning" as "planning" might be easily confused with the LLM-Planning agents

src/corral/agents/react.py
```diff
@@ -69,66 +124,100 @@ def parse_llm_response(self, response: str) -> tuple[Thought | None, Action | No
...
+        history: list[LiteLLMMessage] | None = None,
+        task_prompt: str | None = None,
+    ) -> tuple[str, list[LiteLLMMessage]]:
+        """Main ReAct loop implementation
+
+        Args:
+            interface (BenchmarkInterface): The interface to use
+            task_id (str): The task ID to solve
+            history (List[Dict[str, Any]]): The history items to include. Defaults to None.
+            task_prompt (str): The task prompt to use. Defaults to None.
```
kjappelbaum: should we specify here what `task_prompt` actually means? That is, what information should it contain or where will it be inserted

src/corral/agents/react.py
kjappelbaum: I feel we should be able to solve the graph building with `name`. 

If we want to have additional roles, we can have a naming convention, where `role` is the first part of the name and then we add additional info with some separator

src/corral/agents/tool_calling.py
```diff
@@ -0,0 +1,226 @@
...
+class ToolCallingAgent:
+    """
+    Agent that uses native function calling from the providers to solve the task
+
+    Args:
+        model (str): The model to use for planning
+        max_iterations (int): The maximum number of iterations to plan
+        api_endpoint (str, optional): The API endpoint to use for tool calls
+        system_prompt (str, optional): The system prompt to use
```
kjappelbaum:
```suggestion
```

tasks/samplemath/samplemath/env.py
```diff
@@ -35,12 +35,12 @@ def score(self) -> float:
...
-            "math_4",
-            "What is twenty one thousand four hundred and seventy three * twenty one thousand four hundred and seventy three?",
-            4666829,
-        ),
+        # "math_3": MathEnvironment("math_3", "What is 99 * 63 * 999 * 111?", 691614693),
+        # "math_4": MathEnvironment(
+        #     "math_4",
+        #     "What is twenty one thousand four hundred and seventy three * twenty one thousand four hundred and seventy three?",
+        #     4666829,
+        # ),
```
kjappelbaum, line 43: should this be left commented out?

kjappelbaum (approved):
amazing job! I have one line of code I do not understand https://github.com/lamalab-org/mat-agent-bench/pull/40/files#r2008797378

src/corral/agents/llm_planner.py
MrtinoRG: Thanks for this catch. This was an error of course. 

How it is now and that I think that is correct is when examples is not provided, the prompt is filled with an empty string. The explanation for this is that in this way we can have zero and few-shot prompts in the same template.

src/corral/agents/llm_planner.py
```diff
@@ -0,0 +1,171 @@
...
+    Based on https://arxiv.org/abs/2212.04088
+
+    Args:
+        model (str): The model to use for planning
+        max_iterations (int): The maximum number of iterations to plan
+        api_endpoint (str, optional): The API endpoint to use for tool calls
+        system_prompt (str, optional): The system prompt to use.
+            Defaults to "You are a helpful AI assistant that solves tasks step by step."
+        user_prompt (str, optional): The user prompt to use
+        temperature (float): The temperature to use for sampling
```
kjappelbaum:
```suggestion
        temperature (float): The temperature to use for sampling. 
        	Defaults to 0.7. 
```

src/corral/agents/react.py
```diff
@@ -27,25 +33,74 @@ class Action:
...
-                    "content": "You are a helpful AI assistant that solves tasks step by step.",
-                },
-                {"role": "user", "content": prompt},
-            ],
-            temperature=0.7,
-            max_tokens=1000,
+        self.user_prompt = get_prompt(
+            self.store,
+            user_prompt,
+            "d880c4d3-fe60-4cf4-813b-2008076cd595",
```
kjappelbaum, line 85: why is the logic for system and user prompt different? For system prompt you only run `get_prompt` if `system_prompt=None`, but for `use_prompt` you always run `get_prompt`

src/corral/agents/react.py
```diff
@@ -69,66 +124,100 @@ def parse_llm_response(self, response: str) -> tuple[Thought | None, Action | No
...
+        """Main ReAct loop implementation
+
+        Args:
+            interface (BenchmarkInterface): The interface to use
+            task_id (str): The task ID to solve
+            history (List[Dict[str, Any]]): The history items to include. Defaults to None.
+            task_prompt (str): The task prompt to use. Defaults to None.
+
+        Returns:
+            Tuple[str, List[str]]: The final answer and history
```
kjappelbaum: this docstring does not agree with the type annotation

src/corral/agents/tool_calling.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+class ToolCallingAgent:
+    """
+    Agent that uses native function calling from the providers to solve the task
+
+    Args:
+        model (str): The model to use for planning
+        max_iterations (int): The maximum number of iterations to plan
+        api_endpoint (str, optional): The API endpoint to use for tool calls
```
kjappelbaum: is this the endpoint for VLLM? Then, the docstring might be confusing

src/corral/agents/tool_calling.py
```diff
@@ -0,0 +1,225 @@
...
+        system_prompt (str, optional): The system prompt to use.
+            Defaults to "You are a helpful AI assistant that solves tasks step by step."
+        temperature (float): The temperature to use for sampling
```
kjappelbaum:
```suggestion
        temperature (float): The temperature to use for sampling.
        	Defaults to 0.7.
```

src/corral/agents/tool_calling.py
```diff
@@ -0,0 +1,225 @@
...
+            get_prompt(
+                self.store, system_prompt, "400fcecf-f5f2-464b-aff5-8a4377c9685c"
+            ).fill({})
+            if system_prompt is None
+            else system_prompt
+        )
+
+        self.user_prompt = get_prompt(
+            self.store, user_prompt, "fe04453b-5469-4611-bba6-6d81487df787"
+        )
```
kjappelbaum, line 75: similar to above, I am unsure why system and user prompt are processed differently

src/corral/agents/tool_calling.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def create_prompt(
+        self, task_guide: str, history: list[LiteLLMMessage]
+    ) -> list[LiteLLMMessage]:
+        """Create the initial prompt messages for the agent
+
+        Args:
+            task_guide (str): The task guide to use.
+            history (LiteLLMMessage): The history items to include.
```
kjappelbaum:
```suggestion
            history (List[LiteLLMMessage]): The history items to include.
```

src/corral/agents/tool_calling.py
```diff
@@ -0,0 +1,225 @@
...
+            messages.extend(history)
+
+        return messages
+
+    def convert_to_openai_tool_format(self, tools_dict: dict) -> list:
+        """
+        Convert a dictionary of tools into the OpenAI tool calling format.
+
+        Args:
+            tools_dict: Dictionary with a 'tools' list containing tool specifications
```
kjappelbaum:
```suggestion
            tools_dict (dict): Dictionary with a 'tools' list containing tool specifications
```

src/corral/agents/tool_calling.py
```diff
@@ -0,0 +1,225 @@
...
+
+        Returns:
+            List of tools in OpenAI tool calling format
```
kjappelbaum:
```suggestion
            list: List of tools in OpenAI tool calling format
```

src/corral/agents/tool_calling.py
```diff
@@ -0,0 +1,225 @@
...
+        task_id: str,
+        history: list[LiteLLMMessage] | None = None,
+        task_prompt: str | None = None,
+    ) -> tuple[str, list[LiteLLMMessage]]:
+        """Run the agent to solve the task
+
+        Args:
+            interface (BenchmarkInterface): The interface to use.
+            task_id (str): The task ID to solve.
+            history (List[Dict[str, Any]]): The history items to include. Defaults to None.
```
kjappelbaum: the docstring here does not agree with the type hint

src/corral/utils.py
```diff
@@ -237,3 +241,86 @@ def decorator(func: Callable):
...
+    """Save agent conversation to a JSON file for logging and analysis purposes.
+
+    This function handles both regular dictionaries and LiteLLMMessage objects,
+    properly serializing them for storage.
+
+    Args:
+        messages: List of message objects (LiteLLMMessages or dictionaries)
+        task_id: The ID of the task being solved
+        agent_name: The name of the agent that generated the messages
+        output_dir: Directory to save the logs (will be created if it doesn't exist)
```
kjappelbaum, line 299: for consistency, I would also add types to this docstring here

kjappelbaum:
@MrtinoRG I have a few remaining questions - but if we can resolve them, this looks good, and merging this PR would be very useful. 

If there is anything unclear, I am happy to pass by for a quick chat.

src/corral/agents/react.py
MrtinoRG: The logic is that the system prompt does not need to be filled since it has no variables to fill. However, `user_prompt` has variables to fill

MrtinoRG: I know that this is not ideal because maybe at some point we may want to change the system prompt to make it flexible, but I do not know a good way to solve this in a flexible and general way

src/corral/agents/tool_calling.py
MrtinoRG: Same logic as above

src/corral/agents/react.py
MrtinoRG: Changed to run

src/corral/agents/react.py
MrtinoRG: Changed to: "task_prompt (str): The task prompt to use. task_prompt is intended to be a plan or description about the task, that should always be provided when this agent is called as a subagent of a main orchestrator. Defaults to None."

src/corral/agents/react.py
MrtinoRG: Thanks for pointing

src/corral/agents/tool_calling.py
MrtinoRG: Changed now

MrtinoRG: Or improved

src/corral/agents/tool_calling.py
MrtinoRG: Thanks for the catch

MrtinoRG:
> @MrtinoRG I have a few remaining questions - but if we can resolve them, this looks good, and merging this PR would be very useful.
> 
> If there is anything unclear, I am happy to pass by for a quick chat.

Thanks for the careful review @kjappelbaum

I have reviewed all the doc strings, and I think they are now in good shape. 

The only thing would be about the different prompt management between user and system prompts, but I do not have a general and flexible solution for that; so I am merging this now

PR #41, opened 2025-03-05
feat: add Llamp tasks

MrtinoRG:
## Summary by Sourcery

Add Llamp agent tools and task environment for Materials Project RAG benchmark

New Features:
- Implemented Materials Project RAG tools for querying materials properties
- Created a task environment for materials science benchmark tasks
- Added specialized expert agents for different materials properties

Enhancements:
- Extended corral framework to support materials science tasks
- Implemented flexible task dependency management
- Added scoring functions for materials science queries

Documentation:
- Added README for MP RAG task
- Included documentation in code comments explaining tool functionality


src/corral/agents/llamp.py
```diff
@@ -0,0 +1,386 @@
...
+            ).as_tool(),
+            MPStructureRetriever(
+                model=self.model,
+                interface=interface,
+                task_id=task_id,
+                kwargs=self.kwargs,
+            ).as_tool(),
+            arxiv,
+            wikipedia,
+            PythonREPLTool(),
```
MrtinoRG: Any comment on this if we want it or not? Safety stuff I mean

src/corral/agents/llamp.py
```diff
@@ -0,0 +1,386 @@
...
+                    LiteLLMMessage(
+                        role="user",
+                        content="Incorrect output format. Please follow the correct format.",
+                    )
+                )
+
+        return f"No final answer found after {self.max_iterations}", messages
+
+
+class MPSummaryExpert(MPAgent):
```
MrtinoRG: Right now two experts are missing from the original implementation: 
- `MPStructureVis` intended for visualizing structures in the frontend. I did not implement it because since we do not have frontend, we do not need it
- `SynthesisCallbackHandler` this one is needed, but I need to figure out how to do it since it needs structure outputs

src/corral/agents/llamp.py
MrtinoRG: I am not sure if these 3 are actually needed, but I guess that to be fair is better

src/corral/agents/llamp.py
```diff
@@ -0,0 +1,400 @@
...
+    MaterialsElectronic,
+    MaterialsMagnetism,
+    MaterialsPiezoelectric,
+    MaterialsStructureText,
+    MaterialsSummary,
+    MaterialsSynthesis,
+    MaterialsThermo,
+)
+
+# from corral.agents.utils import llm_tool_call
```
kjappelbaum:
```suggestion
```

src/corral/agents/llamp.py
```diff
@@ -0,0 +1,400 @@
...
+wikipedia = WikipediaQueryRun(api_wrapper=WikipediaAPIWrapper())
+arxiv = ArxivQueryRun(api_wrapper=ArxivAPIWrapper())
+
+
+class LiteLLMMessage(TypedDict, total=False):
+    role: str
+    content: str
+    name: str | None
+    tool_call_id: str | None
+
```
kjappelbaum: didn't you already define this somewhere else? It seems that this should be a class we make reusable

src/corral/agents/llamp.py
```diff
@@ -0,0 +1,400 @@
...
+                task_id=task_id,
+                kwargs=self.kwargs,
+            ).as_tool(),
+            arxiv,
+            wikipedia,
+            PythonREPLTool(),
+        ]
+
+        for tool in tools:
+            interface.add_tool_to_environment(task_id, tool)
```
kjappelbaum: we somehow here cross a bit the boundary that we built in the initial design: that the agent is not about tools but that the environment is about tools. 

i realize that `llamp` makes this separation difficult as it is a lot about materials-specific tools. 

however, in the initial design logic of `corral` we intended that the agents do not bring their own task-specific tools

kjappelbaum: perhaps it is worth if we (@MrtinoRG , @n0w0f and I) discuss this point

src/corral/agents/llamp.py
kjappelbaum: you mean because the Python REPL could do unsafe stuff?

src/corral/agents/llamp.py
kjappelbaum: that is, I would still move the tools away from the agent (also in the codebase) but have the `lammp` agent perhaps implemented such that the sub-agents then see different "available tools". 

depending on the environment, they also might see no tool at all when no tool suitable to a subtask is added to the environment

src/corral/agents/llamp/schemas.py
kjappelbaum, line 1: i must say that I didn't read the entire file - this is too much to review in one go :)

kjappelbaum: but I guess you took it from somewhere? (If yes, I'd perhaps add the source)

src/corral/agents/llamp/utilities.py
kjappelbaum, line 1: is this also taken from somewhere?

src/corral/agents/llamp.py
kjappelbaum: for some models json mode would be relatively easy to run with litellm (https://docs.litellm.ai/docs/completion/json_mode)

src/corral/agents/llamp.py
kjappelbaum: and this might also have the advantage that you could see what different approaches would do if they had the same tools

src/corral/agents/llamp.py
MrtinoRG: Yes, exactly

src/corral/agents/llamp.py
MrtinoRG: Never mind, because that agent is supposed to be used by another agent, but in [the original code it is commented](https://github.com/chiang-yuan/llamp/blob/main/api/src/llamp/mp/agents.py#L321).

src/corral/agents/llamp.py
MrtinoRG: This is needed since it is how the LLM calls are made. I commented because it is not in this branch since it will come from other PR, and it will give errors in some of the tests that I've made if not commented

src/corral/agents/llamp.py
MrtinoRG: Yes, removed now

src/corral/agents/llamp/schemas.py
MrtinoRG: Yes, this comes from the original Llamp implementation, and I just modified some stuff so it works with our environments.

I added now the correct reference at the beginning of the code

src/corral/agents/llamp/utilities.py
MrtinoRG: Yes, from Llamp, added now the correct reference

src/corral/agents/llamp.py
kjappelbaum: I think this can be solved by docs - we anyhow run in a "sandbox" but we could add somewhere a warning that this agent can execute arbitrary code and hence should be ran in a sandbox.

src/corral/agents/llamp/schemas.py
kjappelbaum: cool!

kjappelbaum:
This is in principle looks good to me, but I'd like to catch up on where we move the tools. We can do it perhaps in the jour fixe tomorrow

tasks/mp_rag_task/mp_rag_task/tools.py
```diff
@@ -0,0 +1,319 @@
...
+            interface=interface,
+            task_id=task_id,
+            kwargs=self.kwargs,
+        ).as_tool(),
+        "MP_dielectric_expert": MPDielectricExpert(
+            model=self.model,
+            interface=interface,
+            task_id=task_id,
+            kwargs=self.kwargs,
+        ).as_tool(),
```
MrtinoRG: Here, the problem with this is that these experts as tools need the model and other parameters such as temperature (the interface and task_id can be easily removed). 

I do not have a very clear idea of how to solve this :\

One possibility is to fix them to one configuration, but it is not ideal. 
Another option would be to migrate from experts to simple MP tools, but I guess something is missed in this way

@kjappelbaum some comment would be really appreciated :)

tasks/mp_rag_task/mp_rag_task/env.py
```diff
@@ -0,0 +1,214 @@
...
+    task_group = TaskGroup(
+        group_id="mp_rag",
+        tasks={
+            # Easy tasks
+            "NaCl_elastic_tensor": TaskDefinition(
+                name="NaCl Elastic Tensor",
+                description="What is the full elastic tensor of NaCl?",
+                tools=["MP_elasticity_expert"],
+                scoring_fn=check_elastic_tensor,
+                scoring_inputs={"material": "NaCl"},
```
MrtinoRG, line 42: The idea of this is to make it general so we can evaluate a bunch of different materials. Maybe for more advanced tasks (see Hard tasks), it has sense to test different materials of different complexity

tasks/mp_rag_task/mp_rag_task/tools.py
kjappelbaum: The question is whether you would like to be able for this to be modified during runtime or not. If not, why would passing a config in `create_tools` not work (e.g., a `dict `with keys being the toolnames and the values a `dict` with the `kwargs`).

But I feel I do not completely understand the problem

src/corral/base.py
```diff
@@ -214,25 +215,24 @@ def get_environment_guide(self) -> str:
...
-}}
-"""
+        return (
+            f"Task: {self.get_task_prompt()}\n\n"
+            "Available Tools:\n"
+            f"{tools_guide}\n\n"
+            "How to use tools:\n"
+            "1. Each tool call must specify the tool name and required arguments\n"
+            "2. Tools may return errors if arguments are invalid\n"
+            "3. You can make multiple tool calls as needed\n"
```
kjappelbaum: This comment was left also somewhere else: It might be good to specify that the different tool calls should be sequential

src/corral/base_task.py
kjappelbaum, line 1: this seems quite overlapping to Nawaf's PR

tasks/mp_rag_task/mp_rag_task/env.py
```diff
@@ -0,0 +1,214 @@
...
+                input_from_tasks=["retrieve_bandgap_formation_bulk_modulus"],
+                initial_input={
+                    "work_dir": str(work_dir_path),
+                },
+            ),
+            "create_vacancy_defect": TaskDefinition(
+                name="Create Vacancy Defect",
+                description="Generate a 2x2x2 supercell of the material and introduce a single vacancy defect (remove one atom). Save the resulting structure with the defect as a CIF file.",
+                tools=["MP_structure_retriever", "write_file"],
+                scoring_fn=check_cif_material,
```
kjappelbaum, line 160: probably we should also check if there is a vacancy ...

kjappelbaum: but can also be an enhancement at a later point

tasks/mp_rag_task/mp_rag_task/score.py
```diff
@@ -0,0 +1,428 @@
...
+
+    Returns:
+        list: List of tuples containing (material_id, property_value) for the material
+    """
+    material_ids = retrieve_material_id(material)
+    property_values = []
+
+    for material_id in material_ids:
+        with MPRester(API_KEY) as mpr:
+            if property_type == "elastic_tensor":
```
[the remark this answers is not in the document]
kjappelbaum: switch statements are cool :)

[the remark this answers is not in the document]
kjappelbaum: reminds me of good-old-times with cases in pascal ;)

tasks/mp_rag_task/mp_rag_task/score.py
```diff
@@ -0,0 +1,428 @@
...
+    Retrieves the formation energy values for a given material.
+
+    Args:
+        material: Material name, i.e. "NaCl"
+
+    Returns:
+        list[float]: List of formation energy values
+    """
+    results = retrieve_material_property(material, "formation_energy")
+    return [energy for _, energy in results]
```
kjappelbaum: this is very redundant

kjappelbaum: That is, I would consider one function that is parametrized with the property

tasks/mp_rag_task/mp_rag_task/score.py
```diff
@@ -0,0 +1,428 @@
...
+    """
+    try:
+        with Path.open(file_path) as f:
+            elastic_tensor = f.read()
+            return check_elastic_tensor(elastic_tensor, material)
+    except (OSError, FileNotFoundError):
+        return float(False)
+
+
+def check_simple_bandgap(bandgap: float, material: str) -> float:
```
kjappelbaum: Those range checks also seem a bit redundant and might be one function parametrized with the property

kjappelbaum (approved):

tasks/mp_rag_task/mp_rag_task/utilities.py
```diff
@@ -0,0 +1,658 @@
...
+    def run(self, function_name: str, function_args: str, debug: bool = False) -> str:
+        """
+        Performs an mp-api call and returns the result.
+
+        Args:
+            function_name: a function name to call
+            function_args: arguments for the function
+            debug: whether to print debug information
+        Returns:
+            function response in text format
```
kjappelbaum:
```suggestion
        Args:
            function_name (str): a function name to call
            function_args (str): arguments for the function
            debug (bool): whether to print debug information
            
        Returns:
            str: function response in text format
```

tasks/mp_rag_task/mp_rag_task/utilities.py
```diff
@@ -0,0 +1,658 @@
...
+
+        function_to_call = self.material_routes.get(function_name, None)
+        if function_to_call is None:
+            logger.warning(f"Function {function_name} is not supported yet.")
+            return (
+                re.sub(
+                    r"\s+",
+                    " ",
+                    f"""
+                I want to call {function_name} but it is not supported yet.
```
kjappelbaum: i know you took it from them, but I find the prompt "not supported yet" a bit strange. Not sure why we need the "yet"

tasks/mp_rag_task/mp_rag_task/utilities.py
```diff
@@ -0,0 +1,658 @@
...
+
+        if "formula_pretty" not in query_params.get("fields", []):
+            query_params["fields"] = [*query_params.get("fields", []), "formula_pretty"]
+
+        if "symmetry" not in query_params.get("fields", []):
+            query_params["fields"] = [*query_params.get("fields", []), "symmetry"]
+
+        # query_params["fields"] = query_params.get(
+        #     "fields", []) + ["structure", "material_id"]
+
```
kjappelbaum:
```suggestion

```

tasks/mp_rag_task/mp_rag_task/utilities.py
```diff
@@ -0,0 +1,658 @@
...
+        return_mode = query_params.pop("return_mode", "file")
+
+        limit = query_params.get("_limit", DEFAULT_LIMIT)
+
+        # return self.mpr.materials.summary._search(
+        #     num_chunks=None, chunk_size=1000, all_fields=False, **query_params
+        # )[:limit]
```
kjappelbaum:
```suggestion
```

tasks/mp_rag_task/mp_rag_task/utilities.py
```diff
@@ -0,0 +1,658 @@
...
+                structure.to_file(filename=path, fmt="json")
+                # write(path, atoms, format="extxyz")
+                paths.append(path)
+            return (
+                "All retrieved structures are saved as Pymatgen Structure json files to the following paths: "
+                + ", ".join(paths)
+            )
+        raise ValueError("Invalid return_mode")
+
+        # return [Structure.from_dict(doc["structure"]).to_ase_atoms() for doc in docs]
```
kjappelbaum:
```suggestion
```

tasks/mp_rag_task/mp_rag_task/utilities.py
```diff
@@ -0,0 +1,658 @@
...
+        if condition_mixing_media:
+            query_params["condition_mixing_media"] = condition_mixing_media.split(",")
+
+        response = self.mpr.materials.synthesis._search(**query_params)
+
+        return response[:5]
+
+    def search_materials_oxidation_states(self, query_params: dict):
+        query_params = self._process_query_params(query_params)
+        # FIXME
```
kjappelbaum, line 376: this seems like a concerning comment ;)

tasks/mp_rag_task/mp_rag_task/utilities.py
```diff
@@ -0,0 +1,658 @@
...
+        #             elastic_doc = self.mpr.materials.elasticity.get_data_by_id(
+        #                 document_id=material_id,
+        #                 fields=["pretty_formula", "elasticity", "task_id"],
+        #             )
+        #             elastic_docs.append(elastic_doc)
+
+        #         except Exception:
+        #             continue
+
+        #     return elastic_docs[:query_params.get("_limit", 10)]
```
kjappelbaum:
```suggestion
```

kjappelbaum (approved):
lgtm!

PR #60, opened 2025-04-07
track agent interaction with env as graph

n0w0f:
## Summary by Sourcery

Add graph tracking capabilities to the CORRAL framework, enabling detailed tracking and visualization of agent interactions across tasks

New Features:
- Introduced a GraphTracker class to capture agent interactions as a graph
- Added graph visualization capabilities with NetworkX
- Created a GraphTrackerFactory for managing graph tracking across different tasks and trials
- Implemented server-side graph tracking and visualization endpoints

Enhancements:
- Extended Environment and Agent classes to support graph tracking
- Added methods to track LLM prompts, responses, tool calls, and final submissions
- Implemented flexible metadata tracking for graph nodes
- Created methods for saving and loading graph data

Documentation:
- Updated server and environment initialization to support graph tracking
- Added new tracking methods to agents and environments

Tests:
- Added support for tracking agent and environment interactions
- Implemented graph statistics generation
- Created visualization methods for interaction graphs


src/corral/agents/react.py
MrtinoRG, line 1: I do understand why this has to be inside or with the agent. From my point of view, it can be done in two places.

First, it can be built at the end. The agents return the messages and the final answer. With the messages, it should be easy to build the graphs since, the messages contain everything. If we do not want to do regex again, we can save the actions and thought as different messages or something similar.

If for some reason that I am not appreciating the graph needs to be built dynamically, the second option is that I think it can be built within the evaluation class in `evaluate.py`. If every time you call a tool and when submit answer you pass the messages you can build the graph with the new messages. 

My reasons to say this also are because I feel it is not scalable (since you have to build all the agents with such a graph) and because it is how corral is conceived, right? We keep the agents separated and the environments isolated, so if we now add this dependency on a graph, we start breaking the inception concept right?

n0w0f: Mm,
Ideally i feel we would want to create the graph dynamically. This would also be useful for training.

I don't like the idea of interface managing graph creation. Also not every thoughts/ reason don't reach here. I feel the current design of server side tracking is the right thing to do.

But I agree with the difficulty in scalability. ie, agent need to let know the server/env. But to make this scalable, what i would now do is Enhance the Message itself to work with the `Mixin` to track all the interaction happening in the agent. So 
1. Extend LiteLLMMessage to include tracking metadata
2. Simplify TrackedAgentMixin to use a single track_message() method instead of track_llm_prompt, track_llm_response etc.

Also It is now also easier to track the cost, token count etc with the mixin.

kjappelbaum: I feel we might benefit from systematically comparing different implementation options. Shall we do it here https://www.notion.so/Graph-tracking-design-1d2988a4c88a80e3ba1adf00a992b587?pvs=4?

n0w0f:
closing this as this diverged a lot and would be easier to start from scratch

PR #67, opened 2025-04-10
feat: add benchmark state

### LINKED ISSUE #56: error handling in agents
result = runner.bench(task_ids, trials_per_task=2, k_values=[1, 2], verbose=True)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/n0w0f/git/n0w0f/mat-agent-bench/src/corral/evaluate.py", line 142, in bench
    answer, messages = self.agent.run_agent(self.interface, task_id)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/n0w0f/git/n0w0f/mat-agent-bench/src/corral/agents/llm_planner.py", line 139, in run_agent
    if "Final Answer:" in plan:
       ^^^^^^^^^^^^^^^^^^^^^^^
TypeError: argument of type 'NoneType' is not iterable

MrtinoRG:
## Summary by Sourcery

Add benchmark state management with checkpointing support to allow resuming benchmark runs

New Features:
- Implement checkpointing mechanism for benchmark runs to allow resuming interrupted sessions
- Add session ID support for tracking and resuming benchmark progress

Enhancements:
- Improve error handling in benchmark and agent execution
- Add flexibility to benchmark runs with optional temperature parameter
- Enhance logging and checkpoint management

Chores:
- Update multiple agent scripts to support temperature parameter
- Modify benchmark interface to support checkpointing

### DIFF
--- a/src/corral/agents/tool_calling.py
+++ b/src/corral/agents/tool_calling.py
@@ -3,12 +3,17 @@ import json
 from dataclasses import dataclass
 from typing import Any
 
+from loguru import logger
 from promptstore import PromptStore
 
 from corral.agents.prompt_utils import get_prompt
-from corral.agents.utils import LiteLLMMessage, _build_user_content, llm_call
+from corral.agents.utils import (
+    LiteLLMMessage,
+    _build_user_content,
+    convert_to_openai_tool_format,
+    llm_call,
+)
 from corral.evaluate import BenchmarkInterface
-from loguru import logger
 
 
 @dataclass
@@ -100,88 +105,6 @@ class ToolCallingAgent:
 
         return messages
 
-    def convert_to_openai_tool_format(self, tools_dict: dict) -> list:
-        """
-        Convert a dictionary of tools into the OpenAI tool calling format.
-    
-        Args:
-            tools_dict (dict): Dictionary with a 'tools' list containing tool specifications
-    
-        Returns:
-            list: List of tools in OpenAI tool calling format
-        """
-        openai_tools = []
-    
-        for tool in tools_dict["tools"]:
-            function = {
-                "name": tool["name"],
-                "description": tool["description"],
-                "parameters": {"type": "object", "properties": {}, "required": []},
-            }
-    
-            if isinstance(tool["arguments"], str):
-                arg_names = [arg_name.strip() for arg_name in tool["arguments"].split(",")]
-                
-                for arg_name in arg_names:
-                    if arg_name:
-                        property_entry = {
-                            "type": "string",
-                            "description": f"Argument: {arg_name}"
-                        }
-                        function["parameters"]["properties"][arg_name] = property_entry
-                        function["parameters"]["required"].append(arg_name)
-                
-            else:
-                for arg in tool["arguments"]:
-                    if not isinstance(arg, dict):
-                        raise ValueError(
-                            f"Expected arg to be a dictionary but got: {arg}"
-                        )
-                    
-                    arg_type = arg.get("type")
-                    if not arg_type:
-                        raise ValueError(
-                            f"Argument type is missing for argument: {arg['name']}"
-                        )
-                    
-                    if arg_type == "str":
-                        arg_type = "string"
-                    elif arg_type == "bool":
-                        arg_type = "boolean"
-                    elif arg_type in ["int", "float"]:
-                        arg_type = "number"
-                    elif arg_type == "list[str]":
-                        arg_type = "array"
-                        property_entry = {
-                            "type": arg_type,
-                            "description": arg.get("description", ""),
-                            "items": {"type": "string"},
-                        }
-                        function["parameters"]["properties"][arg["name"]] = property_entry
-    
-                        if arg.get("required", False):
-                            function["parameters"]["required"].append(arg["name"])
-    
-                        continue
-                    else:
-                        raise ValueError(
-                            f"Unsupported argument type: {arg_type} for argument: {arg['name']}"
-                        )
-    
-                    property_entry = {"type": arg_type, "description": arg.get("description", "")}
-    
-                    if arg.get("choices"):
-                        property_entry["enum"] = arg["choices"]
-    
-                    function["parameters"]["properties"][arg["name"]] = property_entry
-    
-                    if arg.get("required", False):
-                        function["parameters"]["required"].append(arg["name"])
-    
-            openai_tools.append({"type": "function", "function": function})
-    
-        return openai_tools
-
     def run_agent(
         self,
         interface: BenchmarkInterface,
@@ -193,19 +116,19 @@ class ToolCallingAgent:
         """Run the agent to solve the task"""
         if history is None:
             history = []
-    
-        tools = self.convert_to_openai_tool_format(
+
+        tools = convert_to_openai_tool_format(
             interface.get_available_tools_for_task(task_id)
         )
         if task_prompt is None:
             task_guide = interface.get_task_prompt(task_id)
         else:
             task_guide = task_prompt
-    
+
         messages = self.create_prompt(
             task_guide=task_guide, history=history, examples=examples
         )
-    
+
         for _i in range(self.max_iterations):
             try:
                 llm_response = llm_call(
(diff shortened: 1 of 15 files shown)


src/corral/evaluate.py
```diff
@@ -99,75 +103,125 @@ def get_task_status(self, task_id: str) -> dict[str, Any]:
...
+    def __init__(self, interface: BenchmarkInterface, agent: Agent, 
+                 checkpoint_dir: str = "./benchmark_checkpoints"):
         self.interface = interface
         self.agent = agent
+        self.checkpoint_dir = Path(checkpoint_dir)
+        self.checkpoint_dir.mkdir(parents=True, exist_ok=True)
+
+    def _get_checkpoint_path(self, session_id: str):
+        """Get path for checkpoint file"""
+        return self.checkpoint_dir / f"checkpoint.pkl"
```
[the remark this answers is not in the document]
MrtinoRG: This did not work...

src/corral/agents/tool_calling.py
```diff
@@ -102,58 +103,83 @@ def create_prompt(
...
-                if arg["choices"]:
-                    property_entry["enum"] = arg["choices"]
-
-                function["parameters"]["properties"][arg["name"]] = property_entry
-
-                if arg["required"]:
-                    function["parameters"]["required"].append(arg["name"])
-
+    
             openai_tools.append({"type": "function", "function": function})
```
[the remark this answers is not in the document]
MrtinoRG: Please write the proposal for this change as a committable suggestion

src/corral/evaluate.py
```diff
@@ -99,75 +103,125 @@
...
+            session_id = str(uuid.uuid4())
+            
         if k_values is None:
             k_values = list(range(1, trials_per_task + 1))
         elif isinstance(k_values, int):
             k_values = [k_values]
         elif isinstance(k_values, list) and max(k_values) > trials_per_task:
             raise ValueError("k value is greater than the number of trials")
 
+        checkpoint = self._load_checkpoint(session_id)
```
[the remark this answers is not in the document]
MrtinoRG: Please write the proposal for this change as a committable suggestion

src/corral/agents/tool_calling.py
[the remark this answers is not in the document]
MrtinoRG: Can you write the code as a committable suggestion?

src/corral/evaluate.py
```diff
@@ -99,75 +103,125 @@ def get_task_status(self, task_id: str) -> dict[str, Any]:
...
+        """Save checkpoint with current benchmark progress"""
+        checkpoint = {
+            'task_results': task_results,
+            'current_task': current_task,
+            'completed_trials': completed_trials,
+            'session_id': session_id
+        }
+        with open(self._get_checkpoint_path(session_id), 'wb') as f:
+            pickle.dump(checkpoint, f)
+        logger.info(f"Checkpoint saved for session {session_id}")
```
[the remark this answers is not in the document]
kjappelbaum: this seems like a valid concern

src/corral/evaluate.py
```diff
@@ -99,75 +103,125 @@ def get_task_status(self, task_id: str) -> dict[str, Any]:
...
 class MatAgentBenchmark:
     """Runs benchmarks using an agent implementation"""
 
-    def __init__(self, interface: BenchmarkInterface, agent: Agent):
+    def __init__(self, interface: BenchmarkInterface, agent: Agent, 
+                 checkpoint_dir: str = "./benchmark_checkpoints"):
         self.interface = interface
         self.agent = agent
+        self.checkpoint_dir = Path(checkpoint_dir)
+        self.checkpoint_dir.mkdir(parents=True, exist_ok=True)
```
kjappelbaum, line 126: potential caveats here: 

- if we have multiple runs that we start from the same directory, those checkpoint files will be conflicting 
- it is unclear if we really want to default to overriding existing checkpoints

src/corral/evaluate.py
```diff
@@ -99,75 +103,125 @@ def get_task_status(self, task_id: str) -> dict[str, Any]:
...
+
+    def _get_checkpoint_path(self, session_id: str):
```
kjappelbaum: did you plan to use the `session_id` in the filename?

src/corral/evaluate.py
MrtinoRG: Yes, you are right. I will try to think in some way to differentiate the checkpoint names in an automatic way

src/corral/evaluate.py
MrtinoRG: I tried, and it was failing somehow

src/corral/evaluate.py
```diff
@@ -99,75 +103,125 @@ def get_task_status(self, task_id: str) -> dict[str, Any]:
...
-        """Run benchmark on specified tasks or all available tasks
+        """Run benchmark on specified tasks or all available tasks with checkpointing
 
         Args:
             task_ids: list of task_ids to run, or None for all tasks
             trials_per_task: Number of trials per task, Default to k=1 to number of trials
             k_values: list of k values, for which pass metrics are calculated. Default to [1, 2, 3, ..., trials_per_task]
-
+            verbose: Whether to save agent messages
+            session_id: Unique ID for this benchmark session, used for checkpointing
```
kjappelbaum: perhaps I'd add a note that you autogenerate one if there is none provided
if it is autogenerated, however, the question is how users find out what session id the current run is/was

src/corral/evaluate.py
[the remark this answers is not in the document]
MrtinoRG: The import is there. Maybe the copilot only has access to the lines changed in this PR?

src/corral/agents/tool_calling.py
```diff
@@ -192,49 +130,65 @@ def run_agent(
...
+                            if result is None:
+                                result = function_call.error
+                        except Exception as e:
+                            result = str(e)
+
+                        function_name = str(called_tool.function.name)
+
+                        messages.append(
+                            LiteLLMMessage(
+                                role="tool",
```
kjappelbaum, line 174: the fact that we specify the roles in this way should be documented in paper and the docs here I guess. 

There might be effects that this choice has (but I think it is the correct choice) but I think it would be good to be very transparent about such implementation details

src/corral/agents/tool_calling.py
```diff
@@ -164,22 +113,11 @@ def run_agent(
...
-
-        Returns:
-            tuple[str, List[LiteLLMMessage]]: The final answer and messages.
-        """
+        """Run the agent to solve the task"""
         if history is None:
             history = []
 
-        tools = self.convert_to_openai_tool_format(
+        tools = convert_to_openai_tool_format(
```
kjappelbaum, line 120: nice,  I like this change!

src/corral/evaluate.py
```diff
@@ -110,64 +112,141 @@ def solve_task(self, interface: BenchmarkInterface, task_id: str) -> str:
...
+        """Run benchmark on specified tasks or all available tasks with checkpointing
 
         Args:
             task_ids: list of task_ids to run, or None for all tasks
             trials_per_task: Number of trials per task, Default to k=1 to number of trials
             k_values: list of k values, for which pass metrics are calculated. Default to [1, 2, 3, ..., trials_per_task]
-
+            verbose: Whether to save agent messages
+            session_id: Unique ID for this benchmark session, used for checkpointing.
+                If None, it will be an empty string, and the checkpoint will be saved as checkpoint_{agent_name}_.pkl
```
kjappelbaum: how about datetime or some UUID if the`session_id`  is`None`?

kjappelbaum (approved):
looks very good!

src/corral/evaluate.py
MrtinoRG: Done now with datetime. Datetime because the runs are more identifiable I think

PR #68, opened 2025-04-11
extended i/o operations for modal framework

chandan21gupta:
This branch extends i/o operations to work with a deployed modal app. This is essential so that same abstraction layer can be used for any type of i/o across many applications/protocols.


scripts/run_react_agent.py
```diff
@@ -15,12 +15,12 @@ def run_benchmark(model: str = "gpt-4o", task_ids: list | None = None):
     """Run the benchmark with specified model and tasks"""
 
     interface = BenchmarkInterface()
-    agent = ReActAgent(model=model)
+    agent = ReActAgent(model=model, max_iterations = 15, temperature = 0.2)
```
MrtinoRG: Do you think we should increase the default value for `max_iterations`?

src/corral/io.py
```diff
@@ -1,29 +1,37 @@
...
 
 from corral.base import Tool, ToolArgument
 
-
 class FSManager:
     """A file-system abstraction layer using fsspec.
     This object is created with a given protocol (e.g., "file", "s3", "ftp")"""
 
-    def __init__(self, protocol: str = "file", **kwargs):
+    def __init__(self, protocol: str = "file", app: Optional[str] = None, **kwargs):
```
MrtinoRG:
```suggestion
    def __init__(self, protocol: str = "file", app: str | None = None, **kwargs):
```
In this way pre-commit checks pass

src/corral/io.py
MrtinoRG, line 1: Perhaps it would be useful if you push the remote functions that you are using in case someone else is facing the same problem with modal io tools

MrtinoRG (approved):
LGTM! Thanks :)

MrtinoRG (approved):

PR #74, opened 2025-04-15
feat: spectra elucidation environments

### LINKED ISSUE #66: organic spectroscopy environment
Tasks: We provide spectra (e.g. NMR, perhaps also IR, MS) - can be in the text from that is typically provided in journals - and we want to obtain structures (as SMILES or selection from MCQ options). 

Tools: 

- NMR prediction model (can perhaps even be something simple as based on HOSE code). Proper shift prediction simulations take some time 
- Database lookup: for example, NMRshiftDB 
- RAG on databases of shifts/assignment tables

MrtinoRG:
## Summary by Sourcery

Add spectra elucidation environments for easy and hard tasks, including vector database creation, tools, and task management

New Features:
- Created two new spectra elucidation task environments (easy and hard modes)
- Implemented vector database creation script for NMRShiftDB2
- Added tools for chemical spectra analysis and molecule similarity scoring

Enhancements:
- Extended vector database search functionality in utils.py
- Added metadata support for vector database creation
- Improved vector database processing with batch handling

Chores:
- Removed unused tools from chembench
- Updated task definition to include scoring inputs
- Organized new task environments with modular structure


db_scripts/create_nmrshiftdb2.py
```diff
@@ -0,0 +1,232 @@
...
+
+load_dotenv("../.env", override=True)
+
+
+def parse_nmr_properties(properties):
+    """
+    Parse NMR properties from a dictionary into a clean format.
+
+    Args:
+        properties (dict): Dictionary of properties from an RDKit molecule
```
kjappelbaum: minor nitpick: it would be useful to know if there are special things that this dictionary needs - or how it can be produced

src/corral/utils.py
```diff
@@ -221,12 +226,157 @@ def _tokenize_and_split_chunks(
...
+    BATCH_SIZE = 1000
+    total_processed = 0
+
+    logger.info(
+        f"Processing {len(processed_chunks)} documents with embeddings in batches of {BATCH_SIZE}"
+    )
+
+    # Process in batches to reduce memory usage
+    for batch_idx in range(0, len(processed_chunks), BATCH_SIZE):
+        batch_end = min(batch_idx + BATCH_SIZE, len(processed_chunks))
```
kjappelbaum: for such iterations I sometimes find `more_itertools` useful. This also has a `chunked` method

kjappelbaum: (o need to change anything, just wanted to point out the package in case you do not already know it)

tasks/spectra_elu_easy/spectra_elu_easy/env.py
```diff
@@ -0,0 +1,180 @@
...
+    # Create all available tools
+    available_tools = create_tools()
+
+    # Create task definitions for each task in TASKS
+    task_definitions = {}
+    for task in tasks:
+        task_id = task["name"]
+        task_definitions[task_id] = TaskDefinition(
+            name="Organic Compound Elucidation",
+            description=f"Return the SMILES string or name for the compound with the following characterization analysis: {task['spectra']}",
```
kjappelbaum:
```suggestion
            description=f"Return the SMILES string or name for the compound with the following characterization data: {task['spectra']}",
```

tasks/spectra_elu_easy/spectra_elu_easy/score.py
```diff
@@ -0,0 +1,67 @@
...
+        return 0.0
+
+    if Chem.MolToSmiles(pred_mol) == Chem.MolToSmiles(true_mol):
+        return 1.0
+
+    # Check scaffold match
+    pred_scaffold = MurckoScaffold.GetScaffoldForMol(pred_mol)
+    true_scaffold = MurckoScaffold.GetScaffoldForMol(true_mol)
+    if Chem.MolToSmiles(pred_scaffold) == Chem.MolToSmiles(true_scaffold):
+        return 0.8
```
kjappelbaum, line 48: what does this mean? why do you do this scaffold check and why do you get a score of 0 if they match?

tasks/spectra_elu_easy/spectra_elu_easy/tools.py
```diff
@@ -0,0 +1,371 @@
...
+
+    Args:
+        smiles (str): SMILES string to search for
+        top_k (int, optional): Maximum number of results to return. Defaults to 10
+
+    Returns:
+        list: Matching entries
+    """
+    collection_name = "nmrshiftdb2"
+    db_path = Path(__file__).resolve().parents[3] / "vector_databases" / "nmrshiftdb2"
```
kjappelbaum, line 224: are we sure that this `parents[3]` way of getting the path will always work?

tasks/spectra_elu_easy/spectra_elu_easy/tools.py
```diff
@@ -0,0 +1,371 @@
...
+    )
+
+
+@tool
+def retrieve_protons_shifts() -> list[dict[str, str]]:
+    """
+    Retrieve the proton shifts ranges for hydrocarbons.
+
+    Returns:
+        list: A list of dictionaries containing the proton shifts ranges
```
kjappelbaum: perhaps I would more explicitly say that the shifts are w.r.t. TMS

tasks/spectra_elu_easy/spectra_elu_easy/tools.py
```diff
@@ -0,0 +1,371 @@
...
+    )
+
+
+@tool
+def retrieve_aromatic_protons_shifts() -> list[dict[str, str]]:
+    """
+    Retrieve the proton shifts ranges for aromatic hydrocarbons.
+
+    Returns:
+        list: A list of dictionaries containing the proton shifts ranges
```
kjappelbaum: I'd also reference here what the standard is 

I'm also not sure if the unicode in the subsitutents might not create some problems

tasks/spectra_elu_hard/spectra_elu_hard/tools.py
```diff
@@ -0,0 +1,370 @@
...
+            },
+            {"Substituent": "NH₂", "Ortho": -0.75, "Meta": -0.24, "Para": -0.63},
+            {"Substituent": "SCH₃", "Ortho": -0.03, "Meta": 0, "Para": -0.3},
+            {"Substituent": "N(CH₃)₂", "Ortho": -0.6, "Meta": -0.1, "Para": -0.62},
+        ]
+    )
+
+
+@tool
+def retrieve_carbon_shifts() -> list[dict[str, str]]:
```
kjappelbaum, line 333: i'd have here similar comments as above (naming the standard and unicode)

tasks/spectra_elu_easy/spectra_elu_easy/tools.py
```diff
@@ -0,0 +1,371 @@
...
+        return rdMolDescriptors.CalcMolFormula(mol)
+
+    except Exception as e:
+        return f"Error: {e!s}"
+
+
+@tool
+def simulate_spectra(smiles: str) -> dict[str, str]:
+    """
+    Simulate 1H NMR, 13C NMR, and IR spectra for a given molecule using its SMILES string.
```
kjappelbaum, line 191: perhaps I'd add some comments on how this simulation works (I guess here it is simply the ML model?)

kjappelbaum (approved):
some nitpick, but looks overall very nice and useful!

tasks/spectra_elu_easy/spectra_elu_easy/score.py
MrtinoRG: If the scaffold matches but the molecules do not, I assume that some parts are correct, so instead of giving the full score (1.0), I give a reduced score of 0.8. I guess that the score can be optimized but I think that for now it will do it

tasks/spectra_elu_easy/spectra_elu_easy/tools.py
MrtinoRG: Yes, by defining the Path depending of `__file__` I think that in that way the path would be robust with the package and then I can assume that the vector database is going to be always in place

PR #81, opened 2025-05-26
fix: run trails based on dependency

n0w0f:
- Issue: subsequent tasks can have dependency on previous tasks, hence there is only particular order in which tasks can be solved.

- Solution:
    - Added a variable in environment
    - If there is dependency between tasks in this environment we set this to True
    - If dependency=True, we run first trail of all tasks and only then will we move to next tasks

- Issue: We need to adapt checkpointing based on dependency.
- Solution
    - If there is dependency we checkpoint after each trail otherwise we do it after each task

    Task-Level (dependency_chain=False):
    ```
    Task1: Trial1 ✓ Trial2 ✓ Trial3 ✓ → Checkpoint: completed_tasks=["task1"]  
    Task2: Trial1 ✓ Trial2 ✗ → Restart Task2 from Trial1
    ```
    Trial-Level (dependency_chain=True):
    ```
    Trial1: Task1 ✓ Task2 ✓ Task3 ✓ → Checkpoint: completed_trials=1
    Trial2: Task1 ✓ Task2 ✗ → Restart Trial2 for ALL tasks
    ```

## Summary by Sourcery

Add support for dependency chains in benchmark runs by detecting environment dependencies and switching between trial-level and task-level checkpointing strategies, and refactor core utilities to provide robust tooling, serialization, and API integration capabilities.

New Features:
- Detect and apply dependency chaining in BenchmarkInterface.bench to control run order and checkpoint granularity
- Expose dependency_chain setting via new server endpoint and environment variable across tasks
- Introduce remote_call decorator for invoking functions in remote Modal environments
- Add web_search function using Brave Search API with embedding-based ranking

Bug Fixes:
- Ensure tasks and trials execute in correct order when dependencies are present by adjusting run and checkpoint logic

Enhancements:
- Overhaul corral.utils with type annotation formatting, docstring parsing, Tool and ModalTool decorators, message serialization, API retry utilities, and vector database helpers
- Replace legacy checkpoint resume logic and TaskResult dataclass with unified checkpoint file methods for task-level and trial-level checkpointing
- Update ChemBench and Catalyst environments to propagate chained_tasks setting from environment and to initialize models consistently

Chores:
- Remove deprecated chembench dependency from pyproject.toml
- Increase default trials_per_task to 2 in run_react_agent.py
- Add chained_tasks attribute to base Environment in corral.base


MrtinoRG:
@n0w0f should we close this one since it is already in #82?

n0w0f:
okay

PR #82, opened 2025-05-26
fix: directory creation based on chaining and trails

n0w0f:
# Temperory directory creation

- Issue: subsequent tasks can have dependency on previous tasks, hence there is only particular order in which tasks can be solved.

- Solution:
    - Added a variable in environment
    - If there is dependency between tasks in this environment we set this to True
    - If dependency=True, we run first trail of all tasks and only then will we move to next tasks

- Issue: We need to adapt checkpointing based on dependency.
- Solution
    - If there is dependency we checkpoint after each trail otherwise we do it after each task

    Task-Level (dependency_chain=False):
    ```
    Task1: Trial1 ✓ Trial2 ✓ Trial3 ✓ → Checkpoint: completed_tasks=["task1"]  
    Task2: Trial1 ✓ Trial2 ✗ → Restart Task2 from Trial1
    ```
    Trial-Level (dependency_chain=True):
    ```
    Trial1: Task1 ✓ Task2 ✓ Task3 ✓ → Checkpoint: completed_trials=1
    Trial2: Task1 ✓ Task2 ✗ → Restart Trial2 for ALL tasks
    ```
    
- Issue : Each trail needs seperate folder where io will be executes, files from tools wil be saved, otherwise there could contamination from previous trails. Which means
    
    - Each trial gets its own working directory (chained=True). Each task get its own directory  (chained=False)
    
    - Tasks that depend on previous tasks need access to their outputs

    - FSManager and file tools need to work with the new directories

## Summary by Sourcery

Implement isolated working directories based on chaining settings, add dependency-aware task execution with two levels of checkpointing, and refactor core utilities for tool creation, file management, embedding, and search functionality

New Features:
- Add support for dependency chaining with an API endpoint and dynamic workspace isolation per trial or per task
- Introduce separate checkpointing strategies (task-level and trial-level) in the benchmark runner

Bug Fixes:
- Ensure each trial or task gets its own working directory to avoid cross-trial contamination

Enhancements:
- Refactor FSManager to use a base path and resolve relative paths automatically
- Extend Environment and TaskGroupEnvironment to create and manage isolated workspaces based on chaining settings
- Overhaul utils for parsing docstrings, generating Tools/ModalTools, embedding text, API requests, web and vector search flows
- Remove outdated checkpointing functions and consolidate related logic into modular methods
- Introduce WorkspaceStrategy enum for configuring workspace isolation


src/corral/base.py
```diff
@@ -210,25 +214,35 @@ def reset_state(self) -> str:
...
+            # For chained tasks: one folder per trial (shared across all tasks in chain)
+            workspace = Path(self.base_work_dir) / f"trial_{trial_id}"
+        else:
+            # For independent tasks: one folder per task per trial
+            workspace = Path(self.base_work_dir) / f"{self.task_id}_trial_{trial_id}"
+
+        logger.info(f"DEBUG: Creating workspace: {workspace}")
+        workspace.mkdir(parents=True, exist_ok=True)
+        logger.info(f"DEBUG: Workspace created successfully: {workspace}")
+        return str(workspace)
```
MrtinoRG: I think that ideally this should, if the directory exist, create it from scratch or create a new different directory to avoid overlapping between different runs

src/corral/evaluate.py
```diff
@@ -214,42 +153,219 @@ def bench(
...
+    def _save_trial_level_checkpoint(
+        self, session_id: str, completed_trials: int
+    ) -> None:
+        """Save trial-level checkpoint"""
+        checkpoint = {
+            "type": "trial_level",
+            "completed_trials": completed_trials,
+            "session_id": session_id,
+            "timestamp": datetime.now(tz=timezone.utc).isoformat(),
+        }
```
MrtinoRG: So, where are the results saved? I guess in the env, right? I would suggest adding them here as well in case the server is turned off

src/corral/io.py
```diff
@@ -10,68 +11,90 @@
...
         except Exception as e:
-            raise RuntimeError(f"Error getting file info: {e}") from e
+            raise RuntimeError(f"Error getting file info for {path}: {e}") from e
 
     def copy_file(self, source: str, destination: str) -> None:
         try:
-            self.fs.copy(source, destination)
+            resolved_source = self._resolve_path(source)
+            resolved_dest = self._resolve_path(destination)
+            self.fs.copy(resolved_source, resolved_dest)
```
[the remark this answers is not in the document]
MrtinoRG: @n0w0f this is a good improvement I think

tasks/chembench/chembench/env.py
```diff
@@ -66,6 +66,19 @@
...
+
+    def generate(self, prompts: list[str], **_kwargs):
+        generations = []
+        for _prompt in prompts:
+            generation = None
+            generations.append([Generation(text=generation)])
+
+        return Generations(generations=generations)
+
+
```
MrtinoRG, line 79: Modify the MaCBench task as well so it works with this changes of having the Dummy Model class repeated around tasks

src/corral/utils.py
MrtinoRG, line 1: So, may I ask why you moved everything here? Now it is impossible to review it since I do not know what the changes are...

src/corral/base.py
n0w0f: ie. you mean delete if exist and then create directories ?

I would not want to add timestamps to directory names, atleast in trail/task directory level

src/corral/evaluate.py
n0w0f: here the checkpoints are saved to hard disk.

are you asking about the state ? Also it is highly unlikely that server goes down, only the benchmarking script would go down.

src/corral/utils.py
n0w0f: I did not, move anything new in here. 

I removed chembench dependency to chembench task alone, as it is a 
1. big package
2. is under constant developments
3. also it is not required by corral as such but the tasks. so i think tasks can have chembench as dependency.

Ideally I would also move vector database creation things to `scripts` but i want to ask you before doing it. But if it is robust and would not require constant developments it is good to have it in the framework as a feature.

tasks/chembench/chembench/env.py
n0w0f: Can i do it in one of my upcoming PR ?

MrtinoRG: Why?

n0w0f: Added it [here](https://github.com/lamalab-org/mat-agent-bench/pull/83/commits/6bf9b93df17a1fcdcc37a77aa4471a02a1a5f59d)

src/corral/utils.py
MrtinoRG: The file is completely sorted from scratch, if you check the PR file, you will see that all file has changed with respect to main because things are sorted diferently...

The creation vector database is used by several environment tools, what do you mean?

tasks/chembench/chembench/env.py
n0w0f: because changing branch now would need me to stash few things 🥲

src/corral/evaluate.py
MrtinoRG: I am talking about the results, the data for the completed tasks so at the end they are available to compute the metrics

src/corral/base.py
MrtinoRG: I observed that the agents tend to use the names of files that end up with appended stuff to the same files over and over again. 

So ideally for each of the runs, empty directories should be created. 

Also, why you do not want datetime there?

tasks/chembench/chembench/env.py
MrtinoRG: Ok, so then now I am closing this PR since all the files are in the new one #83

src/corral/base.py
n0w0f: datetime in directory name is creating issues for agents. as in more mistakes with path submissions

src/corral/evaluate.py
n0w0f: Ah okay. I can add that

tasks/chembench/chembench/env.py
n0w0f: No. Let us please merge this first.

src/corral/utils.py
n0w0f: I grouped things that are together and removed the `Model` thing i am also able to see the diff.

n0w0f: The creation vector database is used by several environment tools, what do you mean?
is everything there required ?

tasks/chembench/chembench/env.py
MrtinoRG: The other one points to this now

src/corral/base.py
MrtinoRG: Ok, I understand that, but perhaps you can add the ones with the trials inside one with datetime, so you avoid both problems

src/corral/utils.py
MrtinoRG: I mean that the vector database creation function is used by MaCBench, ChemBench and spetra elu. And yes, I think all is required there.

Why do you want to move them away? I think they do not have any heavy dependencies being imported...

src/corral/base.py
n0w0f: if the user want such a structure, they can define base path with date time when they execute.

I personally would not want to use complecated nesting here

tasks/chembench/chembench/env.py
n0w0f: I changed it back to main.  I dont want to do all feature in one PR it is as simple as that

src/corral/utils.py
n0w0f: > I mean that the vector database creation function is used by MaCBench, ChemBench and spetra elu. And yes, I think all is required there.

I understand that. I thought every utility there is not something that you need in different tasks

src/corral/base.py
```diff
@@ -184,8 +186,10 @@ def execute(self, **kwargs):
 class Environment(ABC):
     """Base class for task environments"""
 
-    def __init__(self, task_id: str):
+    def __init__(self, task_id: str, base_work_dir: str, chained_tasks: bool = False):
```
kjappelbaum: is `chained_tasks` really something that should and needs to be a `kwargs`?

src/corral/base.py
```diff
@@ -210,25 +214,35 @@ def reset_state(self) -> str:
...
-        timestamp = datetime.now(tz=timezone.utc).strftime("%m%d%H%M")
-        return f"{self.task_id}_{self.state.trial_id}_{timestamp}"
+    def _create_trial_workspace(self, trial_id: str) -> str:
+        """Create workspace directory for this trial"""
+        if self.chained_tasks:
+            # For chained tasks: one folder per trial (shared across all tasks in chain)
+            workspace = Path(self.base_work_dir) / f"trial_{trial_id}"
+        else:
+            # For independent tasks: one folder per task per trial
+            workspace = Path(self.base_work_dir) / f"{self.task_id}_trial_{trial_id}"
```
kjappelbaum: is separation of handling really needed?  Why can't we always default to the naming convention that makes foldernames maximally different?

kjappelbaum: especially if you just describe the issue as

> Issue : Each trail needs seperate folder where io will be executes, files from tools wil be saved, otherwise there could contamination from previous trails. Which means

you should be able to solve this with one form of logic

src/corral/evaluate.py
```diff
@@ -214,42 +153,254 @@ def bench(
...
+
+        if dependency_chain:
+            self._run_with_trial_level_checkpointing(
+                task_ids, trials_per_task, task_results, session_id, verbose
+            )
+        else:
+            self._run_with_task_level_checkpointing(
+                task_ids, trials_per_task, task_results, session_id, verbose
+            )
+
```
kjappelbaum: here, too: Isn't one version a superset of the other? If I do not have a dependency, isn't this just a special case of general solution for dependency chaining? 

I am a bit worried that we are adding more and more conditionals to the code

src/corral/evaluate.py
```diff
@@ -214,42 +153,254 @@ def bench(
...
+        }
+        self._save_checkpoint_file(session_id, checkpoint)
+
+    # Trial-level checkpoint methods
+    def _save_trial_level_checkpoint(
+        self, session_id: str, completed_trials: int, task_results: dict
+    ) -> None:
+        """Save trial-level checkpoint with results"""
+        checkpoint = {
+            "type": "trial_level",
```
kjappelbaum: I feel I must miss something: But I also do not feel that we should need different checkpoint types. 

Can't we just see it from metadata such trial/task counters?

src/corral/task.py
```diff
@@ -96,3 +99,33 @@ def check_dependencies_satisfied(self, task_id: str) -> bool:
             dep_task_id in self.results and self.results[dep_task_id] is not None
             for dep_task_id in task.input_from_tasks
         )
+
+
+class WorkspaceStrategy(Enum):
```
kjappelbaum: why isn't it enough to have one clean unique workdir per trial/task? If the chained 	subtasks require some info from another dir, it could also be copied over?

kjappelbaum:
I think as you saw from my comment I would perhaps benefit a bit from context on why you decided for this design - in particular, having different checkpointing folder approached for chained/unchained runs. While I can see that this code should do the job, i am unsure if this will give us the most maintainable code - now, one needs to keep in mind in multiple places if something is chained or not. At least from the basic abstraction point of view, one unique folder per/trial and task should satisfy the criterion of isolation.

Then, there might be a case where a chained tasks requires some data from a previous run -- wouldn't then then a copy not be an easier option (given that some places of the code anyhow need to already handle some transfer of state)? 

I am sure that you thought about questions like those and I would love to hear what the tradeoffs/thoughts you have been going through have been.

n0w0f:
> I am sure that you thought about questions like those and I would love to hear what the tradeoffs/thoughts you have been going through have been.

My rationale for this design emerged as i solved 2 task sequentially.  

I was first solving the order in which tasks should run.  For which i decided,
1. If there is tasks which are chained run all the tasks, then we run next trail for all the tasks. (Other way of  running trails for each subtask would require keeping track of  - outputs from which trails  should go to trails in subsequent task - this seemed complicated)
2. If there is no dependency we run all trails for a task and move to next one. If we would have been running this also in a similar manner to chained tasks, that is run all tasks for each trail, then there would be no need for a separation in how we should run, things would also be simpler.  ( But this would  mean it is not possible to get passK results for partially done tasks in unexpected cases of not able to benchmark completely. )

So when user runs the benchmark, they set in their env if tasks are chained along with other env variables. The environment takes this variable. And runner check this in the environment and run the benchmark accordingly

The second PR was to now resolve how temporary files should be saved. 

Since chained tasks (subtasks are part of one big tasks) one folder per trail seemed to be the easiest implementation. Io now sees all files in chained tasks and tasks needing files from previous task is also there. Also now if run this chained tasks as one task the directory structure works fine Other way would be we keep {task_id}trial{trial_id} for each task/trail that would mean we need functionality to copy files between the tasks as tasks progress and based on the dependency.

For the checkpointing the best option would be to `{task_id}_trial_{trial_id}`, but for chained task this would mean dependency manager now copies all the files from the previous tasks which we assume are still there and restarting trail from scratch for chained task seemed correct now because there could be affects from this intermediate restarting on result. So I restart from trail for chained tasks. 


Based on your comments, I agree with your concerns one coupling which i believe can be solved by unifying the naming itself
 If I follow a unified naming approach something like 
```
work_dir/
├── trial_0/
│   ├── retrieve_structure/
│   └── enumerate_slabs/
├── trial_1/
│   ├── retrieve_structure/
│   └── enumerate_slabs/

 ```
would remove the conditional inside the environment. Also now we need not pass `chained_tasks` tot he environment
But then we would have to pass it to the runner while initialization or through bench (assuming we run chained tasks differently)



How should we proceed ?

**Comment by n0w0f** [comment]
Based on the discussions from benches subgroup meeting (03-jun-2026),
Current design had issue which is now tried to be solved as below
1. Environment need not know the execution strategy.
 - this is now solved by removing `chained_task` variable from environment. Runner knows if tasks are are chained from `Taskgroup` via an endpoint. 
 
 2. No separate workspace naming required. 
 - Unified the naming, and hence the conditional naming in environment is also removed.
 
3. Complex conditional checkpointing: Two different strategies based on chaining
- removed  `_save_task_level_checkpoint()`, ` _save_trial_level_checkpoint()`, ` _run_with_task_level_checkpointing()`, ` _run_with_trial_level_checkpointing()`, ` checkpoint["type"] checking`.
- Introduced unified checkpoint saving and loading methods and `run_independent_execution()`, `_run_chained_execution()`

File: src/corral/base.py [hunk 10]
```diff
@@ -210,25 +213,28 @@ def reset_state(self) -> str:
...
-        Generate a combined identifier using task_id, trial_id, and a timestamp.
-
-        Returns:
-            A string combining task_id, trial_id, and timestamp in format: "{task_id}_{trial_id}_{timestamp}"
-        """
-        if not hasattr(self, "state") or self.state is None:
-            return f"{self.task_id}_no_trial_{datetime.now(tz=timezone.utc).strftime('%m%d%H%M')}"
+    def _create_trial_workspace(self, trial_id: str) -> str:
+        """Create workspace directory for this trial"""
+        workspace = Path(self.base_work_dir) / f"{self.task_id}_trial_{trial_id}"
```
kjappelbaum, line 230: would there be value in having timestamps? Or should this always be handled via the `base_work_dir`?

kjappelbaum (approved):
looks good. The new implementation seems much simpler, which I like!

src/corral/base.py
n0w0f: the reason for not having timestamp is because, agents are making lot of mistakes, when the directories or paths have complicated names

PR #83, opened 2025-05-28
feat: reporting - time, token, toolcall

n0w0f:
- Issue: It is nice to track the time taken for execution of tasks. Currently it could be available via Task State but not reportes

- Solution:
    - TaskState tracks time
    - Added a method to hit TaskTrail state endpoint in  Interface(soon to be renamed as router) 
    - MatAgentBenchmark (soon to be renamed as Runner/Orchestrator) track time and save it in TaskTrailResult, which is later adde3d to reports

## Summary by Sourcery

Add comprehensive execution time tracking and reporting throughout the benchmark pipeline, enrich reporting output with duration metrics, and introduce utilities for workspace isolation, trial state retrieval, and path resolution adjustments.

New Features:
- Track and record task and trial execution durations
- Provide an API to fetch individual trial state via get_trial_state

Enhancements:
- Compute and include total and average durations in summary tables and JSON reports
- Default TaskState timestamps to UTC and ensure end_time is set consistently
- Introduce WorkspaceStrategy enum for configurable workspace isolation
- Refine run_benchmark script defaults (model name, trials per task, k values)
- Unify path resolution to handle both absolute and relative paths
- Rename interface method get_task_prompt to get_task_guide
- Add stub Model class in macbench environment


src/corral/agents/tool_calling.py
```diff
@@ -122,7 +122,7 @@ def run_agent(
             interface.get_available_tools_for_task(task_id)
         )
         if task_prompt is None:
-            task_guide = interface.get_task_prompt(task_id)
+            task_guide = interface.get_task_guide(task_id)
```
MrtinoRG: I do not know why you did this change but is wrong. With this command also the tools are passed in the prompt. But for this agent we want the tools to be passed only as the function calling native service by the providers, so this change needs to be reverted

src/corral/base.py
```diff
@@ -95,6 +95,12 @@ def get_tool_statistics(self) -> dict[str, Any]:
             },
         }
 
+    def get_duration(self) -> float | None:
+        """Get trial duration in seconds"""
+        if self.end_time and self.start_time:
+            return (self.end_time - self.start_time).total_seconds()
+        return None
+
```
MrtinoRG: I understand this measures the total time of the trial, this is tool + LLM calls. If this is the case I think this needs to be changed. I think we agreed that what we will consider as execution time is tool execution time, since it has no sense to report API call times when you use non-local models

src/corral/report.py
```diff
@@ -42,6 +42,8 @@ class TaskTrailResult:
     score: float
     state: dict[str, Any]  # TODO replace Any with specific types
     tool_statistics: dict[str, Any]  # TODO replace Any with specific types
+    duration: float | None = None
+    token_usage: dict[str, Any] | None = None
```
MrtinoRG: Minor comment is that `token_usage` type should be dict[str, int] right?

src/corral/report.py
```diff
@@ -90,6 +93,103 @@ def all_results(self) -> list[TaskTrailResult]:
...
+                        if isinstance(value, (int | float)):
+                            total_tokens[key] = total_tokens.get(key, 0) + value
+                        elif isinstance(value, dict):
+                            if key not in total_tokens:
+                                total_tokens[key] = {}
+                            for sub_key, sub_value in value.items():
+                                if isinstance(sub_value, (int | float)):
+                                    total_tokens[key][sub_key] = (
+                                        total_tokens[key].get(sub_key, 0) + sub_value
+                                    )
```
MrtinoRG: I think the keys in the tokens key are:

```
{
  "prompt_tokens": <int>,
  "completion_tokens": <int>,
  "total_tokens": <int>
}
```

So I think that with your implementation we are counting the double of tokens than the reality. 

But also I think it can be valuable to preserve the three categories above since perhaps we can correlate something like more reasoning and thus better performance with a bigger amount of completion tokens for example.

src/corral/report.py
MrtinoRG, line 1: In this file, the functions to calculate the overall duration, and the ones for calculating the amount of tokens are mostly repeated code. We should keep in mind that there is a potential of code optimization in this file

src/corral/agents/tool_calling.py
n0w0f: this is a mistake, this is fixed now. i think i was trying something and forgot to revert

src/corral/base.py
n0w0f: why do you say it do not make sense ? some inference are faster and it is better to have fast api calls.
Also tool execution time only will not capture, the internal thoughts , time for the generation of thoughts and so on

MrtinoRG: @kjappelbaum has other thoughts I think

src/corral/report.py
n0w0f: We preserve the three categories here,
but anyways simplified since better typed now

src/corral/report.py
n0w0f: Optimized overall duration part now. I will leave more optimization to later refactoring stages.

src/corral/base.py
kjappelbaum: I think what one might want to measure at some point is if the agent is able to take resource-efficient actions. In a case where there are tools with different runtimes, but both could solve the task, it should take the faster one. 

The API call time is less relevant than that.

kjappelbaum: That is, ideally we could separately track the tool call time

src/corral/report.py
```diff
@@ -90,6 +93,83 @@ def all_results(self) -> list[TaskTrailResult]:
...
+                        else:
+                            failed_calls += 1
+
+        return {
+            "successful": successful_calls,
+            "failed": failed_calls,
+            "total": successful_calls + failed_calls,
+        }
+
     def average_score(self) -> float:
```
kjappelbaum, line 197: this method is almost too long. It combines computation with formatting of a table. I would separate those two things if possible.

src/corral/base.py
```diff
@@ -209,6 +215,8 @@ def save_current_state(self) -> TaskState:
     def reset_state(self) -> str:
         """Reset the environment state with a new trial id and fresh TaskState and return finished trail id."""
         if hasattr(self, "state") and self.state is not None:
+            if self.state.is_completed and self.state.end_time is None:
```
kjappelbaum: why do we also need to check that the end time is none?

src/corral/base.py
n0w0f: I would add this in tool calls, the duration for execution and later report a sum of this for a task

n0w0f: this is done now

src/corral/report.py
n0w0f: I modularized a bit. I would leave major refactoring for the refactiring stages if that is okay

src/corral/base.py
n0w0f: this logic is now changed, 
endtime was initialized as none and then updated after task is complete

kjappelbaum (approved):

PR #90, opened 2025-06-09
fix: correct scoring in the spectra tasks

MrtinoRG:
## Summary by Sourcery

Simplify and correct molecule similarity scoring in both spectra tasks by removing partial-match heuristics and awarding full points only for exact SMILES matches.

Bug Fixes:
- Fix scoring to return 1.0 only for exact matches and default to 0.0 otherwise.

Enhancements:
- Remove scaffold, fingerprint and rotatable bond–based similarity calculations and related RDKit imports.
- Add an early string-level equality check in the easy task to short-circuit exact matches.


n0w0f (approved):
looks good to me

PR #97, opened 2025-07-04
chore: add tool description with template

n0w0f:
Added tool description in tool template format

## Summary by Sourcery

Documentation:
- Add structured docstring templates with tagged sections (e.g., BRIEF, DETAILED, PROCEDURAL, CONTEXTUAL, WORKFLOW_INTEGRATION, SYNTACTICAL, RAISES, LIMITATIONS) to all tools in catalyst/tools.py.


tasks/catalyst/src/catalyst/tools.py
```diff
@@ -9,21 +9,80 @@
...
+    [CONTEXTUAL] How this tool works:
+    - Connects to Materials Project API using authentication key
+    - Searches for the specified material ID in the database
+    - Retrieves the pymatgen Structure object containing atomic positions and lattice parameters
+    - Converts the structure to CIF format string for compatibility with other tools
+    - Returns standardized crystallographic data suitable for further processing
+    [/CONTEXTUAL]
+
+    [WORKFLOW_INTEGRATION] Typical workflow integration:
+    1. [PREREQUISITE] Ensure MP_API_KEY environment variable is set with valid Materials Project API key [/PREREQUISITE]
```
kjappelbaum: is the environment variable something we want the agent to care about?

tasks/catalyst/src/catalyst/tools.py
```diff
@@ -9,21 +9,80 @@
...
+    2. [CURRENT] Apply this tool with a valid MP ID to retrieve bulk structure [/CURRENT]
+    3. [FOLLOW_UP] Use the CIF output with slab generation tools like enumerate_slabs_text to create slab structures [/FOLLOW_UP]
```
kjappelbaum, line 44: do you want to say that this is an example?

tasks/catalyst/src/catalyst/tools.py
```diff
@@ -9,21 +9,80 @@
...
+    - Use when you need to retrieve bulk crystal structures from the Materials Project database
+    - Best suited for materials with known MP IDs
+    - Usuall first step in simulation workflows
+    - Recommended for obtaining crystal structures for preparing bulk structures, supercells, bulk cells, slabs etc.
+    - Avoid when you need multiple structures
+    [/PROCEDURAL]
+
+    [CONTEXTUAL] How this tool works:
+    - Connects to Materials Project API using authentication key
+    - Searches for the specified material ID in the database
```
kjappelbaum: do you want to say how it can get the id?

tasks/catalyst/src/catalyst/tools.py
```diff
@@ -9,21 +9,80 @@
...
-        CIF content string.
+        str: [BRIEF] CIF content string containing the crystal structure data. [/BRIEF]
+             [DETAILED] A properly formatted CIF (Crystallographic Information File) string
+             containing all necessary information about the crystal structure including lattice
+             parameters, atomic positions, space group, and symmetry operations. This format
+             is widely compatible with crystallographic software and other structure analysis tools. [/DETAILED]
+             [EXAMPLES] Example output: "\n_chemical_formula_structural Si\n_cell_length_a 5.468..." [/EXAMPLES]
+
+    [RAISES] Exceptions:
+        ConnectionError: [ERROR_WHEN] When unable to connect to Materials Project API [/ERROR_WHEN]
```
kjappelbaum, line 70: is all the API key stuff something the agent needs to care about? Or shall we assume that the tool "works"?

tasks/catalyst/src/catalyst/tools.py
```diff
@@ -80,18 +226,96 @@ def enumerate_slabs_text(
...
+    [RAISES] Exceptions:
+        ValueError: [ERROR_WHEN] When CIF string is malformed or parameters are invalid [/ERROR_WHEN]
+                   [ERROR_DETAILS] Invalid CIF format, negative size parameters, or incompatible Miller indices [/ERROR_DETAILS]
+                   [ERROR_RECOVERY] Verify CIF format and ensure all parameters are positive numbers [/ERROR_RECOVERY]
+        StructureError: [ERROR_WHEN] When slab generation fails for the given structure [/ERROR_WHEN]
+                       [ERROR_DETAILS] Structure not compatible with specified Miller indices or size constraints [/ERROR_DETAILS]
+                       [ERROR_RECOVERY] Try different Miller indices or adjust size parameters [/ERROR_RECOVERY]
+    [/RAISES]
+
+    [LIMITATIONS] Known limitations:
```
kjappelbaum, line 314: shall we be clearer that this might generate really many structures?

tasks/catalyst/src/catalyst/tools.py
```diff
@@ -121,15 +345,82 @@ def enumerate_slabs_text(
...
+    [CONTEXTUAL] How this tool works:
+    - Parses the JSON string containing multiple slab structures
+    - Locates the slab with the specified index key (e.g., "slab_0", "slab_1")
+    - Extracts the CIF string for the selected slab
+    - Returns the CIF content ready for use in subsequent tools
+    [/CONTEXTUAL]
+
+    [WORKFLOW_INTEGRATION] Typical workflow integration:
+    1. [PREREQUISITE] First run enumerate_slabs_text to generate multiple slab terminations [/PREREQUISITE]
+    2. [CURRENT] Apply this tool to select a specific slab by index [/CURRENT]
```
kjappelbaum: how would i find which index to use? shall we be a bit more concrete?

kjappelbaum (approved):

tasks/catalyst/src/catalyst/tools.py
n0w0f: good point, I fixed this, now prerequisite is only to figure out if this is the right tool, ie we retrieve only one structure

tasks/catalyst/src/catalyst/tools.py
n0w0f: maybe `Typical workflow integration example:` is better

tasks/catalyst/src/catalyst/tools.py
n0w0f: now i mention, it would be part of user input, if not use other tools to figure out mp id based on information available

tasks/catalyst/src/catalyst/tools.py
n0w0f: I think this is usefull because, incase the connection error arise, it can switch to something else, rather than retrying this tool

tasks/catalyst/src/catalyst/tools.py
n0w0f: added info on possible ways

PR #101, opened 2025-07-07
feat: modify the code to handle the different levels of extended docstrings

MrtinoRG:
…ocstrings

## Summary by Sourcery

Enable structured docstring support by extracting named sections and allowing clients to request specific or combined sections for tool descriptions, and adapt parsing, server endpoints, client calls, and agent workflows accordingly.

New Features:
- Accept tool descriptions as a dict of named sections and default to the BRIEF section
- Expose a new get_tool_description_by_keyword method to retrieve specific or all docstring sections
- Allow clients to specify section keywords to concatenate when listing available tools

Enhancements:
- Update parse_docstring to return a sections dict instead of a plain string
- Change /tasks/{task_id}/tools to POST with a keywords list and adjust the evaluate client to use it
- Propagate a sections parameter through LLMPlanner and ToolCallingAgent to fetch detailed tool descriptions based on keywords

### DIFF
--- a/src/corral/agents/llm_planner.py
+++ b/src/corral/agents/llm_planner.py
@@ -1,11 +1,17 @@
-from promptstore import PromptStore
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from promptstore import PromptStore
+
+    from corral.evaluate import BenchmarkInterface
 
 from corral.agents.base_agent import BaseAgent
 from corral.agents.prompt_utils import create_prompt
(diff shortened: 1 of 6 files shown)


n0w0f:
@MrtinoRG I tried to start from this design and it seems to be coupled a lot and we move around information a lot, I feel for now we could do this by filtering based at the endpoint. 
So i started from scratch here #103 , with the intention to no touch base and agents.
But we can discuss and maybe consider this design when we think about the refactoring. For now I would close this , but we can reopen this later ?

PR #124, opened 2025-07-18
test: unit and integration for catalyst env

n0w0f:
## Summary by Sourcery

Apply path resolution to submitted answers in score methods and persist the resolved paths in result storage to ensure consistent scoring behavior

Bug Fixes:
- Use smart_resolve_path on submitted answers before scoring and storing results in both Catalyst and ML environments

Enhancements:
- Log the resolved answer path for each task submission


pyproject.toml
```diff
@@ -85,7 +85,7 @@ addopts = ["-p no:warnings", "--import-mode=importlib"]
 xfail_strict = true
 log_cli_level = "warn"
 pythonpath = "src"
-testpaths = ["tests"]
+testpaths = ["tests", "tasks/catalyst/tests"]
```
kjappelbaum: why do we need to do fancy discovery code (`find` for some filepath patterns) if we have a list here?

tasks/catalyst/pyproject.toml
```diff
@@ -1,10 +1,15 @@
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
 [project]
 name = "catalyst"
 version = "0.1.0"
 description = "Add your description here"
```
kjappelbaum, line 8: :)

kjappelbaum:
In https://github.com/lamalab-org/mat-agent-bench/pull/124/files#diff-af10f2cee3692a4fe4487be7375048bbbe03bf912285811f96e5571268d49911L196 the function name mentions CO2 but the docstring makes it appear as if it were for molecules in general

tasks/catalyst/src/catalyst/tool_utils.py
```diff
@@ -1,16 +1,27 @@
...
-from ase import Atoms
-from ase.constraints import FixAtoms
-from loguru import logger
-from pymatgen.analysis.local_env import VoronoiNN
-from pymatgen.core import Molecule, Structure
-from pymatgen.io.ase import AseAtomsAdaptor
-from pymatgen.symmetry.analyzer import SpacegroupAnalyzer
 
-# Utilities adopted from https://github.com/FAIR-Chem/fairchem
+def execute_python_code_given_code(code: str) -> str:
```
kjappelbaum: this could almost be a general tool for corral as it probably is useful in many different envs

tasks/catalyst/src/catalyst/tools.py
```diff
@@ -3,34 +3,93 @@
...
 from corral.utils import tool
 
 load_dotenv("../.env")
 
 
+# utility function
+def get_bulk_polymorphs_data_func(composition: str) -> str:
+    """
+    Query the Materials Project database to find polymorphs for a given composition. This function returns
+    a JSON string containing polymorph data including MP IDs, structures (CIF),  structures (CIF), energies above hull, formation_energy_per_atom, band gaps, densities,
```
kjappelbaum:
```suggestion
    a JSON string containing polymorph data including MP IDs,  structures (CIF), energies above hull, formation_energy_per_atom, band gaps, densities,
```

tasks/catalyst/src/catalyst/tools.py
```diff
@@ -3,34 +3,93 @@
...
+    volumes, number of sites, symmetry, and stability. The results are sorted by energy above hull.
+
+    Args:
+        composition: Chemical composition (e.g., 'TiO2')
+        api_key: Materials Project API key (optional if set in environment)
```
kjappelbaum, line 33:
```suggestion
```

tasks/catalyst/src/catalyst/score.py
```diff
@@ -178,8 +180,6 @@ def check_slab_structure(path_or_cif: str) -> float:  # TODO: better slab check.
     logger.info("check_slab_structure")
```
kjappelbaum: In principle, it would be cleaner if we would not log the function names as separate log statements but instead config the logger such that it would print the function it was called from

tasks/catalyst/src/catalyst/tool_utils.py
```diff
@@ -1,16 +1,27 @@
...
+        return json.dumps(
+            {"success": True, "execution_result": {"output": exec_locals.get("output")}}
+        )
+    except Exception as e:
+        return json.dumps(
+            {"success": False, "error": str(e), "traceback": traceback.format_exc()}
+        )
 
 
 def generate_output_capture_code() -> str:
```
kjappelbaum, line 4: why do you have this step with the preferred variables? If this is how we then extract results, this should be prominently mentioned in the docstrings of the code execution tools

kjappelbaum:
Do we really need `safe_convert_timeout`?  It just does some casting to int ...

kjappelbaum:
> in `parse_execution_output` how do we know that we have to start with line 17

it is to strip of 17 characters which is EXECUTION_RESULT

kjappelbaum:
```
                      [SYNTACTIC] Format: "Valid CIF format string with atomic coordinates and lattice parameters" [/SYNTACTIC]
                      [EXAMPLES] Examples: CIF string from get_structure_from_mp_text output [/EXAMPLES]
```

might be actually confusing. Perhaps I'd rather say "string in valid CIF syntax in `SYNTACTIC` and then also give an actual example and not reference another function

kjappelbaum:
in the tool docstrings: How are linebreaks actually handled in the prompt? It seems to me that, right now, the model would see linebreaks in quite random positions

kjappelbaum:
Another question for those prompt templates - Should we really mention any other tool outside from "Workflow integration". I feel that mentioning other tools in "details" or other fields makes the ablations more complex than needed.

For instance you mention things like ` (use get_structure_from_mp_text instead)`. I would prefer to minimize those things as this will break stuff as we rename and I think the ablations will be cleaner if we only mention the names of other tools in one place

kjappelbaum:
Sometimes you have quotation marks in `[SYNTACTIC] Format:`, sometimes not. I would make that consistent

kjappelbaum:
In one place you write "See the tool's source code for detailed format." -> can the agent even do that?

kjappelbaum:
An example like " [EXAMPLES] Example output: CIF string with reconstructed surface or JSON with multiple variants [/EXAMPLES]" is not really an example but a description ;)

kjappelbaum:
In `sort_and_get_first_from_json` the function name sounds quite general but the argument `polymorph_data_json` quite specific. Do you want to keep it this way? 
especially in this function a real example for the inputs would be quite useful (and not only the reference to another method)

kjappelbaum:
is something like ` Ensure MP_API_KEY is set and material ID is valid ` relevant for the agent?

kjappelbaum:
and, somehow, ` [EXAMPLES] Examples: 1` having "examples" twice in there is a bit redundant

tasks/catalyst/tests/test_path_resolution.py
```diff
@@ -0,0 +1,233 @@
+import os
+from pathlib import Path
+
+os.environ["CORRAL_WORK_DIR"] = str(Path(__file__).parent / "test_files")
```
kjappelbaum, line 4: isn't this something one could solve with `conftest.py` and some fixture in there?

tasks/catalyst/tests/test_path_resolution.py
```diff
@@ -0,0 +1,233 @@
...
+    os.environ["CORRAL_WORK_DIR"] = str(test_files_dir)
+    yield test_files_dir
+    if original:
+        os.environ["CORRAL_WORK_DIR"] = original
+    else:
+        os.environ.pop("CORRAL_WORK_DIR", None)
+
+
+def test_debug_ci_environment():
+    """Debug test to see what files exist in CI"""
```
kjappelbaum: should this test actually be run? Or be should we move this into some folder with helper methods?

tasks/catalyst/tests/test_path_resolution.py
```diff
@@ -0,0 +1,233 @@
...
+
+# --- Tests for extract_path_from_answer ---
+def test_extract_path_from_answer_basic_filename():
+    assert extract_path_from_answer("filename.txt") == "filename.txt"
+
+    def test_extract_path_from_answer_with_prefix():
+        assert (
+            extract_path_from_answer("final answer is path/to/file.json")
+            == "path/to/file.json"
+        )
```
kjappelbaum, line 35: why are those two test functions inside each other?

tasks/catalyst/tests/test_environment_integration.py
```diff
@@ -0,0 +1,958 @@
...
+from catalyst.score import check_mp_structure, check_slabs_json, check_valid_json_file
+from hypothesis import given
+from hypothesis import strategies as st
+
+from corral.task import TaskDefinition, TaskGroup
+
+TEMP_DIR = Path(os.environ["CORRAL_WORK_DIR"])
+
+
+class TestTaskGroupEnvironment:
```
kjappelbaum, line 26: I realize that the `TaskGroupEnvironment` is defined in `catalyst.env` - but isn't this something that would apply to all of corral?

tasks/catalyst/tests/test_environment_integration.py
```diff
@@ -0,0 +1,958 @@
...
+            # Check that workspace changed
+            assert env.current_work_dir != initial_workspace
+            assert isinstance(trial_id, str)
+
+            # Check that file tools still work
+            assert "list_files" in env.tools
+            assert env.tools["list_files"] is not None
+
+
+class TestTaskSystemIntegration:
```
kjappelbaum, line 491: this again seems like a test that is not only specific to `catalyst` but rather to a core piece of corral

tasks/catalyst/tests/test_environment_integration.py
```diff
@@ -0,0 +1,958 @@
...
+
+                env.state.submitted_answer = submission_text
+                score = env.score()
+
+                # Score should always be a float between 0 and 1
+                assert isinstance(score, float)
+                assert 0.0 <= score <= 1.0
+
+
+class TestPathResolutionInScoring:
```
kjappelbaum, line 702: shouldn't the path resolution be used beyond the `catalysis` environment? That is, should this be in this subpackage?

tasks/catalyst/tests/test_tools.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+def skip_if_no_api_key():
+    """Skip test if MP_API_KEY is not available"""
+    return pytest.mark.skipif(
+        not os.getenv("MP_API_KEY"), reason="MP_API_KEY not available in environment"
+    )
+
+
+def rehydrate_docs(docs_as_dicts):
```
kjappelbaum, line 35: is "rehydration" a common term?

tasks/catalyst/tests/test_tools.py
```diff
@@ -0,0 +1,211 @@
...
+@skip_if_no_api_key()
+def test_get_structure_from_mp_text():
+    cif_str = get_structure_from_mp_text.execute(mp_id="mp-149")
+    assert isinstance(cif_str, str)
+    assert "data_Si" in cif_str
+    struct = Structure.from_str(cif_str, fmt="cif")
+    assert struct.composition.reduced_formula == "Si"
+
+
+@settings(deadline=5000, suppress_health_check=[HealthCheck.too_slow])
```
kjappelbaum, line 110: why do you go for fuzz-testing here? 
that is, do you expect it to be flaky for some inputs?

tasks/catalyst/tests/test_scoring.py
```diff
@@ -0,0 +1,389 @@
...
+    check_slabs_json,
+    check_valid_json_file,
+)
+from hypothesis import given
+from hypothesis import strategies as st
+
+TEST_SCORE_FILES = Path(os.environ["CORRAL_WORK_DIR"])
+
+
+class TestBasicFileValidation:
```
kjappelbaum, line 25: this seems again more general than just the catalysis env

kjappelbaum (approved):
looks overall good. I have some minor comments about: 

- prompt formatting (whitespace, references to other tools) 
- moving general tools/methods to core of `corral`

pyproject.toml
n0w0f: Removing this from main toml. As the task package need to have their own environment for testing, this would not work.

tasks/catalyst/src/catalyst/tool_utils.py
n0w0f: this util function is now removed

n0w0f:
> Do we really need `safe_convert_timeout`? It just does some casting to int ...

update the name now

n0w0f:
> In one place you write "See the tool's source code for detailed format." -> can the agent even do that?

removed this now

n0w0f:
> is something like `Ensure MP_API_KEY is set and material ID is valid` relevant for the agent?

fixed this now

tasks/catalyst/tests/test_path_resolution.py
n0w0f: i will save this for the refactoring stage

tasks/catalyst/tests/test_path_resolution.py
n0w0f: removed this

tasks/catalyst/tests/test_environment_integration.py
n0w0f: i will move to the framework as we move TaskGroupEnvironment to corral

tasks/catalyst/tests/test_environment_integration.py
n0w0f: it is relevant to all TaskGroupEnvironment

tasks/catalyst/tests/test_tools.py
n0w0f: it is used in webdev

n0w0f:
Fixed the line breaks, and removed all the tool names mentioned outside workflow integration

PR #139, opened 2025-08-01
chore: reviewing the task descriptions

n0w0f:
## Summary by Sourcery

Reorganize catalyst task configs by grouping existing files into chained and single subdirectories and adding new chained catalyst configurations

New Features:
- Add chained_cu20.json and chained_tio2.json configuration files under chained

Enhancements:
- Move and rename the original chained.json to tasks/catalyst/config/chained/chained_si.json
- Move single.json into tasks/catalyst/config/single/single.json


tasks/catalyst/config/single/single.json
MrtinoRG, line 1: For the other envs now we have a file for each env and task. Do you think that you can implement this in this one?

tasks/catalyst/config/single/single.json
```diff
@@ -1,7 +1,7 @@
...
   "si_workflow": {
     "name": "Complete CO2 Adsorption on Silicon Slab Workflow",
-    "description": "Complete workflow to create a CO2 adsorption structure on a Silicon slab. This involves: 1) Retrieving the Si bulk structure from Materials Project, 2) Enumerating possible slabs with Miller index (1,1,1), 3) Choosing a suitable slab, 4) Retrieving the CO2 molecule structure (mp-20066), 5) Identifying adsorption sites on the slab, 6) Choosing an appropriate adsorption site (preferably top), and 7) Adding the CO2 molecule to the slab at the chosen site with a height of approximately 2.0 Å. Submit the path to the final combined structure CIF file.",
+    "description": "Create a CO2 adsorbed structure on a Silicon slab. This involves: 1) Retrieving the Si bulk structure from Materials Project, 2) Enumerating possible slabs with Miller index, 3) Choosing a suitable slab, 4) Retrieving the CO2 molecule structure, 5) Identifying adsorption sites on the slab, 6) Choosing an appropriate adsorption site (preferably top), and 7) Adding the CO2 molecule to the slab at the chosen site with a height of approximately 2.0 Angstrom. Submit the path to the final combined structure CIF file.",
```
MrtinoRG: I think this is not fair at all since this is way more description than the other envs. I would remove the steps, and if something is strictly needed, pass it as an input parameter

tasks/ml/config/chained/chained_nitride.json
```diff
@@ -0,0 +1,94 @@
+{
+  "batch_retrieve_nitride_polymorphs": {
+    "name": "Batch Retrieve Nitride Polymorphs",
+    "description": "Generate a diverse list of nitride compositions for dataset creation. Retrieve polymorphs for all nitride compositions from Materials Project database. For each composition, collect multiple polymorphs including both stable and metastable structures. Consolidate all individual polymorph files into a single comprehensive dataset. Add composition-based features and validate data quality including completeness, duplicate detection, and target variable distribution. Ensure the dataset is suitable for machine learning.",
```
MrtinoRG: And here it is the same as the previous comment. If you provide the steps in the prompt, it has no sense. I would split this subtask even more, like retrieving one task, if some nitrile is missing 0.0; second task to create the dataset, if data is missing 0.0; third task validate the data...

tasks/ml/config/chained/chained_nitride.json
```diff
@@ -0,0 +1,94 @@
...
+        "subsample": 0.8,
+        "colsample_bytree": 0.8,
+        "reg_alpha": 0.1,
+        "reg_lambda": 0.1
+      }
+    }
+  },
+  "evaluate_model_performance": {
+    "name": "Evaluate Model Performance",
+    "description": "Comprehensively evaluate the trained XGBoost model using test set metrics, cross-validation. Assess model performance and save results as JSON file. The file should have two things mandatory `test_set_evaluation` and `cross_validation_results` use the same key name when saving results for this.",
```
MrtinoRG: "The file should have two things mandatory `test_set_evaluation` and `cross_validation_results` use the same key name when saving results for this." is very confusing, I do not even understand what you are trying to ask. 

What I understand is those of both keys should be in the final JSON with the final key name (?)

tasks/ml/config/chained/chained_oxide.json
MrtinoRG, line 1: Same comments as in the other file:

- To much description
- Could be further splitted
- Last task is not clear at all

tasks/ml/config/single/single.json
MrtinoRG, line 1: Like, you have more description in the subtasks than in here, despite I think that if you provide the steps in the prompt it has no sense anymore since the planning is gone

MrtinoRG: Also the part of the key names is not clear to me, I understand that is to use the exact the same keys that you provide but the way you express it is super confusing

tasks/catalyst/config/single/single.json
n0w0f: i would like to keep all task in one json because it is easier to run all of them together, and also get metrics all of them together.

what is the benefit of keeping  separate json for each task

n0w0f:
@MrtinoRG I have addressed your comments now, can you take look again please

tasks/ml/config/chained/chained_nitride.json
n0w0f: yes, those should be json key names i have updated those now

MrtinoRG (approved):

PR #145, opened 2025-08-03
fix: manage bools in the parsing of the react agent

### LINKED ISSUE #144: React regex fails to capture Action sometimes
for eg in this case
`
response = "Thought: I have successfully retrieved the polymorph data for the composition Fe2O3. Now, I need to select the polymorphs using a strategy that aligns with the task requirements. I will use the \"diverse_energy\" strategy to select the polymorphs and save the results to a file.\nAction: select_polymorphs_with_strategy_to_file\nAction Input: {\n    \"polymorphs_data\": \"/Users/n0w0f/ml_gpt_react/ml_oxides_trial_0/Fe2O3_polymorphs.json\",\n    \"save_path\": \"/Users/n0w0f/ml_gpt_react/ml_oxides_trial_0/selected_Fe2O3_polymorphs.json\",\n    \"selection_strategy\": \"diverse_energy\",\n    \"max_polymorphs\": 3,\n    \"energy_threshold\": 0.3,\n    \"is_path\": True\n}"`

MrtinoRG:
## Summary by Sourcery

Bug Fixes:
- Convert Python-style True/False to lowercase JSON true/false when parsing LLM responses

### DIFF
--- a/src/corral/agents/react.py
+++ b/src/corral/agents/react.py
@@ -165,11 +165,6 @@ class ReActAgent(BaseAgent):
         )
 
         for _iteration in range(self.max_iterations):
-            # # add a delay of 5 second to avoid hitting rate limits
-            # import time
-
-            # time.sleep(10)
-
             # Create prompt and get LLM response
             llm_response = self.get_llm_response().content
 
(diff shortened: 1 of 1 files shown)


src/corral/agents/react.py
```diff
@@ -121,6 +121,9 @@ def parse_llm_response(
             tool_name = action_match.group(1).strip()
             try:
                 action_input = action_match.group(2).strip()
+                action_input = action_input.replace("True", "true").replace(
```
kjappelbaum, line 124: wouldn't a more robust way be to first case to fully lowercase/fully uppercase and then handle it?

PR #155, opened 2025-08-25
chore: restructure and rename

n0w0f:


n0w0f:
@kjappelbaum, @MrtinoRG , I have now renamed some classes and also restructure the repo to what we discussed at some point.

src/corral/agents/base_agent.py
```diff
@@ -13,7 +13,7 @@
     llm_call,
     save_agent_messages,
 )
-from corral.evaluate import BenchmarkInterface
+from corral.router.routes import CorralRouter as TypeRouter
```
kjappelbaum: if this is only imported for type checking, why not 

```
import typing

if typing.TYPE_CHECKING:
    import CorralRouter
```

and then use `"CorralRouter"` as type annotation. 
Having a different name increases the cognitive load to me

kjappelbaum: but would also love to hear the perspective of @MrtinoRG

kjappelbaum (commented):
there seem to now be two `tool_utils` functions - this can lead to confusion.

src/corral/router/verbosity.py
kjappelbaum, line 1: why is this in the `router` folder?

kjappelbaum (commented):
another question I have is if you want to make the folders subpackages by adding `__init__.py` files

kjappelbaum:
There is a bit too much going on now for me to review every single change. 
But the basic structure now looks ok to me and from my point of view we could move ahead. I added some queries I had.

kjappelbaum (approved):

src/corral/agents/base_agent.py
n0w0f: using TYPE_CHECKING seems to be the better option for runtime.
updating this.

n0w0f:
> there seem to now be two `tool_utils` functions - this can lead to confusion.

you mean, `backend.tool_utils` and `utils.tool_utils` ?

I can maybe change `utils.tool_utils` to `utils.tool_helpers`

src/corral/router/verbosity.py
n0w0f: Right now it is ordered by logic. So i have kept everything related to communication between agent and environment and processing of communication in routes.
Another option would be to put it in backend, but according to me I feel backend need not know about any  run  time setting like verbosity

n0w0f:
> another question I have is if you want to make the folders subpackages by adding `__init__.py` files

I have now made subpackages, the api also feels better.

```

from corral import CorralRunner, CorralRouter
from corral.agents import ReActAgent, ToolcallingAgent
from corral.backend import Environment

```

n0w0f:
merging this now, we can start a separate PR for any related development

PR #158, opened 2025-08-31
feat: add the Q&A for spectra

MrtinoRG:
Hi @kjappelbaum, 

I am still reviewing all the questions 1 by 1 to check that there are not to very similar fragments. But I think anything else should be correct.

## Summary by Sourcery

New Features:
- Add Q&A task JSON files for spectra QA


MrtinoRG:
And would add value adding IR or MS? Like for the small fragments, obviously the multiplicity is not going to be a thing, but there the formula and/or functional groups can vary between the possibilities, that is why I am asking about the two other spectra...

reports/qa/generate_qa_spectra.py
```diff
@@ -0,0 +1,339 @@
...
+    """
+    Parse an already-expanded, dot-free formula into {element: count}.
+    """
+    counts = Counter()
+    for el, cnt in _ELEMENT_PAT.findall(formula):
+        counts[el] += int(cnt or 1)
+    return counts
+
+
+def _expand_parentheses(formula: str) -> str:
```
kjappelbaum: I think many of those methods have been implemented before in another context. Would it be useful to maintain them in one place?

reports/qa/generate_qa_spectra.py
```diff
@@ -0,0 +1,339 @@
...
+                for frag in frags:
+                    # Filter tiny pieces
+                    if (
+                        sum(1 for a in frag.GetAtoms() if a.GetAtomicNum() > 1)
+                        < min_heavy_atoms
+                    ):
+                        continue
+
+                    try:
+                        # Try to sanitize the fragment with ionic charges
```
kjappelbaum: where are the charges here? Is this handled by rdkit?

reports/qa/generate_qa_spectra.py
```diff
@@ -0,0 +1,339 @@
...
+        fragments = return_possible_fragments(smile)
+        fragment_counter = 0
+        for fragment in fragments:
+            task = {}
+            molecular_formula = CalcMolFormula(Chem.MolFromSmiles(fragment))
+            isomers = get_isomers.remote(formula=molecular_formula, limit=3)
+            if len(isomers) < 3:
+                continue
+            sampled_isomers = isomers
+            target_scores = {isomer: 0 for isomer in sampled_isomers}
```
kjappelbaum: for your small molecules this might not be the case, but you might be unlucky in general that some isomers have identical 1H and 13C spectra. This could be, e.g., due to some heteroatoms.
In principle, you would need to check that the spectrum of the isomer you picked here is different from the one of the actual compound

reports/qa/generate_qa_spectra.py
```diff
@@ -0,0 +1,339 @@
...
+            target_scores[fragment] = 1
+            items = list(target_scores.items())
+            random.shuffle(items)
+            target_scores = dict(items)
+            h_nmr = get_h_nmr.remote(smiles=smile)
+            c_nmr = get_c_nmr.remote(smiles=smile)
+            examples = [
+                {
+                    "input": f"What of the four fragments is more likely to match the next C-NMR and H-NMR spectra?\n\nH-NMR: {h_nmr}\n\nC-NMR: {c_nmr}",
```
kjappelbaum: I am not sure how biasing it is to show a potentially large isomer and a small fragment as the option. 

The other option might be to randomly pick a fragment of comparable size of another molecule and ensure that this fragment is not in the molecule.

MrtinoRG: The idea is:

- To have a set of questions of different difficulty. For the small fragments, the changes would be bigger, in the sense of introducing different functional groups or atom types. In bigger fragments, the coupling should be the main focus when determining the correct option.
- Because of the idea for small fragments, I think that maybe adding IR and MS results in the question input might be helpful, and would do the question more fair with the models.

So the question is, should IR and perhaps MS or other info be added to the prompts?

MrtinoRG: And yes, after I have an answer for this, I will go by hand through all generated questions to avoid gray options

reports/qa/spectra_qa/tasks_json/task_00100/task_00100.json
```diff
@@ -0,0 +1,24 @@
+[
+  {
+    "name": "Method MCQ task 1",
+    "uuid": "7ca407bf-de7f-58b1-957a-385ad100100",
+    "description": "Determine which analysis method is more appropriate for getting some certain information about a molecule given its spectra.",
+    "examples": [
+      {
+        "input": "Which analysis method is more appropriate for determining if a molecule contains a carbonyl group (C=O)?",
```
kjappelbaum: "most" appropriate

even though "appropriate" is perhaps not super clear. Perhaps we can use something like. "Gives the clearest and unique signal .."

reports/qa/spectra_qa/tasks_json/task_00100/task_00101.json
```diff
@@ -0,0 +1,24 @@
+[
+  {
+    "name": "Method MCQ task 1",
+    "uuid": "7ca407bf-de7f-58b1-957a-385ad100101",
+    "description": "Determine which analysis method is more appropriate for getting some certain information about a molecule given its spectra.",
+    "examples": [
+      {
+        "input": "Which analysis method is more appropriate for determining if a molecule contains groups that tend to form hydrogen bonds (O-H, N-H, etc.)?",
```
kjappelbaum: I think I have the same comment here about "more appropriate". 


1H NMR would also give you this. 

In both 1H and IR it depends on how "wet" your sample is. Thus, this question might not always be true. 

But perhaps you can ask differently - where do you expect characteristic broad peaks and then IR and 1H-NMR would be true

reports/qa/spectra_qa/tasks_json/task_00100/task_00102.json
```diff
@@ -0,0 +1,24 @@
+[
+  {
+    "name": "Method MCQ task 1",
+    "uuid": "7ca407bf-de7f-58b1-957a-385ad100102",
+    "description": "Determine which analysis method is more appropriate for getting some certain information about a molecule given its spectra.",
+    "examples": [
+      {
+        "input": "Which analysis method is more appropriate for determining if a molecule contains aromatic carbons?",
+        "target": null,
+        "target_scores": "{'UV-Vis': 0, 'COSY': 0, 'MS': 0, 'NMR-13C': 1}"
```
kjappelbaum: UV vis is not much worse than 13C NMR. You will have characteristic adsorptions 

Perhaps I would ask about what is best to distinguish between two substituted benzene derivatives. for this, some form of 2D would be best

reports/qa/spectra_qa/tasks_json/task_00100/task_00103.json
```diff
@@ -0,0 +1,24 @@
+[
+  {
+    "name": "Method MCQ task 1",
+    "uuid": "7ca407bf-de7f-58b1-957a-385ad100103",
+    "description": "Determine which analysis method is more appropriate for getting some certain information about a molecule given its spectra.",
+    "examples": [
+      {
+        "input": "Which analysis method is more appropriate for determining if a molecule contains halides (C-X)?",
```
kjappelbaum: perhaps more -> most

reports/qa/spectra_qa/tasks_json/task_00100/task_00104.json
```diff
@@ -0,0 +1,24 @@
+[
+  {
+    "name": "Method MCQ task 1",
+    "uuid": "7ca407bf-de7f-58b1-957a-385ad100104",
+    "description": "Determine which analysis method is more appropriate for getting some certain information about a molecule given its spectra.",
+    "examples": [
+      {
+        "input": "Which analysis method is more appropriate for determining the connectivity of fragments in a molecule?",
```
kjappelbaum, line 8: looks good!

kjappelbaum:
I left some comments on the methods Q/A, but otherwise it looks good for me and is ready to go

reports/qa/spectra_qa/tasks_json/task_00100/task_00102.json
MrtinoRG: I now change the true options of this question, and I added a new one with the question that you suggested

MrtinoRG:
I am merging this since the changes were simply implementing the suggestions

PR #164, opened 2025-09-03
task with first order kinetics

chandan21gupta:
Kinetic Modelling Environment
Some functions are still missing, will be implemented once data is available.

## Summary by Sourcery

Introduce a comprehensive kinetic modeling environment including tools for building reaction networks, deriving rate laws, generating and solving ODE systems, fitting and validating kinetic parameters, plus a corral-based task framework and supporting utilities.

New Features:
- Add corral tools for kinetic modeling: setup_reaction_network, derive_rate_law, generate_ode_system, fit_kinetic_parameters, validate_model_fit, and lookup_previous_works

Enhancements:
- Implement reaction parsing, rate law generation, ODE system construction, and parameter fitting utilities
- Introduce FittingModel and optimization pipelines using SciPy for kinetic parameter estimation
- Create KineticEnvironment and task definitions for corral-based kinetic modeling benchmarks

Build:
- Add pyproject.toml with project metadata and dependencies


tasks/kinetic_modeling/kinetic_modeling/tasks.py
MrtinoRG, line 1: @kjappelbaum would you please show light on the subtasks, because they would be only the obtain rate law first and second fit the system. But this seems too poor

tasks/kinetic_modeling/kinetic_modeling/tools.py
```diff
@@ -0,0 +1,737 @@
...
+
+    except Exception as e:
+        # Handle any other errors that might occur during validation
+        return json.dumps(
+            {"error": "Model validation failed", "message": str(e), "success": False}
+        )
+
+
+@tool
+def lookup_previous_works(query: str, top_k: int = 5) -> str:
```
MrtinoRG, line 565: @kjappelbaum do you have the papers that you mentioned at hand to build the dataset?

tasks/kinetic_modeling/kinetic_modeling/utils/parameter_fitting.py
```diff
@@ -0,0 +1,578 @@
...
+            "margin_error": margin_error,
+            "confidence_level": (1 - alpha) * 100,
+            "degrees_of_freedom": degrees_of_freedom,
+            "t_critical": t_critical,
+        }
+
+    return confidence_intervals
+
+
+def load_experimental_data(config: str) -> dict:
```
MrtinoRG, line 506: Right now the experiments are uploaded in HF under the train split. Each config is one experiment. I took the ones from here: https://github.com/jschneidewind/hte-streamlit/tree/main/data/Raw_Data 

I do not know if these are the correct ones :\

tasks/kinetic_modeling/kinetic_modeling/tasks.py
```diff
@@ -0,0 +1,149 @@
...
+                            "derive_rate_law",
+                            "generate_ode_system",
+                            "fit_kinetic_parameters",
+                            "validate_model_fit",
+                        ],
+                        scoring_fn=score_parameters,
+                        scoring_inputs={"k1": 0.025},
+                        submission_format="Return a dictionary with the fitted parameters, e.g., {'k1': 0.0}",
+                        input_from_tasks=[],
+                        initial_input={"experiment": "results_mrg_059_zn_12_2"},
```
MrtinoRG, line 30: The experiment right now is the same for all the tasks, ideally this should change, but right now it is a bit difficult to know which experiment is what :\

tasks/kinetic_modeling/kinetic_modeling/utils/ode_generator.py
kjappelbaum, line 1: I think if we took code from somewhere else, we should mention in the docstring of the module where we took it from

tasks/kinetic_modeling/kinetic_modeling/utils/ode_generator.py
```diff
@@ -0,0 +1,181 @@
+from collections.abc import Callable
+
+import numpy as np
+from loguru import logger
+from scipy.integrate import odeint
+
+
+def build_ode_system(
```
kjappelbaum, line 8: I think at least for the ablations where we do not want to have chemical names in there, we perhaps should expose less chemical argument names to the agent

MrtinoRG: So you suggest to change the descriptions and names such that there is not mention to chemistry, but do this only for the ablations

tasks/kinetic_modeling/kinetic_modeling/utils/rate_laws.py
```diff
@@ -0,0 +1,137 @@
...
+        dict: Dictionary containing rate expression, parameters, and units
+    """
+    if mechanism == "elementary":
+        return _generate_elementary_rate_law(reaction)
+    elif mechanism == "michaelis_menten":
+        return _generate_michaelis_menten_rate_law(reaction)
+    elif mechanism == "langmuir_hinshelwood":
+        return _generate_langmuir_hinshelwood_rate_law(reaction)
+    else:
+        raise ValueError(f"Unknown mechanism: {mechanism}")
```
kjappelbaum, line 24: this can either be implemented or go

kjappelbaum: i mean you have methods for this below - not sure if they should perhaps be used here

tasks/kinetic_modeling/kinetic_modeling/env.py
```diff
@@ -0,0 +1,212 @@
...
+Develop a kinetic model that accurately describes the experimental observations. Your final model should include:
+1. A reaction mechanism (if mechanism discovery is required)
+2. Rate laws for each reaction
+3. Fitted kinetic parameters
+4. Statistical validation of the model fit
+
+**Success Criteria:**
+- Model achieves R² > 0.95 on experimental data
+- Physical reasonableness of parameters
+- Statistical validation passes standard tests
```
kjappelbaum, line 67: is this always the case? or should we rather move those details into the task description?

tasks/kinetic_modeling/kinetic_modeling/env.py
```diff
@@ -0,0 +1,212 @@
...
+
+Required submission format:
+{self.current_task.submission_format}
+
+"""
+
+        prompt += "\nAvailable input data:\n"
+
+        # Display input data from dependencies
```
kjappelbaum, line 76: the code below is repeated many times across environments. Perhaps we can move it into some reusable place?

tasks/kinetic_modeling/kinetic_modeling/tasks.py
kjappelbaum: i think just doing the fit (as here) would be a good subtaks

kjappelbaum: Perhaps we need to consider if we stick with the different levels or want to go via subtasks as some of the first levels I proposed would only have very simple subtasks

but from the level one where you need to distinguish between n possible mechanisms, you could make meaningful subtasks

tasks/kinetic_modeling/kinetic_modeling/tasks.py
kjappelbaum: but do you mean with "difficult to know which experiment is what"?

tasks/kinetic_modeling/kinetic_modeling/tools.py
```diff
@@ -0,0 +1,737 @@
...
+            {
+                "error": "Literature search failed",
+                "message": str(e),
+                "query": query,
+                "success": False,
+            }
+        )
+
+
+def create_tools() -> dict[str, Tool]:
```
kjappelbaum, line 728: I am wondering if we should give the agent a tool to have a look at the data. This could either be via some relevant summary statistics or it could be via some sort of ASCII plot.

but it would be cool to know about local minima/maxima, how long the time series is, ...

kjappelbaum:
Another tool we probably need to have is something that reports the concentration over different species over time. This would help the agent the debug the fitting process.

tasks/kinetic_modeling/kinetic_modeling/utils/ode_generator.py
kjappelbaum: What I suggest is that we should build tools such that we can do those ablations where we do not mention chemistry.  Ideally, we would then, e.g, move the core implementation to one function and then just have two different wrappers -- one with chemical descriptions and one with general descriptions -- that call the core method

kjappelbaum:
I think one tool to get some basic description of the time series for the agents could be based around https://tsfel.readthedocs.io/en/latest/

tasks/kinetic_modeling/kinetic_modeling/tools.py
kjappelbaum: The one for the system we looked at with Jacob would be https://pubs.acs.org/doi/full/10.1021/acscatal.6b02595

tasks/kinetic_modeling/kinetic_modeling/utils/parameter_fitting.py
kjappelbaum: I think one part we would need to also do somehow is to provide the system with context about the experiment (light intensity, initial concentrations, ...)

tasks/kinetic_modeling/kinetic_modeling/env.py
```diff
@@ -0,0 +1,212 @@
...
+from corral.backend.server import run_server
+from corral.utils.task_group import TaskGroup
+
+CORRAL_WORK_DIR = os.environ["CORRAL_WORK_DIR"]
+BASE_WORK_DIR = os.environ.get(
+    "CORRAL_WORK_DIR", "../CORRAL_WORK_DIR/spectra_elucidation"
+)
+
+
+class KineticEnvironment(Environment):
```
n0w0f, line 17: Why not use `TaskGroupEnvironment` from utils ?

PR #167, opened 2025-09-04
wetlab initial commit

aaaghajani:
The initial commit for the WetLab environments. The structure is very similar to the `spectra_elucidation` tasks.
`engine.py` contains the crude initial version of the equilibrium solver engine.
`env.py` is mostly copied from `spectra_elucidation/spectra_elucidation/env.py` with minor changes to adapt.
`tools.py` contains a conceptual description of tools (nothing implemented yet).
`score.py` only contains the overall scoring by matching the list of proposed ions with ground truth.
An example task (task_0) is there but subtasks are not yet implemented.

### DIFF
--- a/tasks/wetlab/wetlab/engine.py
+++ b/tasks/wetlab/wetlab/engine.py
@@ -1,0 +1,131 @@
+from dataclasses import dataclass
+from typing import Dict
+import reaktoro as rk
+
+class VolumeError(ValueError):
+    pass
+
+db = rk.Database.fromFile('WetChem.yaml')
+
+aq_phase = rk.AqueousPhase('H2O H+ OH- NO3- Cl- I- K+ '
+                           'NH3 NH4+ '
+                           'H2S HS- S-2 '
+                           'Ag+ Ag(NH3)+ Ag(NH3)2+ AgCl2- '
+                           'Cu+2 Cu(NH3)+2 Cu(NH3)2+2 Cu(NH3)3+2 Cu(NH3)4+2 '
+                           'Mn+2 Pb+2 ')
+solids = [
+    'AgCl(s)', 'AgI(s)',
+    'PbCl2(s)', 'PbI2(s)',
+    'Ag2S(s)', 'CuS(s)', 'MnS(s)', 'PbS(s)',
+    'Cu(OH)2(s)', 'Mn(OH)2(s)', 'Pb(OH)2(s)'
+]
+solid_phases = [rk.SolidPhase(solid) for solid in solids]
+
+DEFAULT_SYS = rk.ChemicalSystem(db, aq_phase, *solid_phases)
+ZERO = 1e-25
+
+solver = rk.EquilibriumSolver(DEFAULT_SYS)
+options = rk.EquilibriumOptions()
+options.epsilon = ZERO
+solver.setOptions(options)
+
+def _new_empty_state(sys=DEFAULT_SYS):
+    state = rk.ChemicalState(sys)
+    state.setSpeciesAmounts(ZERO)
+    return state
+
+def _species_names(sys=DEFAULT_SYS):
+    for sp in sys.species():
+        yield sp.name()
+
+def _copy_state(state):
+    new = _new_empty_state(state.system())
+    for name in _species_names(state.system()):
+        mol = state.speciesAmount(name)
+        if mol > ZERO:
+            new.setSpeciesAmount(name, mol, "mol")
+    return new
+
+@dataclass
+class StockSolution:
+    composition: Dict[str, float]
+    description: str=None
+    
+    def __post_init__(self):
+        for sp, C in self.composition.items():
+            if C < 0:
+                raise ValueError(f"Negative concentration for {sp}: {C}")
+    
+    def __rmul__(self, V_mL: float) -> "Solution":
+        return Solution(self.composition, volume=V_mL)
+
+@dataclass(kw_only=True)
+class Solution(StockSolution):
+    volume: float
+
+    def __post_init__(self):
+        if self.volume <= 0:
+            raise VolumeError("Solution volume must be positive!")
+        
+        state = _new_empty_state()
+        V_L = self.volume/1000
+        state.setSpeciesMass("H2O", V_L, "kg")
+
+        for sp, C in self.composition.items():
+            mol = C * V_L
+            state.setSpeciesAmount(sp, mol, "mol")
+
+        object.__setattr__(self, "state", state)
+
+    def __rmul__(self, V_mL: float) -> "Solution":
+
+        if V_mL > self.volume + 1e-8:
+            raise VolumeError(f"Requested to draw {V_mL:.2f} mL but only {self.volume:.2f} mL available.")
+        
+        self.volume -= V_mL
+        return Solution(self.composition, volume=V_mL)
+    
+    def __add__(self, other: "Solution") -> "Solution":
+        if not isinstance(other, Solution):
+            raise NotImplemented
+        if self.state.system().id() != other.state.system().id():
+            raise ValueError("Cannot mix solutions from different chemical systems!")
+        
+        new_composition = {}
+        new_volume = self.volume + other.volume
+
+        for sp in _species_names():
+            self_mol = self.state.speciesAmount(sp) if self.state.speciesAmount(sp) > ZERO else 0.0
+            other_mol = other.state.speciesAmount(sp) if other.state.speciesAmount(sp) > ZERO else 0.0
+            if (sp not in ['H2O']+solids) and (total_mol := self_mol + other_mol) > 0: 
+                new_composition[sp] = 1000*float(total_mol)/new_volume
+        
+        return Solution(new_composition, volume=new_volume)
+    
+    def equilibrate_and_filter(self) -> tuple["Solution", Dict[str, float]]:
+
+        state = _copy_state(self.state)
+        solver.solve(state)
+
+        precipitates = {}
+        for solid in solids:
+            mol = state.speciesAmount(solid)
+            if mol > ZERO:
+                precipitates[solid] = 1000*float(mol)
+        
+        new_composition = {}
+        for sp in _species_names():
+            mol = state.speciesAmount(sp)
+            if (sp not in ['H2O']+solids) and mol > ZERO:
+                new_composition[sp] = 1000*float(mol)/self.volume
+        
+        return Solution(new_composition, volume=self.volume), precipitates
+
+
+
+
+
+
+
+            
+            
--- a/tasks/wetlab/wetlab/score.py
+++ b/tasks/wetlab/wetlab/score.py
@@ -1,0 +1,60 @@
+ALL_IONS = [
+    # Common cations
+    'Ag+',
+    'Al3+',
+    'Ba2+',
+    'Ca2+',
+    'Cd2+',
+    'Co2+',
+    'Cr3+',
+    'Cs+',
+    'Cu2+',
+    'Fe2+',
+    'Fe3+',
+    'Ga3+',
+    'Hg2+',
+    'Hg2^2+',
+    'K+',
+    'Li+',
+    'Mg2+',
+    'Mn2+',
+    'Na+',
+    'NH4+',
+    'Ni2+',
+    'Pb2+',
+    'Rb+',
+    'Sn2+',
+    'Sr2+',
+    'Zn2+',
+
+    # Common anions
+    'Br-',
+    'Cl-',
+    'CN-',
+    'CO3^2-',
+    'CrO4^2-',
+    'F-',
+    'I-',
+    'NO2-',
+    'NO3-',
+    'OH-',
+    'PO4^3-',
+    'S2-'
+    'SO3^2-',
+    'SO4^2-',
+    'SCN-',
+]
+
+def score_ion_list(prediction: str, ground_truth: str) -> float:
+
+    pred_list = prediction.split(", ")
+    gt_list = ground_truth.split(", ")
+
+    for ion in gt_list:
+        if ion not in ALL_IONS:
+            raise ValueError(f"Invalid ion in the ground truth: {ion}")
+        
+    if set(pred_list) == set(gt_list):
+        return 1.0
+    else:
+        return 0.0
--- a/tasks/wetlab/wetlab/tasks_json/task_0.json
+++ b/tasks/wetlab/wetlab/tasks_json/task_0.json
@@ -1,0 +1,30 @@
+[
+    {
+        "id": "qual_inorg_task_000",
+        "name": "task_000",
+        "keywords": [
+            "qualitative analysis",
+            "inorganic chemistry",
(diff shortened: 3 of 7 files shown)


kjappelbaum:
A minor point: We use `pre-commit` for consistent code (https://github.com/lamalab-org/mat-agent-bench/tree/main?tab=readme-ov-file#development-setup). That is, in the end, all the formatting and linting should pass one your local machine and then also the continuous integration on GitHub should not raise an error.

tasks/wetlab/wetlab/tasks_json/task_0.json
```diff
@@ -0,0 +1,28 @@
+[
+    {
+        "id": "10_15227_orgsyn_084_0077",
```
kjappelbaum: I think you'd still need to adjust names and keywords.

tasks/wetlab/wetlab/tasks_json/task_0.json
```diff
@@ -0,0 +1,28 @@
...
+        "keywords": [
+            "spectra",
+            "spectra-elucidation",
+            "experimental-spectra"
+        ],
+        "metrics": [
+            "binary"
+        ],
+        "input": {
+            "prompt": "Analyze the provided organic compound sample in a lab environment and output the SMILES string, while minimizing resource consumption due to the costly nature of the process.",
```
kjappelbaum: does SMILES make sense for the ions you try to identify here? Will we run into parsing issues? 

Other options: Strings and we sanitize ourselves, multiple-choice.

tasks/wetlab/wetlab/engine.py
```diff
@@ -0,0 +1,131 @@
+from dataclasses import dataclass
+from typing import Dict
+import reaktoro as rk
+
+class VolumeError(ValueError):
+    pass
+
+db = rk.Database.fromFile('WetChem.yaml')
```
kjappelbaum, line 8: this file does not appear in this PR for me. 

In any case, if we would like to "ship" it with the package, it should be included in package data/manifest. Otherwise, it will probably not be packaged by default when we build a Python package (e.g., for PyPi)

tasks/wetlab/wetlab/engine.py
```diff
@@ -0,0 +1,131 @@
...
+
+aq_phase = rk.AqueousPhase('H2O H+ OH- NO3- Cl- I- K+ '
+                           'NH3 NH4+ '
+                           'H2S HS- S-2 '
+                           'Ag+ Ag(NH3)+ Ag(NH3)2+ AgCl2- '
```
kjappelbaum, line 13: How is "Ag(NH3)2+" parsed? as `[Ag(NH3)2]` or we do not need to care because it does not matter in the end for `rk`?

tasks/wetlab/wetlab/engine.py
```diff
@@ -0,0 +1,131 @@
+from dataclasses import dataclass
+from typing import Dict
+import reaktoro as rk
+
+class VolumeError(ValueError):
+    pass
+
+db = rk.Database.fromFile('WetChem.yaml')
+
```
kjappelbaum, line 9: not sure if you also want to care about the following species: 

- MnO, MnO2
- PbO
- Ag2O, AgOH
- CuO
- Cu2O 
- AgCl3-2
- CuCl (and the other chlorides of copper)
- Lead chlorides/iodides

(some of them might need some heating to form)

kjappelbaum: I think I still have some books about the "Trennungsgang" in my office. They have nice flow charts of the relevant species - perhaps we should double check with them.

kjappelbaum: Otherwise, feel free to also order a copy of Jander Blasius/Gerdes for the office. We might need to look stuff up there more frequently if we continue along this line

tasks/wetlab/wetlab/engine.py
```diff
@@ -0,0 +1,131 @@
...
+        for sp, C in self.composition.items():
+            mol = C * V_L
+            state.setSpeciesAmount(sp, mol, "mol")
+
+        object.__setattr__(self, "state", state)
+
+    def __rmul__(self, V_mL: float) -> "Solution":
+
+        if V_mL > self.volume + 1e-8:
+            raise VolumeError(f"Requested to draw {V_mL:.2f} mL but only {self.volume:.2f} mL available.")
```
kjappelbaum, line 83: this is cool!

tasks/wetlab/wetlab/engine.py
```diff
@@ -0,0 +1,131 @@
...
+        V_L = self.volume/1000
+        state.setSpeciesMass("H2O", V_L, "kg")
+
+        for sp, C in self.composition.items():
+            mol = C * V_L
+            state.setSpeciesAmount(sp, mol, "mol")
+
+        object.__setattr__(self, "state", state)
+
+    def __rmul__(self, V_mL: float) -> "Solution":
```
kjappelbaum, line 80: minor thing is that in Python, following PEP8, you would not use this capitalization for your volume variables

tasks/wetlab/wetlab/tasks_json/task_0.json
kjappelbaum, line 1: do you think there is the possibility/value in having subtasks?

tasks/wetlab/wetlab/score.py
```diff
@@ -0,0 +1,60 @@
...
+    'SCN-',
+]
+
+def score_ion_list(prediction: str, ground_truth: str) -> float:
+
+    pred_list = prediction.split(", ")
+    gt_list = ground_truth.split(", ")
+
+    for ion in gt_list:
+        if ion not in ALL_IONS:
```
kjappelbaum, line 54: do you perhaps need to sanitize the ion (strip whitespace, lower case for consistency?)

tasks/wetlab/wetlab/score.py
```diff
@@ -0,0 +1,60 @@
...
+            raise ValueError(f"Invalid ion in the ground truth: {ion}")
+        
+    if set(pred_list) == set(gt_list):
```
kjappelbaum, line 57: I wonder if you could perhaps even have a method that returns a continuous reward (something like Hamming loss) and is then only binarized for the current version. This could either be done using a `threshold` argument, in a second function, ...

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,103 @@
...
+
+@tool
+def possible_ions() -> str:
+    """
+    Returns the list of all possible unknown ions that can be present in the sample.
+    """
+    pass
+
+@tool(hidden_args=['compositions'])
+def measure_ph(compositions, solution_name: str) -> str:
```
kjappelbaum, line 12: One thing that we have been doing in `corral` is to have very verbose tool descriptions in a structured format. Those are still missing here

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,103 @@
...
+    """
+    Measures the pH of the solution using universal indicator pH paper.
+    Costs: 0 mL
+    Returns: the closest integer to the actual pH of the solution
+    """
+    pass
+
+@tool(hidden_args=['compositions'])
+def perform_flame_test(compositions, solution_name: str) -> str:
```
kjappelbaum, line 21: how would you handle this? Just return the color or the wavelengths?

tasks/wetlab/wetlab/score.py
kjappelbaum, line 1: in the longer run, there might perhaps also be value in tracking where the system would do the test. For example, with special safety equipment or at least in the fume hood. We could also try to prioritize "safe" tests over stuff like Marsh test

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,103 @@
...
+    pass
+
+@tool(hidden_args=['compositions'])
+def perform_reagent_test(compositions, solution_name: str, reagent_name: str, solution_vol: int, reagent_vol: int, label: str) -> str:
+    """
+    Mix the solution and the reagent with the given volumes, call it <label> and add it to the inventory
+    Costs: <solution_vol> (to prevent the agent from using very small volumes, there is a minimum cap on this)
+    Returns: a list of observations ("nothing happens" if no observation).
+    -Examples:
+        "a black precipitate forms"
```
kjappelbaum, line 57: for getting the strings with observations, would you use templates or use an LLM to convert it into natural text? Or would you return some structured output you could easily extract from `rk`?

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,103 @@
...
+        reagent_name: H2S_HCl   composition: 0.1 M H2S, 0.01 M HCl
+        etc.
+    """
+    pass
+
+@tool(hidden_args=['compositions'])
+def perform_reagent_test(compositions, solution_name: str, reagent_name: str, solution_vol: int, reagent_vol: int, label: str) -> str:
+    """
+    Mix the solution and the reagent with the given volumes, call it <label> and add it to the inventory
+    Costs: <solution_vol> (to prevent the agent from using very small volumes, there is a minimum cap on this)
```
kjappelbaum, line 54: enhancement (for the future) might be that we also keep in mind that the minimum amount might depend on the test (some tests are more sensitive than others -> we would make it fail with a certain probability that is dependent on the amount the system draws)

tasks/wetlab/wetlab/engine.py
kjappelbaum: and they would also have ions that "disturb" certain tests and this would help us to double check that there is nothing we miss

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,103 @@
...
+
+@tool
+def lookup_solid_color(color: str) -> str:
+    """
+    Returns the list of all possible precipitates with color <color>
+    """
+    pass
+
+@tool
+def lookup_solution_color(color: str) -> str:
```
kjappelbaum, line 84: what is certainly perhaps a bit unrealistic here is that the strength of how well one perceives the color depends on the concentration. As future enhancement (not relevant for the first iteration) one could use Lambert-Beer to approximate this

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,103 @@
...
+    -Example:
+        sample                  -                   remaining: 46 mL
+        test_1      (1 mL NH4Cl + 4 mL sample)      remaining: 1 mL
+        test_2      (1 mL KI + 4 mL test_1)         remaining: 5 mL
+        etc.
+    """
+    pass
+
+@tool
+def lookup_solid_color(color: str) -> str:
```
kjappelbaum, line 77: i think those lookups could be very cool for ablations. 

Perhaps I would also consider a lookup with RAG in something like the course notes on qualitative inorganic analysis

kjappelbaum: or perhaps even https://www.emgywomenscollege.ac.in/templateEditor/kcfinder/upload/files/Vogel%27s%20Textbook%20Of%20Macro%20And%20SemiMicro%20Qualitative%20Inorganic%20Analysis%205th%20ed%20-%20G.Svehla.pdf

kjappelbaum: I used a lot the stuff here http://ruby.chemie.uni-freiburg.de/Vorlesung/nebenfach_praktikum.html

tasks/wetlab/wetlab/tools.py
aaaghajani: Yes, I'm aware. This is just a preliminary list of the functions :)

tasks/wetlab/wetlab/tasks_json/task_0.json
aaaghajani: I did make those adjustments... somehow ended up pushing the wrong file :(
will amend the commit

tasks/wetlab/wetlab/tasks_json/task_0.json
aaaghajani: same as above

tasks/wetlab/wetlab/engine.py
aaaghajani: same as above

tasks/wetlab/wetlab/engine.py
aaaghajani: When creating a phase, it just parses the species names as is and tries to find a species with the same "name" in the given database. So, as long as we are consistent between the `.yaml` and `engine.py` we're good.

tasks/wetlab/wetlab/tools.py
aaaghajani: yes, and I was also thinking of using hexcode colors so we can report the "overall solution color" by mixing them

tasks/wetlab/wetlab/tools.py
aaaghajani: for the flame test I was thinking of returning just the color name. for the solution color maybe we can use hexcode?

tasks/wetlab/wetlab/tasks_json/task_0.json
aaaghajani: amended the commit

tasks/wetlab/wetlab/tasks_json/task_0.json
aaaghajani: amended the commit

tasks/wetlab/wetlab/engine.py
aaaghajani: pdf is available for Jander-Blasius :D

aaaghajani: We need to decide the scope of reactions we care about because that would determine the thermodynamic data we need in the .yaml file. We also need to decide how realistic we want to get. Currently, temperature and pressure are constant, i'm assuming ideal behavior for everything, and the volume of all solutions is just a function of the mass of water. The more realistic we want to get, the more reliable thermodynamic data we would need. I would also stay away from redox reactions for now.

tasks/wetlab/wetlab/tasks_json/task_0.json
aaaghajani: for sure! we can guide the agent through the systematic way of doing things but let it make its own conclusions

aaaghajani: that would only work for the classic "find everything" tasks tho. It's more tricky for IChO-like puzzles

tasks/wetlab/wetlab/score.py
aaaghajani: Sure, and I think IoU is better here since we don't care about true negatives

tasks/wetlab/wetlab/tools.py
aaaghajani: structured output from `rk` turned into text:
It'll be easy to check the state before and after the test. So we can check what formed and what disappeared and turn those into observation statements.

PR #191, opened 2025-10-22
fix: correct the tool description problem

MrtinoRG:
## Summary by Sourcery

Normalize and enhance tool docstring conventions by introducing explicit ARGS_* and RETURNS_* tags and updating the verbosity router to support these new sections

Enhancements:
- Standardize function docstrings across tools by replacing generic [BRIEF], [DETAILED], [SYNTACTIC], [EXAMPLES] tags with ARGS_* and RETURNS_* sections for arguments and return values
- Extend the router verbosity configuration to recognize and filter new ARGS_ and RETURNS_ docstring sections

Documentation:
- Update ML, Catalyst, and Spectra Elucidation tools to use new ARGS_* and RETURNS_* tags in docstrings


src/corral/router/verbosity.py
kjappelbaum, line 115: I'm not sure I would write the code this way.

I would probably use a stricter regex, especially for the keyword sections themselves. For example, I'd use a [lookbehind](https://www.regular-expressions.info/lookaround.html) to ensure that nothing appears before `BRIEF`, `DETAILED`, etc., except perhaps whitespace or newlines.

Then I would separately match the Args and Returns sections.

With this approach, I wouldn't need any if statements. Instead, I would return three dictionaries: one for the keywords, one for the args, and one for the returns. I would then validate that they are consistent with each other.

tests/test_verbosity.py
```diff
@@ -0,0 +1,477 @@
...
+FULL_ARG_MP_ID = """[ARGS_BRIEF] Materials Project identifier string. [/ARGS_BRIEF]
+[ARGS_DETAILED] The unique identifier used by Materials Project to catalog materials.
+Should be in the format "mp-XXXXX" where XXXXX is a numerical ID.
+This ID corresponds to a specific material entry in the Materials Project database. [/ARGS_DETAILED]
+[ARGS_SYNTACTIC] "mp-" followed by digits (e.g., "mp-149", "mp-20066") [/ARGS_SYNTACTIC]
+[ARGS_EXAMPLES] "mp-149" (Silicon), "mp-20066" (CO2), "mp-2" (Li) [/ARGS_EXAMPLES]
+"""
+
+
+class TestFilterToolDescription:
```
kjappelbaum, line 201: why did you structure the tests in classes?

tests/test_verbosity.py
```diff
@@ -0,0 +1,477 @@
...
+        assert result == "No description available"
+
+        result = VerbosityConfig.filter_tool_description(None, ToolVerbosity.WORKFLOW)
+        assert result == "No description available"
+
+    def test_docstring_without_tags(self):
+        """Test filtering docstring without special tags"""
+        simple_docstring = "This is a simple tool that does something useful."
+        result = VerbosityConfig.filter_tool_description(
+            simple_docstring, ToolVerbosity.BRIEF
```
kjappelbaum, line 341: should we also verify this behavior for other verbosities?

tests/test_verbosity.py
```diff
@@ -0,0 +1,477 @@
...
+
+        # None is returned as-is by the function
+        result = VerbosityConfig.filter_argument_description(
+            None, ToolVerbosity.WORKFLOW
+        )
+        assert result is None or result == ""
+
+    def test_simple_arg_no_tags(self):
+        """Test argument without special tags"""
+        simple_arg = "A simple parameter description"
```
kjappelbaum, line 441: should we also test the case where you have `Args` but not with the suffix for verbosity level?

tests/test_verbosity.py
```diff
@@ -0,0 +1,477 @@
...
+
+
+class TestEdgeCases:
+    """Test edge cases and error conditions"""
+
+    def test_malformed_tags(self):
+        """Test handling of malformed tags"""
+        malformed = "[BRIEF] Some text [DETAILED] Missing close tag"
+
+        # Should handle gracefully
```
kjappelbaum: should we really handle this gracefully? I feel we would be better off if the parser complained here and we would just throw and exception if the syntax is not what we expect

kjappelbaum: I am actually surprised because `_keyword_regex` is relatively harsh (requiring the closing group, which I like)

tests/test_verbosity.py
MrtinoRG: Organization such that the tests for the parsing of the arguments are located and separated from the ones about the entire docstring

src/corral/router/verbosity.py
MrtinoRG: I am not sure if I am fully following you.

For the lookbehind, I guess you would look that is the beginning of the line, because in one take into account other lines only for BRIEF applies that there is nothing before.

And then yeah, I guess for the Args and Returns one can have that such that the code is simplified but it is not that much

tests/test_verbosity.py
MrtinoRG: I added that below in the edge cases, with this cases in which Args and Returns do not have keywords

tests/test_verbosity.py
MrtinoRG: So you would raise an Exception if one of the keywords is missing? But that would not work for the "basic" level, i.e., docstring without keywords.

I guess that perhaps what you have in mind is to raise an Exception if only the opening keyword appears and not the closing. Yeah, I think this is a good suggestion.

Now I implemented this such that when only an opening tag is found, and not the closing counterpart, an Exception is raised 👍

MrtinoRG: And I also added some tests for checking that these exceptions are raised :))

MrtinoRG (commented):
Thanks for the review @kjappelbaum!

I think I have now addressed all your comments and this should be ready for merging

tests/test_verbosity.py
```diff
@@ -0,0 +1,606 @@
...
+`execute_python_script("processing.py", None, 600, None)`,
+[/SYNTACTICAL]
+
+Args:
+    script_path: [ARGS_BRIEF] Path to the Python script file to execute. [/ARGS_BRIEF]
+                [ARGS_DETAILED] Complete file path to the Python script that should be executed.
+                The script must exist and be readable.
+                The path can be relative to the current working directory or absolute.
+                The script should be a valid Python file with appropriate shebang or run using the Python interpreter. [/ARGS_DETAILED]
+                [ARGS_SYNTACTIC] "Valid file path to Python script" [/ARGS_SYNTACTIC]
```
n0w0f:
```suggestion
                [ARGS_SYNTACTICAL] "Valid file path to Python script" [/ARGS_SYNTACTICAL]
```

Shouldnt this be ARGS_SYNTACTICAL ?

n0w0f (approved):

tests/test_verbosity.py
MrtinoRG: Yes, you were right 🙃

PR #194, opened 2025-10-23
Update AFM environment

imandal98:


n0w0f:
@imandal98 Can you install the pre commit hook we have, and make sure it is passing

PR #196, opened 2025-10-24
Surface energy level 1 runs

MrtinoRG:


MrtinoRG:
@chandan21gupta Can you remove the context7 and the react file from the PR? Also you have a lot of conflicts with main, is this expected?

chandan21gupta:
> @chandan21gupta Can you remove the context7 and the react file from the PR? Also you have a lot of conflicts with main, is this expected?

Have removed them. Other files are needed as I debugged the environment later. Also have removed the runs since we will be running everything again.

PR #206, opened 2025-11-01
feat: agents can surrender from doing a task

n0w0f:
## Summary by Sourcery

Allow agents to retire from unsolvable tasks by introducing an enable_retire flag, retiring via a new API endpoint, and recording retirement in results and reports.

New Features:
- Enable agents to retire from a task by responding with a retire signal
- Add a new backend API endpoint and router method to handle task retirement

Enhancements:
- Propagate an enable_retire flag through execution, agent, and trial routines
- Extend prompt generation to include retirement instructions when enabled
- Track retirement in TaskTrialResult and update report summaries and tables with retired trial statistics


src/corral/agents/tool_calling.py
```diff
@@ -131,12 +133,21 @@ def run(
         else:
             task_guide = task_prompt
 
+        # Prepare retirement instructions if enabled
+        retire_instructions = ""
+        if enable_retire:
+            retire_instructions = (
+                "If you cannot solve the task or determine it is unsolvable, you can retire from it.\n"
+                'To retire, respond with: "Final Answer: RETIRE"'
+            )
```
MrtinoRG: It is not very clean to add these instructions here right? It would be clearer to add them at the top of the table or using the promptstore

MrtinoRG (commented):
Personally, I do not really like this solution. 

I thought (and understood from the meeting) in a tool to surrender, but I guess it would be too much overhead at this point.

src/corral/agents/base_agent.py
```diff
@@ -226,6 +227,7 @@ def run(
             history (list[LiteLLMMessage], optional): The history items to include. Defaults to None.
             task_prompt (str, optional): The task prompt to use. Defaults to None.
             examples (list[str], optional): List with the few-shot examples to use. Defaults to None.
+            enable_retire (bool, optional): Whether to enable the retire option. Defaults to False.
```
kjappelbaum: I am wondering about the naming - do you think `retire` is the clearest naming? I would find something with clearer association to "giving up" better

n0w0f: changing `retire` to `Forfeit`

src/corral/agents/tool_calling.py
n0w0f: Updated now to use promptstore

n0w0f:
@kjappelbaum and @MrtinoRG I have now incorporated the comments.
It would be good to merge this soon.

src/corral/agents/base_agent.py
```diff
@@ -254,6 +271,7 @@ def run_agent(
             examples (list[str], optional): List with the few-shot examples to use. Defaults to None.
             verbose (bool, optional): Whether to save agent messages. Defaults to False.
             tool_verbosity (str, optional): The verbosity level for tool information. Defaults to "brief".
+            enable_forfeit (bool, optional): Whether to enable the forfeit option. Defaults to False.
```
kjappelbaum: I think this should explain what the option is. Otherwise, it only repeats the argument name, makling the docstring less effective

src/corral/agents/prompt_utils.py
```diff
@@ -78,6 +80,8 @@ def create_prompt(
         user_prompt: The user prompt object
         task_guide (Union[str, list]): The task guide or prompt to use
         history (list[LiteLLMMessage], optional): Message history to include. Defaults to None.
+        forfeit_prompt: The forfeit prompt object. Defaults to None.
+        enable_forfeit (bool, optional): Whether to include forfeit instructions. Defaults to False.
```
kjappelbaum: here, too, I'd explain what it means

src/corral/agents/react.py
```diff
@@ -157,6 +165,7 @@ def run(
             history (List[Dict[str, Any]], optional): The history items to include. Defaults to None.
             task_prompt (str, optional): The task prompt to use. `task_prompt` is intended to be a plan or description about the task, that should always be provided when this agent is called as a subagent of a main orchestrator. Defaults to None.
             examples (List[str], optional): List with the few-shot examples to use. Defaults to None.
+            enable_forfeit (bool, optional): Whether to enable the forfeit option. Defaults to False.
```
kjappelbaum: here, too, I'd explain what it means

src/corral/agents/tool_calling.py
```diff
@@ -75,11 +78,13 @@ def __init__(
         system_prompt: str | Any | None = None,
         user_prompt: str | Any | None = None,
         extractor_prompt: str | Any | None = None,
+        forfeit_prompt: str | Any | None = None,
         temperature: float = 0.7,
         prompt_store: PromptStore | None = None,
         system_prompt_id: str = "400fcecf-f5f2-464b-aff5-8a4377c9685c",
         user_prompt_id: str | None = "fe04453b-5469-4611-bba6-6d81487df787",
         extractor_prompt_id: str | None = "9d37e4a0-26c5-438a-ba1b-a273388fcded",
+        forfeit_prompt_id: str | None = "b2afa52c-ce88-4cb1-93bf-657a7c7b5933",
```
kjappelbaum: to simplify it, I'd like us to go with either str or id and not have arguments for both. 

This should be its own PR, but I think that this should happen.

src/corral/backend/env.py
```diff
@@ -361,6 +362,15 @@ def submit_answer(self, answer: str) -> float:
             self.state.end_time = datetime.now(tz=timezone.utc)
         return score
 
+    def forfeit(self) -> float:
+        """Forfeit from the current task without submitting an answer"""
+        self.state.forfeited = True
+        self.state.score = 0.0
```
kjappelbaum: should this always be the case? If we allow partial scoring, this would cause problems. Isn't the indication that the agent gave up enough?

src/corral/backend/env.py
```diff
@@ -361,6 +362,15 @@ def submit_answer(self, answer: str) -> float:
...
+        self.state.is_completed = True
```
kjappelbaum: what do we use `is_completed` for? 

This might be a sub-optimal naming - I guess it is rather `is_attempted`? Because with and without giving up it might not really "complete"/"solve" a task

src/corral/backend/server.py
```diff
@@ -183,6 +183,30 @@ def submit_answer(task_id: str, answer: dict):
...
+    @app.post("/tasks/{task_id}/forfeit")
+    def forfeit_task(task_id: str):
+        """Forfeit from a task without submitting an answer"""
+        if task_id not in environments:
+            raise HTTPException(status_code=404, detail="Task not found")
+
+        env = environments[task_id]
+
+        # 1. Forfeit and get score (0 for now, can be decided to be something else later)
+        score = env.forfeit()
```
kjappelbaum: I think it should just be the score that it has at this point

src/corral/backend/server.py
```diff
@@ -183,6 +183,30 @@ def submit_answer(task_id: str, answer: dict):
...
+        completed_trial = env.get_completed_trial_data()
+
+        # 3. Reset for next trial
+        finished_trial_id = env.reset_state()
+
+        return {
+            "score": score,
+            "state": completed_trial["state"],
+            "trial_id": finished_trial_id,
+            "forfeited": True,
```
kjappelbaum: It is bad that this key is sometimes in there and sometimes not. 

I believe we could keep it cleaner if we could have a private method `finalize/finish` (or something like that) that would handle state reset and the stuff that always needs to happen in the end. Also, if we'd have an object/namedtuple and not a dict being returned we would not start having problems like that

src/corral/run.py
```diff
@@ -71,8 +72,28 @@ def execute_single_trial(
...
+        if answer == "GIVE UP":
+            try:
+                result = interface.forfeit_task(task_id)
+                result.token_usage = token_usage
+                return result
+            except Exception as forfeiture_error:
+                return TaskTrialResult(
+                    task_id=task_id,
+                    trial_id=f"attempt_{trial_index + 1}",
+                    score=0.0,
```
kjappelbaum: why do you still return in the case of the exception? 

I was first only worried here because you manually set the score in multiple places, which can make it easy to introduce bugs. 

but I also do not fully understand why we cannot simply raise

kjappelbaum (commented):
i'm still not the biggest fan of the naming. I'd have found something like `surrender` almost more intuitive. Perhaps some other options here https://www.thesaurus.com/browse/gave-up (chicken out? ;)) 

I have no big worries about the main functionality, but I have a worry about some choices that might impact maintainability.

src/corral/run.py
n0w0f: The main motivation was that the benchmark would continue running even if individual trail run into issues.

But also because, I was not thinking long term (possibility of partial score within task), since currently this is sort of implemented via task group.
Currently , also for agent errors and submission errors, I similarly return TaskTrialResult object with score=0. (those errors could be frequent and i think continued benchmarking has value there.) I also followed a similar design  here.


But now i am also not liking the idea of setting score=0. maybe for all these cases i can retrieve from server the current score, instead of setting it in the exception block (also to avoid inconsistency).
 Or do you prefer to raise here ?
I think we go for retrieving from server, that can be done for all three in a separate PR ?
what do you think? @kjappelbaum

n0w0f:
> i'm still not the biggest fan of the naming. I'd have found something like `surrender` almost more intuitive. 

Surrender is nice, I will update to this.

src/corral/backend/env.py
n0w0f: I will update this to `is_attempted `  ([1f56042](https://github.com/lamalab-org/corral/pull/206/commits/1f5604290fcb77752937a33722e7fab9879b359c)), this is basically to check if some tasks are benchmarked already.

src/corral/backend/server.py
n0w0f: True updated this to receive the current score. receive 0 only if no score is assigned yet.

src/corral/backend/server.py
n0w0f: fixed in [0979a02](https://github.com/lamalab-org/corral/pull/206/commits/0979a02d8a9a2511a018af5706dcd5f2b419603d)

src/corral/backend/server.py
n0w0f: fixed in [0979a02](https://github.com/lamalab-org/corral/pull/206/commits/0979a02d8a9a2511a018af5706dcd5f2b419603d)

src/corral/backend/env.py
n0w0f: this is now fixed

src/corral/agents/react.py
n0w0f: fixed

src/corral/agents/prompt_utils.py
n0w0f: fixed

src/corral/agents/base_agent.py
n0w0f: fixed

n0w0f:
@kjappelbaum i have now incorporated all the suggestions.
The [exception handling](https://github.com/lamalab-org/corral/pull/206#discussion_r2509726363) need to be addressed.

src/corral/run.py
kjappelbaum: Thanks for clarifying this design decision. I still do not like setting score 0. 

I prefer returning the current score (we can still have a field to track if it was success/surrender/error. This seems to be the best option. I also prefer returning an exception over setting the score to 0.

src/corral/agents/react.py
```diff
@@ -183,6 +194,19 @@ def run(
             # Parse response
             thoughts, actions = self.parse_llm_response(llm_response)
 
+            # Check for surrender (XML format) if enabled
+            if enable_surrender and (
+                surrender_match := re.search(
+                    r"<surrender>(.*?)</surrender>",
```
kjappelbaum, line 200: A minor reminder here that we should track: It seems that we now want to have a note in the docs that we use XML notation for instructions. That means that if someone wants to implement another agent, going with those XML tags would be nice for consistency

src/corral/backend/env.py
```diff
@@ -35,10 +35,11 @@ class TaskState:
     trial_id: str = "0"
     messages: list[LLMMessage] = field(default_factory=list)
     tool_calls: list[ToolCall] = field(default_factory=list)
-    is_completed: bool = False
+    is_attempted: bool = False
     score: float | None = None
     submitted_answer: str | None = None
     feedback: str | None = None
+    forfeited: bool = False
```
kjappelbaum: surrendered? otherwise we have too many different words :D

src/corral/backend/schema.py
```diff
@@ -41,3 +41,18 @@ class ToolCall:
 class ToolRequest(BaseModel):
     tool_name: str
     arguments: dict[str, Any]
+
+
+class ForfeitRequest(BaseModel):
```
kjappelbaum: same here, I'd also change the name to "surrender" (in the docstrng and comment, too).

src/corral/backend/schema.py
```diff
@@ -41,3 +41,18 @@ class ToolCall:
...
+    # No parameters needed for forfeit
+
+
+class TrialCompletionResponse(BaseModel):
+    """Response when a trial is completed (via submission or forfeiture)"""
+
+    score: float
+    state: dict[str, Any]
+    trial_id: str
+    forfeited: bool = False
```
kjappelbaum: same here, I'd also change the name to "surrender" (in the docstrng and comment, too).

src/corral/backend/server.py
```diff
@@ -5,14 +5,47 @@
...
+from corral.backend.schema import ToolRequest, TrialCompletionResponse
 from corral.router.verbosity import (
     ToolVerbosity,
     VerbosityConfig,
     get_tools_guide_with_verbosity,
 )
 
 
+def _finalize_trial(
+    env: Environment, score: float, forfeited: bool = False
```
kjappelbaum: perhaps I forgot to load the most recent commit - but here, too, I would change to "surrender"

kjappelbaum: perhaps I would do a "search all"

src/corral/backend/server.py
```diff
@@ -5,14 +5,47 @@
...
+    completed_trial = env.get_completed_trial_data()
+
+    # Reset state for next trial
+    finished_trial_id = env.reset_state()
+
+    return TrialCompletionResponse(
+        score=score,
+        state=completed_trial["state"],
+        trial_id=finished_trial_id,
+        forfeited=forfeited,
```
kjappelbaum: here, too

src/corral/backend/server.py
```diff
@@ -5,14 +5,47 @@
...
+
+    This centralizes the 3-step finalization process:
+    1. Capture completed trial data (before reset)
+    2. Reset environment state for next trial
+    3. Return structured response
+
+    Args:
+        env: The environment instance
+        score: Trial score
+        forfeited: Whether trial was forfeited
```
kjappelbaum: here, too

src/corral/backend/server.py
```diff
@@ -162,26 +195,26 @@ def get_state(task_id: str):
...
-
-        # 1. Submit answer and score
         score = env.submit_answer(answer["answer"])
 
-        # 2. Get completed trial data (before any reset)
-        completed_trial = env.get_completed_trial_data()
+        return _finalize_trial(env, score, forfeited=False)
+
+    @app.post("/tasks/{task_id}/forfeit")
+    def forfeit_task(task_id: str) -> TrialCompletionResponse:
```
kjappelbaum: surrender?

src/corral/router/routes.py
```diff
@@ -3,11 +3,37 @@
...
+        TaskTrialResult with data from the completion response
+    """
+    completion = TrialCompletionResponse(**response_data)
+    return TaskTrialResult(
+        task_id=task_id,
+        trial_id=completion.trial_id,
+        score=completion.score,
+        state=completion.state,
+        tool_statistics=completion.state["tool_statistics"],
+        forfeited=completion.forfeited,
```
kjappelbaum: surrendered

src/corral/router/routes.py
```diff
@@ -99,14 +125,14 @@ def submit_answer(self, task_id: str, answer: str) -> TaskTrialResult:
...
-        return TaskTrialResult(
-            task_id=task_id,
-            trial_id=data["trial_id"],
-            score=data["score"],
-            state=data["state"],
-            tool_statistics=data["state"]["tool_statistics"],
-        )
+        return _parse_trial_completion(task_id, response.json())
+
+    def forfeit_task(self, task_id: str) -> TaskTrialResult:
```
kjappelbaum: surrender

src/corral/backend/env.py
n0w0f: updated now, i left one module out

src/corral/run.py
n0w0f: Thanks, I now when exceptions occur retrieve the current score.

n0w0f:
> New security issues found

[sourcery-ai](https://github.com/apps/sourcery-ai) what is the secutiry issue ?

kjappelbaum (approved):
looks good now to me!

PR #208, opened 2025-11-04
feat: enhance the visualization tool to allow annotations

MrtinoRG:
@kjappelbaum please let me know if something else is needed. Thanks!

## Summary by Sourcery

Enhance the agent trace visualizer by adding full annotation support with metadata, markers, notes, and validation, introduce node navigation and a no-tools filter, and refactor by externalizing styles and scripts into dedicated files

New Features:
- Add comprehensive annotation capabilities including annotator metadata inputs, behavioral marker selection, note-taking, persistence, and submission validation
- Introduce Previous/Next node navigation controls to streamline the annotation workflow
- Add a “No Tools” filter mode to exclude tool result nodes from the visualization

Enhancements:
- Extract inline CSS and JavaScript into external styles.css and app.js for improved maintainability
- Restructure the details panel into separate visualization and annotation sections


kjappelbaum:
at least when I look at the code, it does not seem that the mongodb is used? it right now just makes a console log instead of a submission. Is this expected?

kjappelbaum:
and for the markers: I am not sure, but do you believe that there can be a case where we would like to mark one node with two markers?

kjappelbaum:
Right now, it would also be limited in annotating traces from the tool-calling agent. If I'm looking at traces from `afm` in the `reports` directory, there is very little reasoning in the assistant messages and only tool-calling. As the annotation is right now only allowed on the assistant messages, we would get very little data there

kjappelbaum:
In addition, I might add a field for global notes on the trajectory

kjappelbaum:
but otherwise, it looks quite good - do you think that the markers are good ones? I did think a bit about them but I am sure you have way more elaborate thoughts

kjappelbaum:
and also here, it might make sense if we use the meeting tomorrow to label a few cases together to agree on a "style" and then we split up and I will also go over the PRs as I did for MaCBench

MrtinoRG:
> at least when I look at the code, it does not seem that the mongodb is used? it right now just makes a console log instead of a submission. Is this expected?

It is expected for now, because I wanted to have a first opinion such that now I can correct what you have pointed, and then add the MongoDB integration.

Do you think that asking for the MongoDB Key in a box is good? I mean it is not safe so perhaps we should use another method maybe?

MrtinoRG:
> and for the markers: I am not sure, but can there be a case where we would like to mark one node with two markers?

Yeah, it can happen, and we can work a bit on the markers as well, because right now, for example, the last two are mostly the same. But even with clear markers, several markers might align with one message.

MrtinoRG:
@kjappelbaum What is the best way to do the MongoDB integration? 

I was thinking in deploying a modal FastAPI endpoint that hosts the backend code for uploading each annotation when submitting, and then in the app ask for some identifier, that the annotator use, that will give access to the endpoint.

But what do you think?

MrtinoRG:
@kjappelbaum I now implemented something. 

But it has inherently some work for you (sorry). You have to either deploy the endpoint yourself, ask someone with the credentials to do it, or share the MongoDB credentials with me so I can deploy it.

PS: sorry for the spam of PRs

kjappelbaum:
> @kjappelbaum I now implemented something.
> 
> But it has inherently some work for you (sorry). You have to either deploy the endpoint yourself, ask someone with the credentials to do it, or share the MongoDB credentials with me so I can deploy it.
> 
> PS: sorry for the spam of PRs

Personally, I would have used MongoDB Atlas as this is free and easy to use. That is, you can still use modal for the backend but you could then create a database in MongoDB Atlas and directly set up the credentials for the modal app yourself.

kjappelbaum:
@cursor review

docs/trace-visualizer/index.html
```diff
@@ -386,712 +59,128 @@ <h1>🔬 Agent Trace Visualizer</h1>
...
+                                <label for="annotatorName">Annotator Name:</label>
+                                <input type="text" id="annotatorName" class="metadata-input" placeholder="Enter your name">
+                            </div>
+                            <div class="input-group">
+                                <label for="mongodbKey">Personal Identifier:</label>
+                                <input type="password" id="mongodbKey" class="metadata-input" placeholder="Enter your personal identifier">
+                            </div>
+                        </div>
+                    </div>
+                    <div class="behavioral-markers-section">
```
kjappelbaum, line 84: One question I had when trying this out: 
- Would it make sense to separate into two/three dropdowns to simplify use: Positive, negative, neutral behaviors?

docs/trace-visualizer/index.html
```diff
@@ -386,712 +59,128 @@ <h1>🔬 Agent Trace Visualizer</h1>
...
+                <div class="details-panel-right">
+                    <div class="metadata-section">
+                        <div class="metadata-inputs">
+                            <div class="input-group">
+                                <label for="annotatorName">Annotator Name:</label>
+                                <input type="text" id="annotatorName" class="metadata-input" placeholder="Enter your name">
+                            </div>
+                            <div class="input-group">
+                                <label for="mongodbKey">Personal Identifier:</label>
+                                <input type="password" id="mongodbKey" class="metadata-input" placeholder="Enter your personal identifier">
```
kjappelbaum, line 80: and without the `mongodbKey` I couldn't fully test it

kjappelbaum (approved):
code looks good! It is super useful! Thanks for taking initative!

PR #210, opened 2025-11-06
feat: add tests to the retro env + add database checks

MrtinoRG:
## Summary by Sourcery

Introduce data validation, database readiness checks, SMILES parsing safeguards, and extensive test coverage for retrosynthesis components

New Features:
- Add validate_molecule and validate_reaction functions for checking retrosynthesis tree nodes
- Implement check_database to verify connectivity, RDKit extension, and schema before environment startup
- Integrate pre-start database validation into create_rethrosynthesis_environments with user guidance on failures

Enhancements:
- Extend template search to validate SMILES input and raise clear errors

Tests:
- Add comprehensive unit and integration tests for scoring functions, SMILES utilities, price filtering, functional group detection, and retrosynthesis validation
- Introduce shared pytest fixtures in conftest.py for commonly used SMILES, template IDs, and retrosynthesis trees


tasks/retrosynthesis/retrosynthesis/checks.py
```diff
@@ -0,0 +1,185 @@
+import psycopg2
+from loguru import logger
+
+# Database configuration - should match the production database
+PRODUCTION_DB_CONFIG = {
```
kjappelbaum: perhaps it is just the naming - but certainly strange to have credentials for a "production" DB in plain text

tasks/retrosynthesis/retrosynthesis/checks.py
```diff
@@ -0,0 +1,185 @@
...
+    "host": "localhost",
+    "port": 5432,
+    "database": "reactions_production_db",
+    "user": "postgres",
+    "password": "postgres",
+}
+
+# Expected schema - tables and their required columns
+EXPECTED_SCHEMA = {
```
kjappelbaum, line 9: Something I am a bit worried about for this test here (and some of our Mocks) is that they create a lot of technical debt - if just one thing in the schema changes, we have to update two places. 

If the test is really supposed to test that the schema is constant, that makes sense. But if we only want to test that we can query the DB, that is overkill (and the code generation tools are quite prone to introduce this kind of technical debt)

kjappelbaum: Below the docstring speaks about "required columns" - here you seem to test the entire schema. 

This is not a big deal for this test itself — but those small things start adding up at some point

tasks/retrosynthesis/retrosynthesis/checks.py
```diff
@@ -0,0 +1,185 @@
...
+        logger.info("Checking required columns in each table...")
+        schema_errors = []
+
+        for table_name, expected_columns in EXPECTED_SCHEMA.items():
+            cursor.execute(
+                """
+                SELECT column_name
+                FROM information_schema.columns
+                WHERE table_schema = 'public'
+                AND table_name = %s
```
kjappelbaum, line 141: I'm wondering why you preferred this notation over f-strings. f-strings might be a bit easier to read

tasks/retrosynthesis/tests/test_retrosynthesis_utils.py
```diff
@@ -0,0 +1,631 @@
...
+
+    def test_tuple_of_tuples_no_match(self):
+        """Test with predicted as tuple of tuples with no match."""
+        ground_truth = ["CCO", "CC"]
+        predicted = (("CCC", "C"), ("CCCC", "C"))
+        assert species_match(ground_truth, predicted) is False
+
+    def test_invalid_smiles_in_ground_truth(self):
+        """Test with invalid SMILES in ground truth."""
+        ground_truth = ["CCO", "invalid"]
```
kjappelbaum, line 98: why would you have `invalid` in the ground truth?

tasks/retrosynthesis/tests/test_retrosynthesis_utils.py
```diff
@@ -0,0 +1,631 @@
...
+
+
+class TestFragmentMappedSmiles:
+    """Tests for _fragment_mapped_smiles function."""
+
+    def test_fragment_simple_molecule(self):
+        """Test fragmenting a simple molecule."""
+        mol = Chem.MolFromSmiles("CCO")
+        # Fragment containing first two carbons (indices 0, 1)
+        result = _fragment_mapped_smiles(mol, (0, 1))
```
kjappelbaum, line 373: in the code, you do some `+1` magic on the atom map indices. It seems "safe" because you clear previous annotations. But somehow it still feels scary - is this used mostly for readability or is it also used for functionality in some part?

tasks/retrosynthesis/tests/test_retrosynthesis_utils.py
```diff
@@ -0,0 +1,631 @@
...
+            ],
+        }
+        is_valid, error = validate_reaction(reaction)
+        assert is_valid is True
+        assert error == ""
+
+    def test_invalid_type(self):
+        """Test validation with wrong type."""
+        reaction = {"type": "mol", "template_id": 1, "children": []}
+        is_valid, error = validate_reaction(reaction)
```
kjappelbaum, line 556: why do you return the error as a string? This is not super pythonic and also not super consistent with the rest of the codebase

kjappelbaum: pythonic would be to just raise the error or to return some exception

kjappelbaum (commented):
It all looks plausible to me. I am more and more worried about the amount of code we pile up. There is nothing you can do about it - you do your best to implement stuff and write tests. But we collect so much code that it becomes harder and harder for one person to review a PR in a lot of detail (I also skimmed through some parts rather than line-by-line checks) or to fully have the full codebase in mind. 

Community adoption will be hard for a complex project - but without community, we will have a hard time maintaining it. This codebase is now big enough to have at least one full-time engineer for it. What I want to say with that: The more we can be concise with code and tests, the better. 
The more we can prepare the code for easy adoption (and perhaps fully migrating their environments to their own repos would help a lot to reduce the cognitive burden), the better.

tasks/retrosynthesis/tests/test_retrosynthesis_utils.py
MrtinoRG: Because as in the spectra env, the human that designed (me) is prone to errors

tasks/retrosynthesis/tests/test_retrosynthesis_utils.py
MrtinoRG: It is done because I think that normally the Mapped SMILES never contain an index 0 (zero). So it is done to avoid this index

tasks/retrosynthesis/tests/test_retrosynthesis_utils.py
MrtinoRG: What it is returned it is not an error as that, perhaps the naming is confusing here. What is being returned is the reason why the validation failed for the corresponding reaction. I believe that for this case, this is the correct way of returning this.

```
    if "template_id" not in reaction:
        return False, f"Reaction at {path} is missing required 'template_id' field"

    if not isinstance(reaction["template_id"], int):
        return False, f"Reaction at {path} has non-integer 'template_id' field"

    if "children" not in reaction:
        return False, f"Reaction at {path} is missing required 'children' field"

    if not isinstance(reaction["children"], list):
        return False, f"Reaction at {path} has non-list 'children' field"
```

Do you think that I should change the name of this variable?

tasks/retrosynthesis/retrosynthesis/checks.py
MrtinoRG: I have now changed this to only check for the columns that are required by the tools

tasks/retrosynthesis/retrosynthesis/checks.py
MrtinoRG: I have now moved this into being managed by a set config file that takes these variables from the .env file. As a results I have also added an example .env file :)

tasks/retrosynthesis/retrosynthesis/checks.py
MrtinoRG: I think that for writing SQL queries, normally this is the standard and recommended way

tasks/retrosynthesis/retrosynthesis/checks.py
kjappelbaum: 👍

tasks/retrosynthesis/retrosynthesis/checks.py
kjappelbaum: ok, I am not aware of a reason why for SQL query text one would do it a different way - what is the reason for that?

kjappelbaum (approved):
looks good!

tasks/retrosynthesis/retrosynthesis/checks.py
MrtinoRG: https://stackoverflow.com/questions/74257869/will-using-f-strings-to-request-data-cause-sql-injections

kjappelbaum: ok, it is not f-string vs. non-f-string but sanitization or not https://bobby-tables.com/python.html

PR #212, opened 2025-11-06
feat: add tests to the spectra elucidation environment

MrtinoRG:
## Summary by Sourcery

Add a new spectra_utils module by extracting reusable utilities from tools.py and refactor the tools to use it, and introduce extensive tests for both scoring logic and tool endpoints

Enhancements:
- Extract common spectra utility functions from tools.py into a dedicated spectra_utils module
- Refactor tools.py to import utility functions from spectra_utils instead of defining them inline

Tests:
- Add comprehensive unit tests for scoring functions in test_scoring.py
- Add end-to-end tests for spectra tools in test_tools.py

### DIFF
--- a/tasks/spectra_elucidation/spectra_elucidation/spectra_utils.py
+++ b/tasks/spectra_elucidation/spectra_elucidation/spectra_utils.py
@@ -1,0 +1,272 @@
+import re
+from collections import Counter
+from itertools import combinations
+
+import requests
+from rdkit import Chem
+
+_ELEMENT_PAT = re.compile(r"([A-Z][a-z]?)(\d*)")
+_PAREN_PAT = re.compile(r"\(([^()]*)\)(\d*)")
+_DOT_PAT = re.compile(r"·|\.")
+
+
+def differs_by_one_atom(parent_formula, fragment_formula):
+    """Check if fragment differs from parent by exactly one atom"""
+    total_diff = 0
+    all_elements = set(parent_formula.keys()) | set(fragment_formula.keys())
+
+    for element in all_elements:
+        parent_count = parent_formula.get(element, 0)
+        fragment_count = fragment_formula.get(element, 0)
+        diff = abs(parent_count - fragment_count)
+        total_diff += diff
+
+    # If total difference is exactly 1, it means one atom was added/removed
+    return total_diff == 1 or total_diff == 0
+
+
+def _parse_simple(formula: str) -> dict[str, int]:
+    """
+    Parse an already-expanded, dot-free formula into {element: count}.
+    """
+    counts = Counter()
+    for el, cnt in _ELEMENT_PAT.findall(formula):
+        counts[el] += int(cnt or 1)
+    return counts
+
+
+def _expand_parentheses(formula: str) -> str:
+    """
+    Recursively expand parentheses so that C6H5(CH3) becomes C6H5C1H3 etc.
+    """
+    while True:
+        m = _PAREN_PAT.search(formula)
+        if not m:
+            return formula
+        inner, mult = m.groups()
+        mult = int(mult or 1)
+        expanded = "".join(
+            f"{el}{int(cnt or 1)*mult}" for el, cnt in _ELEMENT_PAT.findall(inner)
+        )
+        formula = formula[: m.start()] + expanded + formula[m.end() :]
+
+
+def parse_molecular_formula(formula: str) -> dict[str, int]:
+    """
+    Parse molecular formula into an element-count mapping, handling
+    parentheses and dot adducts.
+    """
+    parts = _DOT_PAT.split(formula.replace(" ", ""))
+    total = Counter()
+    for part in parts:
+        expanded = _expand_parentheses(part)
+        total += _parse_simple(expanded)
+    return dict(total)
+
+
+def enumerate_fragments_from_smiles(
+    smi: str,
+    max_cuts: int = 2,
+    skip_ring_bonds: bool = True,
+    min_heavy_atoms: int = 2,
+    max_combos: int = 100000,
+):
+    """
+    Generate fragment SMILES by breaking up to `max_cuts` bonds.
+    Returns a sorted list of unique canonical SMILES of fragments.
+    Preserves ions by setting formal charges instead of adding hydrogens.
+    """
+    mol = Chem.MolFromSmiles(smi)
+    if mol is None:
+        raise ValueError("Invalid SMILES")
+
+    # Candidate bonds: heavy-atom bonds only; optionally skip ring bonds
+    cand_bond_idxs = []
+    for b in mol.GetBonds():
+        a1, a2 = b.GetBeginAtom(), b.GetEndAtom()
+        if a1.GetAtomicNum() == 1 or a2.GetAtomicNum() == 1:
+            continue
+        if skip_ring_bonds and b.IsInRing():
+            continue
+        cand_bond_idxs.append(b.GetIdx())
+
+    # Early exit: nothing to cut
+    if not cand_bond_idxs or max_cuts < 1:
+        return []
+
+    uniq = set()
+    tried = 0
+
+    for ncuts in range(1, min(max_cuts, len(cand_bond_idxs)) + 1):
+        for cutset in combinations(cand_bond_idxs, ncuts):
+            tried += 1
+            if tried > max_combos:
+                # Safety valve against combinatorial explosion
+                break
+
+            rw = Chem.RWMol(mol)
+            cut_atom_info = {}
+
+            # Count how many bonds each atom will lose
+            for bidx in cutset:
+                b = rw.GetBondWithIdx(bidx)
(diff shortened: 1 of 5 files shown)


tasks/spectra_elucidation/spectra_elucidation/spectra_utils.py
```diff
@@ -0,0 +1,272 @@
...
+
+import requests
+from rdkit import Chem
+
+_ELEMENT_PAT = re.compile(r"([A-Z][a-z]?)(\d*)")
+_PAREN_PAT = re.compile(r"\(([^()]*)\)(\d*)")
+_DOT_PAT = re.compile(r"·|\.")
+
+
+def differs_by_one_atom(parent_formula, fragment_formula):
```
kjappelbaum: do you still use this function? I think you mentioned that you do not use it anymore?

tasks/spectra_elucidation/spectra_elucidation/spectra_utils.py
```diff
@@ -0,0 +1,272 @@
...
+        )
+        formula = formula[: m.start()] + expanded + formula[m.end() :]
+
+
+def parse_molecular_formula(formula: str) -> dict[str, int]:
+    """
+    Parse molecular formula into an element-count mapping, handling
+    parentheses and dot adducts.
+    """
+    parts = _DOT_PAT.split(formula.replace(" ", ""))
```
kjappelbaum, line 44: I guess one thing that we have to keep in mind here is that this won't fail for stuff formulas with functional groups (`Me`, `Et`). That is, we might have silent errors.

tasks/spectra_elucidation/spectra_elucidation/spectra_utils.py
```diff
@@ -0,0 +1,272 @@
...
+    Args:
+        zones_dict (dict): Dictionary containing zones data with signals
+        frequency (str): NMR frequency (default: "600 MHz")
+        solvent (str): NMR solvent (default: "DMSO-d6")
+
+    Returns:
+        str: Formatted HSQC notation string
+    """
+    if "zones" not in zones_dict or "values" not in zones_dict["zones"]:
+        return "Invalid input format"
```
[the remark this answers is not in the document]
kjappelbaum: I guess this is a good suggestion

tasks/spectra_elucidation/spectra_elucidation/tools.py
kjappelbaum, line 1: thanks for deleting code ;)

tasks/spectra_elucidation/tests/test_scoring.py
```diff
@@ -0,0 +1,629 @@
...
+    def test_invalid_prediction(self):
+        """Test that invalid SMILES prediction returns score of 0.0."""
+        prediction = "INVALID_SMILES"
+        ground_truth = "CCO"
+        assert score_molecule(prediction, ground_truth) == 0.0
+
+    def test_invalid_ground_truth(self):
+        """Test that invalid ground truth SMILES raises ValueError."""
+        prediction = "CCO"
+        ground_truth = "INVALID_SMILES"
```
kjappelbaum, line 61: why would a ground truth be invalid?

tasks/spectra_elucidation/tests/test_scoring.py
```diff
@@ -0,0 +1,629 @@
...
+
+    def test_benzene(self):
+        """Test with aromatic molecule."""
+        prediction = "c1ccccc1"
+        ground_truth = "C1=CC=CC=C1"
+        assert score_molecule(prediction, ground_truth) == 1.0
+
+
+class TestNeutralizeCharges:
+    """Tests for the neutralize_charges function."""
```
kjappelbaum, line 73: where do you actually need this?  Looking at the source code, I am a bit surprised that rdkit does not complain more

tasks/spectra_elucidation/tests/test_scoring.py
```diff
@@ -0,0 +1,629 @@
...
+
+    def test_invalid_fragment(self):
+        """Test with one invalid fragment in the list."""
+        prediction = ["C", "CCCCCCCC"]  # Second fragment too large
+        ground_truth = "CCO"
+        score = score_molecule_fragments(prediction, ground_truth)
+        assert score == 0.0
+
+    def test_empty_list(self):
+        """Test with empty list returns 1.0."""
```
kjappelbaum, line 131: why is this the right behavior? you predict nothing and get full score?

tasks/spectra_elucidation/tests/test_scoring.py
```diff
@@ -0,0 +1,629 @@
...
+
+    def test_invalid_string_format(self):
+        """Test with invalid string format returns 0.0."""
+        prediction = "not a valid list"
+        ground_truth = "CCO"
+        score = score_molecule_fragments(prediction, ground_truth)
+        assert score == 0.0
+
+    def test_non_list_input(self):
+        """Test with non-list input returns 0.0."""
```
kjappelbaum, line 152: the question here is whether this should not raise an exception. 

This would also allow the agent to learn from this formatting error

tasks/spectra_elucidation/tests/test_scoring.py
```diff
@@ -0,0 +1,629 @@
...
+        """Test with extreme charge fragments (should be lenient)."""
+        prediction = ["[O-2]", "C"]
+        ground_truth = "CO"
+        score = score_molecule_fragments(prediction, ground_truth)
+        # Extreme charge fragments are treated leniently
+        assert score >= 0.0
+
+    def test_unparseable_fragment(self):
+        """Test that unparseable fragments are skipped."""
+        prediction = ["C", "INVALID_SMILES", "CC"]
```
kjappelbaum, line 183: here, too, the question if at least a warning should be propagated to the agent

tasks/spectra_elucidation/tests/test_scoring.py
```diff
@@ -0,0 +1,629 @@
...
+        prediction = "C2H6O"
+        ground_truth = "INVALID_SMILES"
+        with pytest.raises(ValueError, match="Invalid ground truth SMILES string"):
+            validate_molecular_formula(prediction, ground_truth)
+
+    def test_invalid_prediction_format(self):
+        """Test with invalid prediction format."""
+        prediction = "not_a_formula"
+        ground_truth = "CCO"
+        # Should return False or raise exception
```
kjappelbaum, line 222: I tend to favor the exception - but, in any case, it should do one thing and not "either or"

tasks/spectra_elucidation/tests/test_scoring.py
```diff
@@ -0,0 +1,629 @@
...
+        assert result == {"C": 2, "H": 6, "O": 1}
+
+    def test_formula_with_parentheses(self):
+        """Test parsing formula with parentheses."""
+        result = parse_formula("Ca(OH)2")
+        assert result == {"Ca": 1, "O": 2, "H": 2}
+
+    def test_formula_with_dot_adduct(self):
+        """Test parsing formula with dot notation (hydrates)."""
+        # Note: The parser doesn't handle multipliers outside parentheses like "5H2O"
```
kjappelbaum, line 257: This actually would be a "funny" ablation for the validation at some point. The tools might have bugs/limitations. 

For instance, they might only consume formulas formatted a certain way. If we then return what the tool parsed and the agents would validate the responses, they should spot it and do something about it 

but I guess, at this point, they would just ignore it

kjappelbaum (approved):
Looks good in general to me. 

My comments have not been about function but mostly enhnancement

tasks/spectra_elucidation/tests/test_scoring.py
MrtinoRG: Because the human (me) can potentially commit errors 🙃

tasks/spectra_elucidation/tests/test_scoring.py
MrtinoRG: I need it for the fragments. What I do there is to remove charges and other stuff and then compare the neutral fragments

tasks/spectra_elucidation/tests/test_scoring.py
MrtinoRG: An error

tasks/spectra_elucidation/tests/test_scoring.py
MrtinoRG: This is when the agent does not return a list, so I would say that it is correct to simply return 0

tasks/spectra_elucidation/tests/test_scoring.py
MrtinoRG: This is when scoring, so after this the agent receives a new trial or task...

tasks/spectra_elucidation/tests/test_scoring.py
MrtinoRG: I think when it comes from the agent, it should simply assign 0

PR #224, opened 2025-11-08
feat: add MCP integration

### LINKED ISSUE #46: Do we want to support Anthropic's MCP?
https://www.claudemcp.com/

it seems that this is now the standard people converge to

MrtinoRG:
## Summary by Sourcery

Enable Model Context Protocol integration by adding methods to convert Corral tools to MCP-compatible JSON Schema, implementing a schema converter module, updating documentation, and ensuring functionality through extensive tests.

New Features:
- Add for_mcp method to Tool class for converting tools to MCP JSON Schema format
- Introduce corral.mcp.schema_converter module for converting ToolArgument specifications to JSON Schema
- Expand Tool.from_mcp stub to raise NotImplementedError for unsupported conversion

Enhancements:
- Update documentation (TOOLS_README.md and README.md) with MCP integration guide and examples

Documentation:
- Add MCP Integration section in TOOLS_README.md
- Add MCP instructions in README.md

Tests:
- Add comprehensive tests for schema_converter functions (_extract_inner_type, _python_type_to_json_schema, etc.)
- Extend tool tests to cover for_mcp behavior, verbosity levels, choices, hidden args, and from_mcp error


src/corral/mcp/mcp_server.py
```diff
@@ -0,0 +1,181 @@
...
+                if tool.hidden_args:
+                    # If work_dir is in hidden_args and we have one, use it
+                    if "work_dir" in tool.hidden_args and self.work_dir:
+                        call_args["work_dir"] = self.work_dir
+                    else:
+                        # Use all hidden args from tool definition
+                        call_args.update(tool.hidden_args)
+
+                # Execute the tool in a thread pool to avoid blocking
+                result = await asyncio.to_thread(tool.execute, **call_args)
```
MrtinoRG: I do not have a good solution for this. I mean right now these arguments have no description because right now they are passed from the env, and not from the env. 

I think only the spectra env has these kind of args

src/corral/mcp/mcp_server.py
MrtinoRG, line 1: I understand that deploying this public (not locally) is relatively easy, and the code to do it has to come from the user, right?

kjappelbaum:
Before doing a detailed review: What is the design you have in mind here? Do you want to have bidirectional support at some point (corral can use MCP tools and expose them) or only unidirectional support?

MrtinoRG:
> Before doing a detailed review: What is the design you have in mind here? Do you want to have bidirectional support at some point (corral can use MCP tools and expose them) or only unidirectional support?

So, right now, in this PR what I aimed is that the tools in the different environments integrated in Corral can be deployed as an MCP server (different envs different servers). 

What I have built here, is just the conversion into the protocol of MCP, that allows to deploy a local server, or integrate the tools for local models such as copilot or Claude Code.

I believe, and that's the reason why I did not build beyond, that if someone (either us or any user) wants to deploy such tools in a public server; with the current code it would be almost trivial. But maybe this is my perception because of the lack of expertise. 

I would also love to include some integration such that MCP tools can be easily integrated in Corral. This is that if someone builds an environment and thinks that such tools are needed, can use this integration to easily deploy such tools. **But I think it is better if this goes in a different PR**

kjappelbaum:
The reason I asked this is because if you, too, would like to enable the support in both direction I'd like us to consider how a good API would look like. Ideally, we should not add additional modules but integrate it in a somewhat more elegant way. 

For instance, 

```
class Tool 

@classmethod 
def from_mcp(self, mcp_endpoint): 
	... 

def to_mcp(self): 
	# exposes this tool as MCP server 
```

I am not saying that the API here is the best one - but that I'd aim for an integration that is tighter. It feels natural to directly do it on the tools or the server. I tend to avoid adding new modules.

MrtinoRG:
> The reason I asked this is because if you, too, would like to enable the support in both direction I'd like us to consider how a good API would look like. Ideally, we should not add additional modules but integrate it in a somewhat more elegant way.

Ok, I agree with what you propose, but then one still needs to have the server module that manages the verbosity, or by default you give the entire description (perhaps this verbosity stuff should be at the tool docstring...)? Giving the entire description can be ok, but in the original issue you mentioned something about the docstrings so I am not super sure what to do.

At this point I would almost tend to drop the different levels of description and pass the entire docstring. If one wants to build the server, it can be easily done by taking the tools that are required with the classmethod `to_mcp`. Similarly, the tool decorator will have the from_docstring method that will take a tool from an MCP server and convert it into Corral server.

src/corral/backend/tool.py
```diff
@@ -64,6 +67,66 @@ def execute(self, **kwargs) -> str:
...
+
+        Raises:
+            NotImplementedError: This conversion is not currently supported
+        """
+        raise NotImplementedError(
+            "Converting from MCP to Corral Tool is not currently supported. "
+            "Corral tools are designed to be converted TO MCP format, not FROM it."
+        )
+
+    def for_mcp(self, verbosity: "ToolVerbosity | None" = None) -> dict[str, Any]:
```
kjappelbaum:
```suggestion
    def to_mcp(self, verbosity: "ToolVerbosity | None" = None) -> dict[str, Any]:
```

src/corral/backend/tool.py
```diff
@@ -64,6 +67,66 @@ def execute(self, **kwargs) -> str:
         """Execute the tool functionality"""
         raise NotImplementedError
 
+    @classmethod
+    def from_mcp(cls, mcp_tool_definition: dict[str, Any]) -> "Tool":
+        """
+        Create a Tool instance from an MCP tool definition.
+
+        This method is currently not implemented as Corral tools are the source,
```
kjappelbaum:
```suggestion
        This method is currently not implemented.
```

src/corral/backend/tool.py
```diff
@@ -64,6 +67,66 @@ def execute(self, **kwargs) -> str:
...
+        not the target of MCP conversions.
```
kjappelbaum:
```suggestion
```

src/corral/mcp/schema_converter.py
```diff
@@ -0,0 +1,221 @@
...
+
+    Args:
+        python_type: String representation of the Python type (e.g., "str", "int", "list[str]", "list[list[int]]")
+        required: Whether the field is required
+
+    Returns:
+        JSON Schema type specification
+    """
+    # Handle optional types (e.g., "str | None")
+    is_optional = "| None" in python_type or "|None" in python_type
```
kjappelbaum: do we have to code all of this ourselves again? Isn't there some parsing code we can reuse? There are tons of edge cases ...

README.md
```diff
@@ -390,6 +390,66 @@ tool_instance = MODAL_TOOL_REGISTRY["complex_calculation"]
...
+        param: Parameter description
+
+    Returns:
+        Result description
+    """
+    return f"Result: {param}"
+
+
+# Convert to MCP format
+mcp_definition = my_scientific_tool.for_mcp()
```
kjappelbaum:
```suggestion
mcp_definition = my_scientific_tool.to_mcp()
```

README.md
```diff
@@ -390,6 +390,66 @@ tool_instance = MODAL_TOOL_REGISTRY["complex_calculation"]
...
+
+# With specific verbosity level
+mcp_brief = my_scientific_tool.for_mcp(verbosity=ToolVerbosity.BRIEF)
```
kjappelbaum:
```suggestion
mcp_brief = my_scientific_tool.to_mcp(verbosity=ToolVerbosity.BRIEF)
```

README.md
```diff
@@ -390,6 +390,66 @@ tool_instance = MODAL_TOOL_REGISTRY["complex_calculation"]
...
+module = importlib.import_module("my_domain.tools")
+tools = {name: obj for name, obj in inspect.getmembers(module) if isinstance(obj, Tool)}
+
+# Create MCP server
+server = Server("my-corral-tools")
+
+
+@server.list_tools()
+async def list_tools():
+    return [MCPTool(**tool.for_mcp()) for tool in tools.values()]
```
kjappelbaum:
```suggestion
    return [MCPTool(**tool.to_mcp()) for tool in tools.values()]
```

README.md
kjappelbaum: if you do this change, you should - i guess - just do a find/replace over the entire codebase

docs/TOOLS_README.md
```diff
@@ -0,0 +1,550 @@
...
+    """
+    return str(a + b)
+```
+
+### Requirements for Tool Functions
+
+Every tool function must have:
+
+1. **Type hints** for all parameters and return value
+2. **Complete docstring** with:
```
kjappelbaum: should we name the docstring format (google/numpy/...) we expect

MrtinoRG: Yes!

docs/TOOLS_README.md
```diff
@@ -0,0 +1,550 @@
...
+        property_type: Type of property to analyze (choices: ["mass", "polarity", "reactivity"])
+
+    Returns:
+        Analysis results
+    """
+    # Implementation
+    return f"Analyzing {property_type} of {smiles}"
+```
+
+The docstring format `(choices: ["option1", "option2"])` automatically creates validation.
```
kjappelbaum: what kind of validation do you mean here? I feel it would help to be a bit more specific

docs/TOOLS_README.md
```diff
@@ -0,0 +1,550 @@
...
+    Returns:
+        API response
+    """
+    # The agent only sees 'endpoint', but both parameters are used
+    return f"Querying {endpoint} with key {api_key}"
+```
+
+## Tool Class Methods
+
+### `for_mcp(verbosity=None)`
```
kjappelbaum:
```suggestion
### `to_mcp(verbosity=None)`
```

docs/TOOLS_README.md
```diff
@@ -0,0 +1,550 @@
...
+
+Convert a Corral tool to MCP (Model Context Protocol) format:
+
```
kjappelbaum, line 117: I guess I would add a note that this requires that the verbosity be encoded in a certain way. And you should also say what happens if this is not encoded

docs/TOOLS_README.md
```diff
@@ -0,0 +1,550 @@
...
+```
+
+**Verbosity Levels:**
+- `BRIEF`: Minimal descriptions, just the essentials
+- `WORKFLOW`: Moderate detail, suitable for workflow documentation
+- `COMPREHENSIVE`: Full detail (default), all information included
+
+### `from_mcp(mcp_tool_definition)`
+
+This method is intentionally not implemented. Corral tools are the source of truth for tool definitions:
```
kjappelbaum: This reads strange :) Why is then there even the method if it is intentionally not implemented

docs/TOOLS_README.md
```diff
@@ -0,0 +1,550 @@
...
+# Create instance
+db_tool = DatabaseTool("postgresql://localhost/mydb")
+```
+
+### Modal Tools (Cloud Execution)
+
+Run computationally expensive operations in the cloud using Modal:
+
+```python
+from corral.utils.modal import modal_tool, MODAL_TOOL_REGISTRY
```
kjappelbaum, line 333: we should clarify when we need/should use the `MODAL_TOOL_REGISTRY`

docs/TOOLS_README.md
```diff
@@ -0,0 +1,550 @@
...
+        await server.run(streams[0], streams[1], server.create_initialization_options())
+
+
+if __name__ == "__main__":
+    import asyncio
+
+    asyncio.run(main())
+```
+
+For complete MCP integration examples and patterns, see [MCP Migration Guide](MCP_MIGRATION.md).
```
kjappelbaum: does this file exist?

docs/TOOLS_README.md
```diff
@@ -0,0 +1,550 @@
...
+tools = discover_tools("retrosynthesis.tools")
+for name, tool in tools.items():
+    print(f"{name}: {tool.description}")
+```
+
+## Best Practices
+
+### 1. Clear Documentation
+
+Always provide clear, comprehensive docstrings:
```
kjappelbaum: should we then also ask for the special syntax with `Workflow`, `Brief`, ...

docs/TOOLS_README.md
```diff
@@ -0,0 +1,550 @@
...
+    Returns:
+        JSON string containing computed properties
+    """
+    # Implementation
+    pass
+```
+
+### 2. Type Safety
+
+Always use proper type hints:
```
kjappelbaum, line 525: then we should really have `mypy` as a pre-commit hook ;)

docs/TOOLS_README.md
```diff
@@ -0,0 +1,675 @@
+# Corral Tools Documentation
+
+## Overview
+
+Corral provides a powerful tool system that allows you to create reusable, type-safe tools for AI agents. Tools are the building blocks that agents use to perform actions and gather information.
+
+**Key Features:**
+
+- Simple `@tool` decorator for creating tools from functions
+- Automatic MCP (Model Context Protocol) compatibility via FastMCP integration
```
kjappelbaum:
```suggestion
- Automatic [MCP (Model Context Protocol)](https://modelcontextprotocol.io/docs/getting-started/intro) compatibility via [FastMCP](https://github.com/jlowin/fastmcp) integration
```

docs/TOOLS_README.md
kjappelbaum: It seems it is not here yet. But we're using Google Docstrings (https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings)

docs/TOOLS_README.md
```diff
@@ -0,0 +1,675 @@
...
+- Type-safe parameter validation with comprehensive type hint support
+- Verbosity levels for controlling description detail
```
kjappelbaum:
```suggestion
- Verbosity levels for controlling description detail (if provided in the tool description) 
```

docs/TOOLS_README.md
kjappelbaum: Perhaps I did not see the update to the docs, but I guess we should be clearer on what kind of validation we provide.

kjappelbaum: perhaps even just a link to https://github.com/lamalab-org/corral/pull/224/files#diff-a6855df32ceaf8fa1e7d05c38fb9be1ffc7769fa78886104d0d9808e1c1700a5R235 would do it

docs/TOOLS_README.md
```diff
@@ -0,0 +1,675 @@
...
+
+# Or get specific tools by name
+simulation_tool = MODAL_TOOL_REGISTRY.get("complex_simulation")
+```
+
+**You don't need the registry if:**
+- You assign the decorated function to a variable and use it directly (though this is less common)
+- You're just defining the tool but not immediately using it
+
+The registry pattern ensures modal tools are easily discoverable and accessible throughout your application, especially when tools are defined across multiple modules.
```
kjappelbaum:
```suggestion

```

I think the text before is clear enough and this almost adds a bit of confusion

src/corral/backend/tool.py
```diff
@@ -64,6 +67,57 @@ def execute(self, **kwargs) -> str:
...
+
+        Returns:
+            Tool instance
+
+        Raises:
+            NotImplementedError: This conversion is not currently supported
+        """
+        raise NotImplementedError(
+            "Converting from MCP to Corral Tool is not currently supported. "
+            "Corral tools are designed to be converted TO MCP format, not FROM it."
```
kjappelbaum:
```suggestion
```

src/corral/backend/tool.py
```diff
@@ -64,6 +67,57 @@ def execute(self, **kwargs) -> str:
...
+            - description: str
+            - inputSchema: dict (JSON Schema)
+
+        Raises:
+            NotImplementedError: This base implementation is not supported.
+                                Use the @tool decorator for function-based tools.
+        """
+        raise NotImplementedError(
+            "to_mcp is only supported for function-based tools. "
+            "Use the @tool decorator or extend this method for custom tools."
```
kjappelbaum: I believe just saying "this is not implemented" might be clearer 

I feel it is strange to direct people to use the simpler tool decorator if they implement already the more complex class. If they go via the more complex task, they might have a more complex usecase they cannot solve with just the function

kjappelbaum (approved):
looks good! only have some comments on the docs

docs/TOOLS_README.md
MrtinoRG: Yes, I know which docstring style we are using, I forgot to add this, sorry.

docs/TOOLS_README.md
MrtinoRG: The tool decorator expects that the tools are typed. It is just because of that. I am not sure if Mypy will work for these cases, since the tools are not called in the code, but by the agent

docs/TOOLS_README.md
MrtinoRG: I describe below the different levels, so the use should be allowed to use any of them. But I do understand that comprehensive is confusing, so I will change this word.

docs/TOOLS_README.md
MrtinoRG: I added

docs/TOOLS_README.md
MrtinoRG: Added now

src/corral/agents/prompts/planner_prompt/user_prompt/prompt.md
MrtinoRG, line 1: @kjappelbaum now should be more clearly stated what it is intended from this agent

kjappelbaum:
I believe it looks good now - the failing tests are not caused by this PR, right?

PR #229, opened 2025-11-11
feat: add reports from resistor env chained

n0w0f:


n0w0f:
closing this as it is part of #242

PR #232, opened 2025-11-13
feat: allow intervention experiments

n0w0f:
## Summary by Sourcery

Implement a flexible hook system for Corral agents, enabling custom callbacks at key lifecycle points, integrate hook execution into agents and benchmarking workflows, and provide an intervention hook to inject predefined reasoning and optional tool calls.

New Features:
- Add core AgentHooks system with HookPoint enum and HookContext for lifecycle callbacks
- Integrate hook execution at BEFORE_TASK and BEFORE_ITERATION into ReActAgent and ToolCallingAgent run loops
- Introduce intervention hook factory with adapter registry to inject thoughts and actions into agent execution
- Allow passing a hooks object into BaseAgent initialization and the benchmark runner to enable hook processing

Enhancements:
- Extract Action and Thought dataclasses into a shared corral/agents/schema module


src/corral/agents/base_agent.py
```diff
@@ -242,6 +244,8 @@ def run(
             task_prompt (str, optional): The task prompt to use. Defaults to None.
             examples (list[str], optional): List with the few-shot examples to use. Defaults to None.
             enable_surrender (bool, optional): Whether to enable the surrender option, which allows the agent to give up solving a task. Defaults to False.
+            intervention_thought (str, optional): An intervention thought (It can have tool calls in there)to inject at the start of the task. Defaults to None.
```
kjappelbaum:
```suggestion
            intervention_thought (str, optional): An intervention thought (It can have tool calls in there) to inject at the start of the task. Defaults to None.
```

kjappelbaum: can it also have observations?

kjappelbaum: from a design point of view, I do not like adding one more `kwarg`.

I'd much more like a system where we could have hooks before/after sending a message or before/after starting the interaction. This would be more flexible and extensible, and avoid `kwarg`-explosion.

kjappelbaum: and it seems, that if this hook would be called on the env, it could maybe also ingest the tools

n0w0f: > can it also have observations?

Right now since this is the first message i did not think of having observation. But if we would want to intervene anywhere in trajectories. In that case a hook triggered by something  
would be a better design like you suggested. There would also be observation in those scenarios.

What are your thoughts ?  should we try intervention anywhere in trajectory and not only at the first step ?

kjappelbaum: At least from the software design I would go with a`hook`. The hook could then also figure out, based on the history if it is at the right place (first message or some other place in the conversation). This would allow us to do many interesting things in a very flexible way.

n0w0f:
@kjappelbaum what do you think about this design ?
If we abstract out agents a bit more to base agent, then we can cleanly add hooks at many places, including interaction or also maybe before and after tool calls

src/corral/agents/hooks/core.py
```diff
@@ -0,0 +1,224 @@
...
+    def __init__(self):
+        self._hooks: dict[str, list[tuple[HookCallback, int]]] = {}
+
+    def register(
+        self, hook_point: HookPoint | str, callback: HookCallback, priority: int = 0
+    ) -> None:
+        """Register a hook callback for a specific hook point.
+
+        Multiple hooks can be registered for the same hook point.
+        They will be executed in priority order (higher priority first).
```
kjappelbaum, line 148: This priority surprised me first. In this way of using it, it makes sense. What is the advantage of priority over some iterable?

src/corral/agents/hooks/core.py
```diff
@@ -0,0 +1,224 @@
...
+        # Sort by priority (descending)
+        self._hooks[point].sort(key=lambda x: x[1], reverse=True)
+
+    def execute(self, hook_point: HookPoint | str, context: HookContext) -> HookContext:
+        """Execute all hooks registered for a hook point.
+
+        Hooks are executed in priority order. If a hook sets
+        context.should_continue = False, execution stops early.
+
+        Errors in individual hooks are logged but don't stop execution
```
kjappelbaum: Do we really want to fail silently? What would happen if someone wants to use a hook to do early stopping on an agent trajectory. Perhaps they would like to throw an exception to enable this

src/corral/agents/hooks/intervention.py
```diff
@@ -0,0 +1,232 @@
...
+    """Inject intervention and execute any tools found in it.
+
+    Args:
+        context: The hook context
+        intervention_thought: The intervention text containing potential tool calls
+    """
+
+    agent_type = context.agent.__class__.__name__
+
+    if agent_type == "ToolCallingAgent":
```
kjappelbaum: This is a good point, I did not think about that much yet. Can we somehow avoid those checks for the classname? Some ideas (that are all not great): 

- the agent has some attribute about what things it supports
- we can use a parsing function on the agent and this returns an attribute about tool execution if the agent supports this

src/corral/agents/hooks/core.py
n0w0f: Main advantage would be that, user can still define hooks that would get priority over the existing hooks by core framework. With lists this manipulation would be difficult.

n0w0f: But i guess this would require documenting numbering scheme

src/corral/agents/hooks/core.py
n0w0f: For gracefully stopping the runs one can still use `context.should_continue=False`.

For exceptions, I am thinking if this should be based on a hook level policy,
then register hook signature would be verbose.
```
def register(
    self, 
    hook_point: HookPoint | str, 
    callback: HookCallback, 
    priority: int = 0,
    on_error: HookErrorPolicy = HookErrorPolicy.CONTINUE )
```

Other option is to define  a critical exception and use it in hooks.

What do you like and why ?

src/corral/agents/hooks/intervention.py
n0w0f: I think you were right about not including intervention as part of the agents, for the same reason, adding more variables to support this seems to be not the right thing.

After i looked at many design , having an adapterregistry for Hooks. Is what i liked the most.

This would also allow the user to write custom intervention logic  or agent specific logic which then can be registered and then the hook would use this registered logic.

Going forward i think there would be many agent specific logic to new hooks, so i guess this registry pattern is better than new attributes in the base class. 
What do you think about this ?

n0w0f: @kjappelbaum if you like this approach, then i can go ahead

kjappelbaum: how would an "adapterregistry for Hooks" look like? Do you have some pseudocode or system diagram?

n0w0f:
```
class HookAdapterRegistry:
    """Internal registry for agent-specific hook behavior and used by individual hooks.
    """
    
    def __init__(self, default_adapter: Callable):
        self._adapters: dict[str, Callable] = {}
        self._default = default_adapter
    
    def register(self, agent_class_name: str, adapter: Callable) -> Callable:
        """Register adapter for an agent type.
        """
        self._adapters[agent_class_name] = adapter #map based on name is fien ?
        return adapter
    
    def execute(self, context: HookContext, *args, **kwargs) -> Any:
        """Execute appropriate adapter based on agent type in context."""
        agent_type = context.agent.__class__.__name__
        adapter = self._adapters.get(agent_type, self._default)
        return adapter(context, *args, **kwargs)
        
```


Register agent specific logic for a hook
```
def _react_intervention(context: HookContext, intervention: str, execute_tools: bool):
    """ReAct-specific intervention logic."""


def _toolcalling_intervention(context: HookContext, intervention: str, execute_tools: bool):
    """ToolCalling-specific intervention logic."""


_intervention_registry = HookAdapterRegistry(default_adapter=_default_intervention)
_intervention_registry.register("ReActAgent", _react_intervention)
_intervention_registry.register("ToolCallingAgent", _toolcalling_intervention)



def create_intervention_hook(
    intervention_map: dict[str, str], 
    execute_tools: bool = False
) -> HookCallback:  # <-- Returns callback for AgentHooks.register()
    """Factory function to create an intervention hook.
    
    Returns:
        A HookCallback that can be registered with AgentHooks
    """
    
    def intervention_hook(context: HookContext) -> None:  # <-- This is the callback
        """The actual hook callback registered with AgentHooks."""
        if context.task_id not in intervention_map:
            return
        
        intervention_thought = intervention_map[context.task_id]
        logger.info(f"Injecting intervention for task {context.task_id}")
        
        context.metadata["intervention_applied"] = True
        
        # Use registry internally to delegate to right agent handler
        _intervention_registry.execute(context, intervention_thought, execute_tools)
    
    return intervention_hook  
```  
    
User API    

```    
from corral.agents.hooks import AgentHooks, HookPoint
from corral.agents.hooks.intervention import create_intervention_hook


hooks = AgentHooks()

# Create intervention callback
intervention = create_intervention_hook(
    intervention_map={"task1": "<thought>Consider X...</thought>"},
    execute_tools=False
)

hooks.register(HookPoint.BEFORE_TASK, intervention)
```

For new agents,
```
from corral.agents.hooks.intervention import intervention_registry

@intervention_registry.register("MyCustomAgent")  
def my_custom_intervention(context, intervention, execute_tools):
    # Custom handling
    ...
```

n0w0f: @kjappelbaum What do you think ? We still would have to register with agent names but i think it is cleaner and flexible

kjappelbaum: I think it is much cleaner. I still would love it if we could minimize coupling with the agents. But it might be that there really is no way to avoid this, as the prompts are too specific for different agents.

Unless we can think of a way to avoid this coupling, we can go for that. 

However, I would consider the case that there are some hooks that are really agnostic to the agent implementation. For instance, they might just compress something, or they might ingest information, or they might log something. 
And it should be easy to write something that does not depend on the specific agent implementation.

src/corral/agents/hooks/core.py
n0w0f: defined a critical exception here [042c366](https://github.com/lamalab-org/corral/pull/232/commits/042c3661b9c5242222298f6cb3d7f05bdea47778)
this would stop the execution

src/corral/agents/hooks/intervention.py
n0w0f: I have made the changes now.
With this implementation user can register an agent specific logic for their hooks, for general hooks, they need not register this adapter.
```
hooks = AgentHooks()
def my_hook(context: HookContext) -> None:
       print(f"Task: {context.task_id}")
hooks.register(HookPoint.BEFORE_TASK)
```

src/corral/agents/base_agent.py
```diff
@@ -78,7 +80,9 @@ def __init__(
         self.api_endpoint = api_endpoint
         self.temperature = temperature
         self.messages: list = []
-        self.token_usage: dict = {}  # Track token usage per LLM call
+        self.token_usage: list = {}  # Track token usage per LLM call
```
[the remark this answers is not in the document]
MrtinoRG: The sourcery is right @n0w0f, I do not know why you changed this but should be a dict

src/corral/agents/hooks/core.py
```diff
@@ -0,0 +1,245 @@
...
+        iteration: Current iteration number (0-indexed)
+        should_continue: Flag to control execution flow (set to False to stop)
+        skip_current_step: Flag to skip the current step without stopping
+        metadata: Dictionary for storing arbitrary hook-specific data
+        iteration_data: Rich data from current iteration (populated in AFTER_ITERATION)
+            Contains: llm_response, parsed_actions, tool_results, final_answer, etc.
+    """
+
+    task_id: str
+    agent: Any  # BaseAgent
```
MrtinoRG, line 89: I guess is Any to avoid circular imports (?)

MrtinoRG: And the same for the arguments below

src/corral/agents/react.py
```diff
@@ -176,7 +161,14 @@ def run(
...
         )
 
+        # Execute BEFORE_TASK hooks
+        self._execute_hooks(HookPoint.BEFORE_TASK, interface, task_id)
+
         for _iteration in range(self.max_iterations):
+            self._current_iteration = _iteration
+
+            # Execute BEFORE_ITERATION hooks
+            self._execute_hooks(HookPoint.BEFORE_ITERATION, interface, task_id)
```
MrtinoRG, line 171: This won't execute hooks every single time. What happens when we do not want to execute hooks? Maybe I am missing something

MrtinoRG: I understand that when we do not run hooks the context is returned as before. Still, I am not sure if it is optimal to have those modules active and initialized when we do not want to use them

src/corral/agents/hooks/intervention.py
```diff
@@ -0,0 +1,179 @@
...
+def _default_intervention(context: HookContext, intervention: str, execute_tools: bool):
+    """Generic intervention - just add as assistant message."""
+    context.messages.append(LiteLLMMessage(role="assistant", content=intervention))
+    if execute_tools:
+        logger.warning(
+            f"execute_tools not supported for {context.agent.__class__.__name__}"
+        )
+
+
+def _react_intervention(context: HookContext, intervention: str, execute_tools: bool):
```
MrtinoRG, line 31: Doesn't this need to return the context?

MrtinoRG: And the same for the one below

tests/agents/test_react.py
MrtinoRG, line 1: Don't you want to add tests to these hooks?

src/corral/agents/hooks/intervention.py
n0w0f: We don't need to return the context, since we are using callback system here python objects are passed by reference and this would directly modify the object.

src/corral/agents/hooks/core.py
n0w0f: Yea this essentially was to avoid that. there is possibility to use import guard but this any ways need to be done across the codebase. probably in a PR with mypy integration

src/corral/agents/react.py
n0w0f: The hooks are executed every single time these points are reached. But user control which hooks are registered when they initialize the agent, so hooks only run if explicitly added.

n0w0f: > What happens when we do not want to execute hooks? Maybe I am missing something

If we do not register any hooks it wont run. Also any mutation to the context happens only if there is is hook. There would be a dict lookup, by default the hook system is inactive

tests/agents/test_react.py
n0w0f: I have now added tests. If everything looks good to you, I can merge this

MrtinoRG (approved):
Yeah, it looks great!

PR #236, opened 2025-11-19
feat: refactor metrics and report

### LINKED ISSUE #52: in task definition, make it possible to define more than one scoring function
- in general, we would like to be able to measure things in more than one way 
- hence, we would like to be able to use multiple scoring functions

### LINKED ISSUE #237: Some report should contain the agent messages
Even if this implies that the messages have to be passed backwards to the runner, I think it has sense

MrtinoRG:
## Summary by Sourcery

Introduce a modular, registry-driven metrics system and refactor the reporting pipeline to use it

New Features:
- Add portable Metric and TaskMetric base classes with MetricMetadata and MetricContext protocol
- Implement a global MetricRegistry supporting dynamic registration, unregistration, plugin discovery, and parallel batch calculation
- Provide core metric implementations (average score, success rate, durations, token usage, tool calls, pass@k, etc.) as independent classes
- Extend BenchmarkResult and wandb_logger to delegate all metric calculations to the registry

Enhancements:
- Remove legacy metric calculation methods from BenchmarkResult and migrate report generation to consume registry outputs
- Add TaskTrialResult.tool_execution_duration property and improve error handling in context7_tools
- Refactor console and JSON report builders to format and fetch metrics dynamically from the registry

Documentation:
- Add a comprehensive Custom Metrics Guide and expose metrics API (get_registry, register_default_metrics, discover_plugin_metrics) in the public module

Tests:
- Introduce extensive unit tests for MetricRegistry, core metrics, base classes, protocol compatibility, and updated report generation behaviors


tests/test_metrics.py
```diff
@@ -1,12 +1,17 @@
...
 
-from corral.report.results import BenchmarkResult, TaskTrialResult, TaskTrialResults
-from corral.types import (
-    InsufficientTrialsError,
-    TaskNotFoundError,
+from corral.report.metrics import (
+    AverageScoreMetric,
+    Metric,
+    MetricMetadata,
+    get_registry,
```
kjappelbaum: i might name this registry with a somewhat more specific name

docs/custom_metrics.md
```diff
@@ -0,0 +1,433 @@
...
+
+The registry supports parallel calculation of metrics for improved performance:
+
+```python
+from corral.report.metrics import get_registry
+
+registry = get_registry()
+
+# Calculate all metrics in parallel
+results = registry.calculate_all(
```
kjappelbaum, line 180: this is cool :)

docs/custom_metrics.md
```diff
@@ -0,0 +1,433 @@
...
+You can register metrics before running a benchmark:
+
+```python
+from corral import CorralRunner
+from corral.report.metrics import get_registry
+from my_metrics import MyCustomMetric
+
+# Register custom metric
+registry = get_registry()
+registry.register(MyCustomMetric())
```
kjappelbaum: The thing I am hesitant with in this design pattern is that many things are hidden. You do not see the actual metrics unless you look into the code. 

It is also not obvious which runs it impacts (all in this script/process/thread, or only the subsequent ones, or is it persistent) 

Thus,  passing them in some way to the runner in an explicit way might be clearer

MrtinoRG: > The thing I am hesitant with in this design pattern is that many things are hidden. You do not see the actual metrics unless you look into the code.

There is one function that list all the metrics in the registry. Do you think that by exposing the registry in the runner is enough?

> It is also not obvious which runs it impacts (all in this script/process/thread, or only the subsequent ones, or is it persistent)

But this would be also related to improve the docs right?

MrtinoRG: Ok, now the metrics are exposed in the runner, such that they can be inspected before running the bench method

docs/custom_metrics.md
MrtinoRG, line 1: This is just that the person who writes the docs have it easier without having to navigate the codebase

src/corral/agents/base_agent.py
MrtinoRG, line 1: The changes here are such that the messages are also present in the main report. This simplifies mapping the scores with the run traces

src/corral/agents/utils.py
```diff
@@ -106,6 +106,11 @@ def llm_call(
 
         message = response.choices[0].message
 
+        # If message content is None, try to get reasoning_content
```
MrtinoRG, line 109: I observed that for GPT-oss, sometimes the model answer in content and others in reasoning content. So when the content is empty, this would incorporate the reasoning content there.

The tokens are already correctly considered since there is no reasoning tokens, they are included in completion tokens.

I am hesitant to add the reasoning content to the messages that are saved, since I think that generally the reasoning content has no much sense

src/corral/backend/env.py
MrtinoRG, line 1: The changes here are too solve the fact that the start time was set when initializing the environment, and not when initializing each trial

src/corral/report/results.py
```diff
@@ -658,65 +267,111 @@ def _prepare_report_data(
...
             report_data["task_results"][task_id] = task_result_data
 
         return report_data
 
-    def _display_console_report(
-        self, pass_at_k_results: dict, pass_hat_k_results: dict
-    ) -> None:
-        """
-        Display benchmark results to console using Rich tables.
+    def _build_summary_string(self, calculated_metrics: dict[str, Any]) -> str:
```
MrtinoRG, line 302: This is basically what now will be printed in the terminal.

I think that the results in the terminal were mostly not checked since most of the times the terminal window was not enough for all the results.

Also, with the docker implementation, to my understanding the terminal will be even more suboptimal

src/corral/run.py
```diff
@@ -24,17 +27,59 @@ def create_session_id() -> str:
     return f"session_{datetime.now(tz=timezone.utc).strftime('%Y%m%d_%H%M%S_%f')}"
 
 
+def sanitize_model_name(model_name: str, max_length: int = 50) -> str:
```
MrtinoRG, line 30: This is in order to generate the report name automatically.

MrtinoRG: Perhaps it would have sense to move it to the end of the file where the report is generated

src/corral/run.py
```diff
@@ -366,6 +580,9 @@ def trial_executor(task_id: str, trial_index: int) -> TaskTrialResult:
             self._save_finished_checkpoint(session_id, task_results)
             self._remove_original_checkpoint(session_id)
 
+            # Generate and save report
```
MrtinoRG, line 598: I incorporated this here, because otherwise the docker and cli implementations are not that clean

src/corral/agents/utils.py
kjappelbaum: but by not including them we would mostly save disk space and make reports easier to read, right?

MrtinoRG: Yes, I also think the reasoning is not needed, so it should be ok as it is right now

kjappelbaum:
It seems that there are still failing tests that should be fixed before merging

docs/custom_metrics.md
kjappelbaum: do you need this TOC? It will become updated soon and a documentation framework will handle that automatically

docs/custom_metrics.md
```diff
@@ -0,0 +1,707 @@
...
+# Get metrics as structured data
+metrics_info = result.list_metrics()
+
+for m in metrics_info:
+    print(f"{m['name']} ({m['type']}): {m['description']}")
+```
+
+### The MetricContext Protocol
+
+Metrics work with any object that satisfies the `MetricContext` protocol. This makes the metrics system portable and testable without tight coupling to `BenchmarkResult`.
```
kjappelbaum: here, an example of an object that satisfies this protocol would help

docs/custom_metrics.md
```diff
@@ -0,0 +1,707 @@
...
+### Parallel Metric Calculation
+
+The registry supports parallel calculation of metrics for improved performance:
+
+```python
+from corral.report.metrics import get_metrics_registry
+
+registry = get_metrics_registry()
+
+# Calculate all metrics in parallel
```
kjappelbaum: parallel = multiprocessing or threaded? (And does this type of parallelization make optimum sense?)

docs/custom_metrics.md
kjappelbaum: ` This makes the metrics system portable and testable without tight coupling to `BenchmarkResult`.` can perhaps even be dropped

docs/custom_metrics.md
```diff
@@ -0,0 +1,707 @@
...
+            task_id: ID of the task to calculate metric for
+
+        Returns:
+            The calculated metric value
+        """
+        task_trials = context.get_task_trials(task_id)
+        if not task_trials:
+            raise TaskNotFoundError(f"Task ID '{task_id}' not found.")
+
+        trials = task_trials.trials
```
kjappelbaum, line 266: This is perhaps more for the docs that @n0w0f is starting right now. But I believe we would benefit a lot from explaining some key terms like "task", "trial", "subtask" and where those are managed in the code

docs/custom_metrics.md
```diff
@@ -0,0 +1,707 @@
...
+# Uses all default metrics with k=[1, 3, 5]
+result = BenchmarkResult(task_results=my_results, k=[1, 3, 5])
+
+# See what metrics are active
+result.print_metrics()
+```
+
+### Configuring k Values for Pass@k and Pass^k Metrics
+
+The `k` parameter controls which pass@k and pass^k metrics are calculated. By default, `k=[5]` is used, which means pass@5 and pass^5 metrics will be computed.
```
kjappelbaum, line 300: I guess it is ok for now (and would not change in this PR). But the `k` argument is specific to one metric. But still, you expose it to the top-level `Result` class. Other metrics might have other arguments that you do not expose here. This could be seen as a bit inconsistent.

docs/custom_metrics.md
```diff
@@ -0,0 +1,707 @@
...
+**Task-Level Metrics:**
+
+- `task_success_rate`: Success rate for each individual task
+- `task_average_score`: Average score for each individual task
+- `task_average_duration`: Average trial duration for each task (seconds)
+- `task_total_token_usage`: Total token usage for each task (dict)
+- `task_pass_at_k`: Pass@k for each individual task (for each k value)
+- `task_pass_hat_k`: Pass^k for each individual task (for each k value)
+
+The `k` values for pass@k and pass^k metrics default to `[1]` but can be configured when creating a `BenchmarkResult`.
```
kjappelbaum, line 40: you start here talking about `BenchmarkResult` - but it is not really clear here what it is, how it is produced and why it is important

src/corral/agents/base_agent.py
```diff
@@ -225,7 +225,10 @@ def run_agent(
             enable_surrender (bool, optional): Whether to enable the surrender option, which allows the agent to give up solving a task. Defaults to False.
 
         Returns:
-            str: The final answer from the agent
+            tuple[str, list[dict[str, Any]], dict[str, int]]: A tuple containing:
```
kjappelbaum, line 228: This is a personal taste preference. But I started preferring datastructures with fields (named tuples, pydantic, ...)  over returning such complex things. This makes it easier to avoid misuse and also makes the documentation easier.

src/corral/backend/env.py
kjappelbaum: i.e. our original timings were not reliable, right?

kjappelbaum: but when I read the code I am not sure why the environment should handle timing of the tasks. It seems that the environment should mostly care about its own timing. The tasks can be timed in some other way.

kjappelbaum: my "problem" is that ` env.state.ensure_started()` is not elegant and easy to miss. It seems like something that should be automated/not there

kjappelbaum: it is just so easy to miss this call

src/corral/agents/base_agent.py
MrtinoRG: I would perhaps do that in a different PR

docs/custom_metrics.md
MrtinoRG: The docs that I included in this PR are just to help the person actually writing the docs, as I mentioned above

src/corral/backend/env.py
MrtinoRG: Ok, now I have added the time measuring code to the function `execute_single_trial` in the runner, and returned the env and server to the original version.

Then I have added a metric to measure the time of the literal trial time, and let the environment time as another one, as maybe can be useful in the future

src/corral/agents/utils.py
kjappelbaum: This probably interacts with the PR @n0w0f has opened to capture the reasoning

src/corral/report/metrics/base.py
kjappelbaum, line 11: good that you added `runtime_checkable`. Since we don't use type checking yet, and `ABC` would have been better without that. However, a `Protocol` seems philosophically more correct here. That is, I would not change anything but highlight that enabling static type checking is something we should do to really benefit from the `Protocol`

src/corral/report/metrics/core.py
kjappelbaum, line 31: an enhancement for another PR could be to allow the specification of the aggregation function (`mean`, `max`, `min`, `std`, `range`). In some other packages (e.g., `mofdscribe`) we allow passing a list of aggregation functions for such use cases.

kjappelbaum: then also the metrics like `AverageDuration` and `TotalDuration` would have less repeated code

kjappelbaum: it seems that there is quite some repeated code with the `Metric` implementation above. Do we need both? especially since the `_calculate_for_task` function exists

src/corral/report/metrics/registry.py
kjappelbaum: Is threading the best mode for parallelization here? I would expect that we are CPU bound and not I/O bound. Hence, multiprocessing seemed more intuitive to me.

src/corral/report/wandb_logger.py
kjappelbaum: shouldn't we be able to just check the type?

kjappelbaum (approved):
Thanks, @MrtinoRG, for this contribution, which will make it much cleaner. Perhaps you could review the comments, work on the ones you consider crucial, and then open issues for the ones you would like to delay.

Don't hesitate to repeatedly ping me if you want input or another review.

src/corral/report/metrics/registry.py
MrtinoRG: I chose threading because the metrics calculations are very lightweigth (at least right now).

If we think that some metrics could be heavy, CPU stuff has sense.  What was the reason behind your suggestion for multiprocessing?

MrtinoRG:
Merging this into `dev` as now I implemented the suggested changes

PR #238, opened 2025-11-27
WetLab + Tasks

aaaghajani:
The wetlab environment plus 30 tasks (3 levels, 10 each):

- Level1: aqueous solution of a single unknown salt, one anion + one cation, 20 mL of sample
- Level2: solution of 3 different cations, 20 mL of sample. The cations are chosen so that the detection is relatively straightforward; no two cations from the same analytical group
- Level3: solution of 3, 4, or 5 different cations, 15 mL of sample (or 20 mL in the case of 5 cations). The cations are chosen to make the task tricky; e.g. cations from the same analytical group.


aaaghajani:
Sorry, forgot to run pre-commit...
Please take a look at the code and let me know of any issues / suggestions / etc. I will then close this PR and make a new one with some cleanup, pre-commit, and changes after your reviews ;)

reports/wetlab-test/agent_logs-ToolCallingAgent-claude-sonnet-4-5-20250929-brief/qual_inorg_001_20251118_161736.json
MrtinoRG, line 1: I guess these files are going to go in the cleanup

tasks/wetlab/pyproject.toml
MrtinoRG, line 1: Do you perhaps want to add an author?

tasks/wetlab/wetlab/WetChem.yaml
```diff
@@ -0,0 +1,2887 @@
...
+
+- Name: Hg(+1)
+  Symbol: Hg(+1)
+  MolarMass: 0.20059
+
+- Name: Hg(0)
+  Symbol: Hg(0)
+  MolarMass: 0.20059
+
+# - Name: Oxalate
```
MrtinoRG: Why are some commented? Either remove them, or if they do not interfere/give errors just uncomment all

tasks/wetlab/wetlab/colors.py
```diff
@@ -0,0 +1,558 @@
...
+    'Co(NH3)6+2': [(450, 45, 6), (500, 60, 9), (650, 140, 3), (550, 1000, 1.5)], # The last "peak" is there just to uniformly reduce the absorbance over all wavelengths, resulting in a "brownish" color
+    'FeOH+2': (340, 70, 4000),
+    'Fe(OH)2+': (340, 70, 4000),
+    'Fe2OH2+4': (340, 70, 4000),
+    'FeCl+2': (335, 70, 1000), # DOI: 10.1016/j.chemgeo.2006.02.005
+    'FeCl2+': (345, 70, 3800), # DOI: 10.1016/j.chemgeo.2006.02.005
+    'FeCl3': (375, 80, 3000), # DOI: 10.1016/j.chemgeo.2006.02.005
+    'FeSCN+2': (460, 85 , 5000),
+    'Fe(SCN)2+': (485, 85, 9000),
+    #'FeOx+':     (300, 110, 1000),  # DOI: 10.1021/jp8040583
```
MrtinoRG: Same comment as previous file with the commented lines

tasks/wetlab/wetlab/colors.py
```diff
@@ -0,0 +1,558 @@
...
+    'black':            '#111111',
+}
+
+
+# Binning and mapping HSL lightness values
+L_BINS = np.array([0.20, 0.30, 0.45, 0.65, 0.75, 0.85, 0.91])
+L_PREFIX = {
+    1: 'dark',      # [0.20, 0.30)
+    2: 'deep',      # [0.30, 0.45)
+    3: '',          # [0.45, 0.65)
```
MrtinoRG, line 199: Is this supposed to be empty?

tasks/wetlab/wetlab/colors.py
```diff
@@ -0,0 +1,558 @@
...
+    18: 'pinkish red',          # [323, 340)
+    19: 'red'                   # [340, 360]
+}
+
+# creating the augmented pallet by adding slightly darker and lighter versions of the colors 
+AUG_PALLETT = {}
+for name, hex in PALLETT.items():
+    sRGB = sRGBColor.new_from_rgb_hex(hex)
+    h, s, l = convert_color(sRGB, HSLColor).get_value_tuple()
+    if 0.05 < l < 0.95:
```
MrtinoRG: Why do you need to check this?

tasks/wetlab/wetlab/env.py
```diff
@@ -0,0 +1,392 @@
...
+                initial_input=initial_input,
+                task_sys=data["input"]["sys"],
+                sample_list=data["input"]["samples"],
+                reagent_set=data["input"]["reagent_set"],
+                additional_reagents=data["input"].get("additional_reagents",[]),
+            )
+    return tasks
+
+
+class TaskEnvironment(Environment):
```
MrtinoRG: What do you to use a more descriptive name for the class? Like something related to the env

tasks/wetlab/wetlab/score.py
```diff
@@ -0,0 +1,91 @@
+from tools import CATIONS, ANIONS
+from typing import Dict
+import json
+
+ALL_IONS = CATIONS + ANIONS
```
MrtinoRG, line 5: Something we should improve in the chemistry envs, and that is already there in resistor is to avoid such hardcoded stuff

MrtinoRG: Like also to make easier creating new tasks

tasks/wetlab/wetlab/score.py
```diff
@@ -0,0 +1,91 @@
...
+
+CARBONATES = ["CO3-2", "HCO3-"]
+SULFATES = ["SO4-2", "HSO4-"]
+PHOSPHATES = ["PO4-3", "HPO4-2", "H2PO4-"]
+OXALATES = ["C2O4-2", "HC2O4-", "Ox-2", "HOx-"]
+CHROMATES = ["CrO4-2", "HCrO4-"]
+SULFIDES = ["S-2", "HS-"]
+
+
+def _normalize_ion(ion: str) -> str:
```
MrtinoRG, line 15: Like, would be possible to remove this, if one asks the agent to follow a convention when submitting?

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    "Rb+": "red-violet",
+    "Sr+2": "red",
+    #"Tl+": "green",
+}
+
+@tool
+def possible_cations() -> str:
+    """[BRIEF] Returns the list of possible cations. [/BRIEF]
+    
+    [DETAILED] This functions returns a space-separated string of all the possible cations that can be present in an unknown sample. One or more of these cations could be present in the unknown samples. [/DETAILED]
```
MrtinoRG: I would remove the last part, "could" can be confusing I think

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+
+    [PROCEDURAL] When to use this tool:
+    - Used to define the space of possible cations in unknown samples.
+    - Usually used early on in the analysis.
+    - Only suitable when the task involves identifying unknown cations in the sample(s). [/PROCEDURAL]
+
+    [WORKFLOW_INTEGRATION] Typical workflow integration:
```
MrtinoRG, line 77: In this one, we tried to provide the names of the specific tools in the before and after

MrtinoRG: And when there is a clear workflow of course

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    1. [PREREQUISITE] Make sure the task involves identifying one or more unknown cations in the samples. [/PREREQUISITE]
+    2. [CURRENT] Use this tool to know what possible cations can be present in the unknown sample(s). [/CURRENT]
+    3. [FOLLOW_UP] Can be followed up by tests to check for specific cations or groups of cations. [/FOLLOW_UP] [/WORKFLOW_INTEGRATION]
+
+    [CONTEXTUAL] How this tool works:
+    - It returns a pre-defined list of cations as a space-separated string.
+    - Each cation is formatted as 'X+n' where 'X' is the elemental symbol and '+n' is the charge.
+    - The ammonium ion is represented as 'NH4+' and mercury(I) dimer is represented as 'Hg2+2'.
```
MrtinoRG, line 85: Is this specific from the tool? I would perhaps remove this

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+        `possible_cations()`,
+    ]
+    [/SYNTACTICAL]
+
+    Args:
+        None
+
+    Returns:
+        str:
+            [BRIEF] a string containing all the possible cations [/BRIEF]
```
MrtinoRG: Could you add the "RETURNS" for each ney please?

MrtinoRG (commented):
You can also do not close this and simply run ruff format and ruff check that is what pre-commit does :)

src/corral/agents/react.py
kjappelbaum, line 1: to make the PR clean, I would stash all changes in `agents`. Your env should not require changes in this subpackage

tasks/wetlab/wetlab/WetChem.yaml
kjappelbaum, line 1: wow, that is a lot of stuff. I won't manage to check here every line

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+        - The list contains only the bare aqueous anions; anionic metal-complexes are not listed.
+        - This tool is only useful when the task involves identifying unknown anions. If the task is about identifying solutions from a possible list of solutions with known compositions, there is no need for this tool.
+    [/LIMITATIONS]
+    """
+    return ' '.join(ANIONS)
+
+    
+@tool(hidden_args=['compositions'])
+def measure_pH(compositions, label: str) -> str:
+    """[BRIEF] Measures the pH of the solution using a pH paper.[/BRIEF]
```
kjappelbaum, line 167: pH paper or indicator or probe?

kjappelbaum: it is just reading a bit inconsistent throughout the docstring

tasks/wetlab/wetlab/WetChem.yaml
aaaghajani: It'd be great if you can just take a look at the tools

src/corral/agents/react.py
aaaghajani: Yeah, I had made that change before merging `main` into this branch

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+                    [ERROR_DETAILS] The given `label` exists in the Inventory but the object it points to is not a solution (for example it could be a precipitate) [/ERROR_DETAILS]
+                    [ERROR_RECOVERY] Check the Inventory and make sure the `label` you use points to a solution [/ERROR_RECOVERY]
+
+        VolumeError: [ERROR_WHEN] When the solution contains a precipitate or less than 1 mL of it is remaining [/ERROR_WHEN]
+                     [ERROR_DETAILS] The given `label` exists in the Inventory as a solution, but it either contains a precipitate or the remaining volume is less than 1 mL [/ERROR_DETAILS]
+                     [ERROR_RECOVERY] If the solution contains a precipitate, filter it first and then perform the flame test again. If there is not enough solution to perform the test, you may be able to make more of it by repeating the steps that lead to it [/ERROR_RECOVERY]   
+    [/RAISES]
+
+    [LIMITATIONS] Known Limitations:
+        - The flame test's result depend on the concentration of cations; only species with a concentraion higher than 5e-4 molar will give a positive result.
```
kjappelbaum:
```suggestion
        - The flame test's result depends on the concentration of cations; only species with a concentration higher than 5e-4 molar will give a positive result.
```

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+        - If you have inroduced known interfering cations to the solution in previous steps, they will affect the results.
```
kjappelbaum:
```suggestion
        - If you have introduced known interfering cations to the solution in previous steps, they will affect the results.
```

perhaps you can install a spell-check extension in VSCode

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    """
+    try:
+        target = compositions[label]
+    except KeyError as e:
+        raise KeyError(f"Invalid label: {e}")
+    
+    if not isinstance(target, StockSolution):
+        raise ValueError(f"{label} is not a solution")
+    
+    test = 1 * target
```
kjappelbaum: out of curiosity - why is the explicit multiplication with 1 needed?

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+
+    colors = []
+    total_copper = 0
+    for sp, conc in test.composition.items():
+        if 'Cu' in sp:
```
kjappelbaum: why the special copper treatment? Sodium also gives a super strong yellow in traces

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    [RAISES] Exceptions:
+        KeyError: [ERROR_WHEN] When the given `label` is invalid [/ERROR_WHEN]
+                  [ERROR_DETAILS] The given `label` is not found in the Inventory [/ERROR_DETAILS]
+                  [ERROR_RECOVERY] Check the Inventory and ensure you are using a correct `label` [/ERROR_RECOVERY]
+    [/RAISES]
+
+    [LIMITATIONS] Known Limitations:
+        - If the target object is the result of a previously performed test, all color observations were already reported as part of that test, so using this tool to check those colors again will be redundant.
+        - All reported colors are qualitative and approximate.
+        - The perceived color of solutions will depend on the concentraion of species in that solution. Both the hue and the lightness of the perceived color can change as the concentration of species in the solution change.
```
kjappelbaum:
```suggestion
        - The perceived color of solutions will depend on the concentration of species in that solution. Both the hue and the lightness of the perceived color can change as the concentration of species in the solution change.
```

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    [/LIMITATIONS]
+    """
+    try:
+        target = compositions[label]
+    except KeyError as e:
+        raise KeyError(f"Invalid label: {e}")
+    
+    if type(target) == StockSolution:
+        sample = 10 * target
```
kjappelbaum: why this multiplication here?  why 10?

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+            [BRIEF] the list of precipitate colors [/BRIEF]
+            [DETAILED] a string where each separate line contains a single pure precipitate followed by its perceived color [/DETAILED]
+            [EXAMPLES] "AgBr     pale yellow\nAg2CO3     pale yellow\nAg2CrO4     brick red\n..." [/EXAMPLES]
+
+    [RAISES] Exceptions:
+        None
+    [/RAISES]
+
+    [LIMITATIONS] Known Limitations:
+        - This tool just returns the pre-defined list of characteristic flame colors and does not perform any calculations or tests.
```
kjappelbaum: flame colors?

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    [/LIMITATIONS]
+    """
+    statement = "NOTE: This list only describes colored (non-white) precipitates. If the color of a precipitate is not listed below, it means that it's white.\n"
+    precipitate_colors = [f"{prec} :    {color}" for prec,color in PREC_COLOR_LOOKUP.items()]
+
+    return statement + '\n'.join(precipitate_colors)
+
+@tool
+def simulate_color_mixture(mixture: list[tuple[str, float]]) -> str:
+    """[BRIEF] Given a mixture of colors, it mixes them with the given fractions and returns the name of the resulting color. [/BRIEF]
```
kjappelbaum: I'd perhaps also here mention that it is for precipitates and not solutions

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+            [DETAILED] a string where each available reagent appears on a separate line and is represented by "reagent label" and "composition". If no additional reagents are available in the tasks, it will be mentioned by this tool. [/DETAILED]
+            [EXAMPLES] "reagent label: HCl(0.02M)     composition: HCl 0.02 M, in water\nreagent label: KOH(6M)     composition: KOH 6.0 M, in water\n..." [/EXAMPLES]
+
+    [RAISES] Exceptions:
+        None
+    [/RAISES]
+
+    [LIMITATIONS] Known Limitations:
+    - This tool does not perform any tests; it simply returns the full list of available reagents.
+    - Some tasks do not have any external known reagent solutions. In that case, the task can be solved by mixing the sample solutions among themselves.
```
kjappelbaum:
```suggestion
```

i would make the tool descriptions independent of the tasks

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    [/LIMITATIONS]
+    """
+    reagent_descriptions = [f"reagent label: {k}     composition: {v.description}" for k,v in compositions.items() if type(v)==StockSolution]
+    if len(reagent_descriptions)==0:
+        return "There are no external reagents available in this task. You can still perform tests by mixing the sample solutions among themselves and make observations."
```
kjappelbaum:
```suggestion
        return "There are no external reagents available."
```

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    except KeyError as e:
+        raise KeyError(f"Invalid sol2_label: {e}")
+
+    if not isinstance(sol1, StockSolution):
+        raise RuntimeError(f"{sol1_label} is not a solution!")
+    if not isinstance(sol2, StockSolution):
+        raise RuntimeError(f"{sol2_label} is not a solution!")
+    
+    if type(sol1_vol) != int or sol1_vol<1:
+        raise ValueError("sol1_vol must be a positive integer")
```
kjappelbaum:
```suggestion
        raise ValueError("sol1_vol must be a positive integer larger than 1")
```

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    if type(sol2_vol) != int or sol2_vol<1:
+        raise ValueError("sol2_vol must be a positive integer")
```
kjappelbaum:
```suggestion
        raise ValueError("sol2_vol must be a positive integer larger than 1")
```

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    if test.has_precipitate:
+        observations.append(f"The supernatant solution is {sol_color}.")
+    else:
+        observations.append(f"The resulting solution is {sol_color}.")
+
+    return '\n'.join(observations)
+
+
+@tool(hidden_args=['compositions'])
+def add_a_solution(compositions, *, test_label: str, sol1_label: str, sol2_label: str,  sol2_vol: int) -> str:
```
kjappelbaum: why do you need `*` here? While it can be convenient, it can also be intransparent.

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    [RAISES] Exceptions:
+        KeyError: [ERROR_WHEN] When `sol1_label` or `sol2_label` is invalid [/ERROR_WHEN]
+                  [ERROR_DETAILS] The given `sol1_label` or `sol2_label` was not found in the Inventory [/ERROR_DETAILS]
+                  [ERROR_RECOVERY] Make sure you are passing the corrent labels. You can use the `get_available_reagets` and `check_inventory` tools. [/ERROR_RECOVERY]
+        
+        RuntimeError: [ERROR_WHEN] When `sol1_label` or `sol2_label` is not a solution [/ERROR_WHEN]
+                      [ERROR_DETAILS] The given `sol1_label` or `sol2_label` was found in the Inventory but the object it points to is not a solution. [/ERROR_DETAILS]
+                      [ERROR_RECOVERY] Make sure you are passing the corrent labels. You can use the `get_available_reagets` and `check_inventory` tools. [/ERROR_RECOVERY]
+        
+        ValueError: [ERROR_WHEN] When `sol2_vol` is not a positive integer [/ERROR_WHEN]
```
kjappelbaum: greater 1 should perhaps be added here, too

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    [RAISES] Exceptions:
+        KeyError: [ERROR_WHEN] When `sol1_label` or `sol2_label` is invalid [/ERROR_WHEN]
+                  [ERROR_DETAILS] The given `sol1_label` or `sol2_label` was not found in the Inventory [/ERROR_DETAILS]
+                  [ERROR_RECOVERY] Make sure you are passing the corrent labels. You can use the `get_available_reagets` and `check_inventory` tools. [/ERROR_RECOVERY]
+        
+        RuntimeError: [ERROR_WHEN] When `sol1_label` or `sol2_label` is not a solution [/ERROR_WHEN]
+                      [ERROR_DETAILS] The given `sol1_label` or `sol2_label` was found in the Inventory but the object it points to is not a solution. [/ERROR_DETAILS]
+                      [ERROR_RECOVERY] Make sure you are passing the corrent labels. You can use the `get_available_reagets` and `check_inventory` tools. [/ERROR_RECOVERY]
+        
+        ValueError: [ERROR_WHEN] When `sol1_vol` or `sol2_vol` is not a positive integer [/ERROR_WHEN]
```
kjappelbaum: greater 1 should perhaps be added here, too

kjappelbaum: i know that "positive integer" might imply this, but I'd be a bit more explicit perhaps

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+            new_color = test.precipitate.color_name
+            # if there is at least one shared color name between the old and new color, we add a "slightly" modifier 
+            old_color_set = set(old_color.split(' / '))
+            new_color_set = set(new_color.split(' / '))
+            shared_colors = old_color_set.intersection(new_color_set)
+            slightly = "slightly " if len(shared_colors)>0 else ""
+            
+            precipitate_ratio = new_amount / old_amount 
+
+            if precipitate_ratio >= 1.1 :
```
kjappelbaum: why 1.1?

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+                # getting the newly formed precipitate
+                test_no_prec = old_supernatant + sol2_vol * sol2.clone() # .clone() is used to prevent the volume of sol2 from decreasing
+                test_no_prec.equilibrate()
+                additional_color = " (color: " + test_no_prec.precipitate.color_name + ")"
+                tiny = "tiny amount of "  if (1000 * test_no_prec.precipitate.total_amount / test.volume < 5e-4) else "" # precipitates with an amount lower than 0.5 mmol/L are described as 'tiny'.
+                if additional_color == old_color:
```
kjappelbaum, line 954: do you ever hit this condition? It seems that the strings might be formatted differently

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+                    observations.append(f"A {tiny}precipitate with the same color as the existing precipitate forms.")
+                elif new_color == old_color: 
+                    observations.append(f"A {tiny}new precipitate{additional_color} forms, but does not cause the color of the existing precipitate to noticeably change.")
+                else:
+                    observations.append(f"A {tiny}new precipitate{additional_color} forms, mixing with the existing precipitate causing it to {slightly}change color. New color: {new_color}.")
+            
+            elif 0.8 <= precipitate_ratio < 1.1 :
```
kjappelbaum: why this range?

kjappelbaum: (such decisions are good to document in comments)

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+                if new_color == old_color:
+                    observations.append("The amount and color of the existing precipitate does not noticeably change.")
+                else:
+                    observations.append(f"The amount of the existing precipitate does not noticeably change, but its color {slightly}changes. New color: {new_color}.")
+
+            elif 0.5 <= precipitate_ratio <= 0.8 :
```
kjappelbaum: you have `=0.8` also in the condition above. Which one do you want to have precedence?

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+
+
+@tool(hidden_args=['compositions'])
+def add_precipitate_to_solution(compositions, *, test_label: str, prec_label: str, sol_label: str,  sol_vol: int) -> str:
+    """[BRIEF] Adds a precipitate to a specific volume of a solution and returns observations about the changes in the amount/color of the added precipitate or the solution color. [/BRIEF]
+    
+    [DETAILED] Draws `sol_vol` mL of `sol_label` (which must not contain any precipitates), adds to it all of the precipitate `prec_label` and reports observations about any changes in the color or the amount of the added precipitate and any color changes in the solution. It also adds the resulting solution to the Inventory and labels it `test_label`. [/DETAILED]
+
+    [PROCEDURAL] When to use this tool:
+    - This is the main tool to perform chemical tests on the filtered precipitates.
```
kjappelbaum: "the main tool" -> I am always a bit afraid of such statements. If you add a new tool this might impact this docstring here and the docstring might quickly be outdated. I am afraid of such coupling for that reason

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+                     [ERROR_RECOVERY] If the solution contains a precipitate, filter it using the `filter_solution` tool. If the remaining amount of it is less that the requested amount, try doing the test will a smaller `sol_vol` if possible, or make more of that solution by repeating the steps that lead to it. [/ERROR_RECOVERY]
+    [/RAISES]
+
+    [LIMITATIONS] Known Limitations:
+    - The solution may not contain any precipitates.
+    - The minimum allowed volume for the solution is 4 mL.
+    - The reported colors are qualitative and approximate.
+    - The perceived color of precipitates will depend on the composition of the precipitated solids. If more than compound co-precipitate at the same time, the color may be different from the color of pure precipitates.
+    - The perceived color of solutions will depend on the concentraion of species in that solution. Both the hue and the lightness of the perceived color can change as the concentration of species in the solution change.
+    - Any observations mentioning that the added precipitate "partially" or "mostly" dissolves are qualitative and approxiamte statements.
```
kjappelbaum:
```suggestion
    - Any observations mentioning that the added precipitate "partially" or "mostly" dissolves are qualitative and approximate statements.
```

tasks/wetlab/wetlab/score.py
```diff
@@ -0,0 +1,91 @@
...
+    
+    if binarize:
+        return 1.0 if (iou == 1) else 0.0
+    else:
+        return iou
+
+
+def score_salt(prediction: str, ground_truth: Dict, binarize=True) -> float:
+    if prediction.startswith("`"):
+        prediction = prediction.strip("`json")
```
kjappelbaum, line 54: this should be a general util for sanitization and not be repeated in different envs or different functions

tasks/wetlab/wetlab/env.py
```diff
@@ -0,0 +1,392 @@
...
+    "H2S(acidic)": StockSolution(composition={"H2S": 0.1, "H+": 0.05, "Cl-": 0.05}, description="H2S 0.1 M, HCl 0.05 M, in water"),
+    "(NH4)2S": StockSolution(composition={"NH4+": 2, "S-2": 0.5, "NO3-": 1}, description="(NH4)2S 0.5 M + NH4NO3 1.0 M, in water"),
+    "(NH4)2CO3": StockSolution(composition={"NH4+": 0.4, "CO3-2": 0.2}, description="(NH4)2CO3 0.2 M, in water"),
+    "(NH4)2HPO4": StockSolution(composition={"NH4+": 0.4, "HPO4-2": 0.2}, description="(NH4)2HPO4 0.2 M, in water"),
+    "(NH4)2SO4": StockSolution(composition={"NH4+": 1, "SO4-2": 0.5}, description="(NH4)2SO4 0.5 M, in water"),
+    "K2CrO4": StockSolution(composition={"K+": 0.4, "CrO4-2": 0.2}, description="K2CrO4 0.2 M, in water"),
+    "NH4SCN": StockSolution(composition={"NH4+": 0.1, "SCN-": 0.1}, description="NH4SCN 0.1 M, in water"),
+    "DMG": StockSolution(composition={"K+": 0.01, "Hdmg-": 0.01}, description="dimethylglyoxime potassium salt 0.01 M, in water"),
+}
+SYS_1 = "Cl N S I C K P Cr dmg"
```
kjappelbaum, line 65: if one could automate the creation of this based on the compositions, it might be more robust and scalable (can be done in a separate PR)

tasks/wetlab/wetlab/env.py
```diff
@@ -0,0 +1,392 @@
...
+        logger.info(f"Initializing environment for task {self.task_id}")
+        logger.info(f"Task name: {self.current_task}")
+        self._add_task_tools()
+
+    def configure_additional_apps(self):
+        """Setting the Reaktoro chemical system"""
+        set_chemical_system(self.current_task.chemical_system)
+        logger.info(f"Reaktoro chemical system is set to '{self.current_task.sys}' for {self.task_id}.")
+
+    def _add_task_tools(self):
```
kjappelbaum, line 233: all this code here does not seem to be specific to the wet lab env and probably should be in a more reusable location

tasks/wetlab/wetlab/engine.py
```diff
@@ -0,0 +1,424 @@
...
+    real = list(elements - PSEUDO_ELEMENTS)
+    pseudo = list(elements.intersection(PSEUDO_ELEMENTS))
+
+    results = rk.speciate(real)
+    results.symbols += pseudo
+    return results
+
+
+def set_chemical_system(system: str | rk.ChemicalSystem) -> rk.ChemicalSystem:
+    global DEFAULT_SYS, SOLVER, SOLVER_OPTIONS
```
kjappelbaum, line 55: do you have to work with globals? I find them a bit scary because of side effects they can cause

tasks/wetlab/wetlab/engine.py
```diff
@@ -0,0 +1,424 @@
...
+    def sys_species(self) -> Generator:
+        sys = self.state.system()
+        for sp in sys.species():
+            yield sp.name()
+
+
+    @property
+    def color(self) -> str:
+        fracs = [mol/self.total_mol for mol in self.composition.values()]
+        color_names = [PRECIPITATE_COLORS.get(name, 'white') for name in self.components]
```
kjappelbaum, line 147: i.e. if someone forgets to put the color in the list here, we do silently fill in "white" even though the color might be very different?

tasks/wetlab/wetlab/WetChem.yaml
aaaghajani: The commented ones are not required for the current tasks, but I wanted to keep them for the future. But you're right, I'll keep them somewhere else and remove them here.

tasks/wetlab/wetlab/colors.py
aaaghajani: Yes!

tasks/wetlab/wetlab/colors.py
aaaghajani: There's no need to create darker and lighter versions of the color if it is already super dark or super light.

tasks/wetlab/wetlab/score.py
aaaghajani: It's not a matter of notation/convention. In level 1 tasks, the agent is required to identify exactly which anion is present and will only get partial score if the protonation state is not correct.

tasks/wetlab/wetlab/tools.py
aaaghajani: Those are the only multi-atomic cations in the list, that's why I singled them out

tasks/wetlab/wetlab/score.py
aaaghajani: Any suggestions how to change this?

tasks/wetlab/wetlab/tools.py
aaaghajani: In the cases where there are specific tools, I name them.

tasks/wetlab/wetlab/tools.py
aaaghajani: it's a pH-indicator paper, using universal indicator. Like these: https://en.wikipedia.org/wiki/Universal_indicator#/media/File:Universal_indicator_paper.jpg

tasks/wetlab/wetlab/tools.py
aaaghajani: This "draws" 1 mL from the target solution, causing the target volume to decrease by 1

tasks/wetlab/wetlab/tools.py
aaaghajani: Copper is the only one that can exist as different species (e.g. complexes). This takes all of the copper containing species into account instead of just "Cu+2"

tasks/wetlab/wetlab/tools.py
aaaghajani: multiplication draws a specific volume, thus converting a StockSolution object (without volume) to a Solution object. Equilibrium calculations are only available for the Solution objects.

aaaghajani: 10 is arbitrary

tasks/wetlab/wetlab/tools.py
aaaghajani: to force kw-only arguments, so that the agent has to be explicit with the labels and volumes

tasks/wetlab/wetlab/engine.py
aaaghajani: How do you suggest we change this? One alternative is to include ALL of the white precipitates, but there are too many of them. At least this way only non-white precipitates must be added to that list which are by far fewer than the white ones.

tasks/wetlab/wetlab/engine.py
aaaghajani: In the current implementation: yes I do.
But I can think about how to change this in the future

tasks/wetlab/wetlab/score.py
aaaghajani: where do I move this to?

tasks/wetlab/wetlab/env.py
aaaghajani: you mean setting the chemical system? or... ?

tasks/wetlab/wetlab/score.py
MrtinoRG: In here it would be to include all the ions in the world 🙃 

I will try to think on a better solution :)

MrtinoRG: Perhaps have a small set of rules that allow to sample some ion and normalize it into the shape that the environment expects

tasks/wetlab/wetlab/score.py
MrtinoRG: There is an open issue for this already, so I would not change it in this PR

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+
+    Args:
+        label (str):
+            [BRIEF] label of the target solution [/BRIEF]
+            [DETAILED] a string representing the label of the solution in the Inventory, for which the pH will be measured [/DETAILED]
+
+    Returns:
+        int:
+            [BRIEF] the closest integer value to the actual pH of the solution [/BRIEF]
+            [DETAILED] an integer value between 0-14, representin the closest integer to the actual pH of the solution [/DETAILED]
```
MrtinoRG: Representing, there is a typo

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+        `measure_pH("test_01")`,
+        `measure_pH("sample_1_test_2")`,
+    ]
+    [/SYNTACTICAL]
+
+    Args:
+        label (str):
+            [BRIEF] label of the target solution [/BRIEF]
+            [DETAILED] a string representing the label of the solution in the Inventory, for which the pH will be measured [/DETAILED]
+
```
MrtinoRG, line 203: In the args, I think you are missing the Syntactical and examples in all of them

tasks/wetlab/wetlab/tools.py
MrtinoRG: Ok, perfect then. I was not sure if you knew about it

tasks/wetlab/wetlab/tools.py
MrtinoRG: Also, for the args, you also need to add `args_` to the keys. Just to be sure that you are aware

tasks/wetlab/wetlab/tools.py
aaaghajani: This was before I added the "tiny" modifier. I will change the logic and remove this.

tasks/wetlab/wetlab/tools.py
aaaghajani: You're right!

tasks/wetlab/wetlab/tools.py
aaaghajani: the 1.1 will be removed and I changed 0.8 to 0.85
It is arbitrary, but I think we can safely assume that any change less than 15% is not "noticeable"

tasks/wetlab/wetlab/tools.py
aaaghajani: It also checks if there is a precipitate in the solution

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    - If a solution does not contain a precipitate, it will be called a "clear solution" and its color will be reported.
+    - Color names are subjective and qualitative, so treat the reported color names as only approximate.
+    - Since color names are subjective and approximate, sometimes it is possible that a color can be described with more than one color name. In those cases the possible color names are separated by a slash '/'. [/CONTEXTUAL]
+
+    [SYNTACTICAL] Usage examples:
+    [
+        `checkout_color("sample")`,
+        `checkout_color("test_1_filtrate")`,
+        `checkout_color("test_2_precipitate")`,
+    ]
```
MrtinoRG, line 415: Here I understand that there are no more possible examples, right? Otherwise we try to have 5

Like just checking that you are aware of the conventions that we followed since you joined later

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    [/SYNTACTICAL]
+
+    Args:
+        label (str):
+            [BRIEF] label of the target object [/BRIEF]
+            [DETAILED] a string representing the label of the object in the Inventory which can be a solution (with or without a precipitate), a reagent, or a filtered precipitate [/DETAILED]
+
+    Returns:
+        str:
+            [BRIEF] the type and color of the object [/BRIEF]
```
MrtinoRG: Perhaps indicate what "type" is. At least to me it is not super clear what do you mean with type just (based on the description here). Perhaps you can add within parentheses the options as you have in the detailed one

tasks/wetlab/wetlab/tools.py
MrtinoRG: But then why you do not use 1 as above? (Perhaps is a different situation the one above)

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    - It receives a mixture of up to 3 color names and their fractions as a list of tuples.
+    - Each tuple must be formated as (color_name, fraction).
+    - The fractions must be all positive and sum to 1.0. They will be rounded to one decimal place before performing the prediction, so there is no point in having more than one decimal points.
+    - The colors will be mixed and the name of the closest matching color will be reported. [/CONTEXTUAL]
+
+    [SYNTACTICAL] Usage examples:
+    [
+        `simulate_color_mixture([("black", 0.5), ("pale yellow", 0.5)])`,
+        `simulate_color_mixture([("cyan", 0.2), ("white", 0.2), ("pink", 0.6)])`,
+    ]
```
MrtinoRG, line 544: Here again the example count

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    return closest_color_names(result_hex, mode='precipitate', max_names=1)
+
+@tool(hidden_args=['compositions'])
+def get_available_reagents(compositions) -> str:
+    """[BRIEF] Returns the list of available reagent solutions. [/BRIEF]
+    
+    [DETAILED] Returns a string where each reagent appears on a separate line. Each line contains the reagent's label as well as its composition. [/DETAILED]
+
+    [PROCEDURAL] When to use this tool:
+    - Usually called in the begining to know all of the possible reagents available (if any) to solve the task.
```
MrtinoRG: At the begining of what? I would try to be more specific

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    - If you want to perform a certain test and need a specific known solution, you can use this tool to check if that is available as a reagent.
+    - You can use this tool to get the exact concentration of the components in the reagent solutions. [/PROCEDURAL]
+
+    [WORKFLOW_INTEGRATION] Typical workflow integration:
+    1. [PREREQUISITE] None. [/PREREQUISITE]
```
MrtinoRG, line 607: Perhaps add something regarding that interpreting the input

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    [/LIMITATIONS]
+    """
+    reagent_descriptions = [f"reagent label: {k}     composition: {v.description}" for k,v in compositions.items() if type(v)==StockSolution]
+    if len(reagent_descriptions)==0:
+        return "There are no external reagents available in this task. You can still perform tests by mixing the sample solutions among themselves and make observations."
+    else:
+        return '\n'.join(reagent_descriptions)
+
+@tool(hidden_args=['compositions'])
+def mix_two_solutions(compositions, *, test_label: str, sol1_label: str, sol1_vol: int, sol2_label: str,  sol2_vol: int) -> str:
```
MrtinoRG: Why you add the star? I mean based on how the framework works, it should not be needed. If you observed some failure cases related to this perhaps we can open a new issue because I believe this should be handled by the framework

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    1. [PREREQUISITE] Make sure that `sol1_label` and `sol2_label` both point to reagents or solutions in the Inventory and that neither of them contain a precipitate. If a solution has precipitation, filter it using the `filter_solution` tool.  [/PREREQUISITE]
+    2. [CURRENT] Use this tool to mix the solutions and add the new resulting solution to the Inventory. Notice the reported observations. [/CURRENT]
+    3. [FOLLOW_UP] You can use your chosen `test_label` to refer to this solution when performing further tests. Sometimes for a given color change or precipitation to happen, you need to add more of one of the initial solutions, in that case you can use the `add_a_solution`. If the test resulted in the formation of a precipitate, you can either add other solutions to the mixture using the `add_a_solution` tool or filter the precipitate using the `filter_solution` tool. You can also lookup the color of precipitates by calling the `lookup_precipitates_color` tool. If you have a hypothesis about the composition of the formed precipitate you can try to check your hypothesis using the `simulate_color_mixture` tool. [/FOLLOW_UP] [/WORKFLOW_INTEGRATION]
```
MrtinoRG: I kind of believe that this is too many hints. Like more than we have in the other environments.

What do you think?

MrtinoRG: Like I mean in the task workflow side

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    - Sometimes it is possible that the resulting precipitate's color can be described using more than one color name. In those cases the different given names will be separated by a slash '/'.
+    - Remember that the reported colors are qualitative and approximate. [/CONTEXTUAL]
+
+    [SYNTACTICAL] Usage examples:
+    [
+        `mix_two_solutions(test_label="test_1", sol1_label="HCl(1M)", sol1_vol=5, sol2_label="sample", sol2_vol=5)`, # mixing 5 mL of reagent 'HCl(1M)' and 5 mL of 'sample', labeling the result as 'test_1'
+        `mix_two_solutions(test_label="test_2_iodide", sol1_label="test_1", sol1_vol=3, sol2_label="NH4I", sol2_vol=2)`, # mixing 3 mL of 'test_1' and 2 mL of reagent 'NH4I', labeling the result as 'test_2_iodide'
+        `mix_two_solutions(test_label="A_B", sol1_label="sample_A", sol1_vol=10, sol2_label="sample_B", sol2_vol=10)`, # mixing 10 mL of 'sample_A' and 10 mL of 'sample_B', labeling the result as 'A_B'
+        `mix_two_solutions(test_label="A_B_HCl", sol1_label="A_B", sol1_vol=10, sol2_label="HCl(6M)", sol2_vol=2)`, # mixing 10 mL of 'A_B' and 2 mL of reagent 'HCl(6M)', labeling the result as 'A_B_HCl'
+    ]
```
MrtinoRG: I would remove the comments here since you have to add the "[EXAMPLES]" field in the arguments

tasks/wetlab/wetlab/tools.py
MrtinoRG: But I think the framework handles this. Like the agent will always answer with the arguments, otherwise an error is raised

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    1. [PREREQUISITE] If you used the `mix_two_solutions` tool and want to keep adding more of one of the solutions, you can use this tool. You can also use this tool if a previous test resulted in the formation of precipitate and you want to perform further tests on the resulting mixture.  [/PREREQUISITE]
+    2. [CURRENT] Use this tool to add the given volume of sol2 to the whole remaining amount sol1. Notice the reported observations. [/CURRENT]
+    3. [FOLLOW_UP] You can use your chosen `test_label` to refer to the resulting solution when performing further tests. If the test resulted in the formation of a precipitate, you can either add other solutions to the mixture by calling this tool again or filter the precipitate using the `filter_solution` tool. You can also lookup the color of precipitates by calling the `lookup_precipitates_color` tool. If you have a hypothesis about the color change of a pre-existing precipitate you can try to check your hypothesis using the `simulate_color_mixture` tool. [/FOLLOW_UP] [/WORKFLOW_INTEGRATION]
+
+    [CONTEXTUAL] How this tool works:
+    - It draws `sol2_vol` mL from `sol2_label`.
```
MrtinoRG, line 812: I am wondering if these sol1 and sol2 labels are not confusing?

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    - When an observation mentions that a precipitate has mostly dissolved, it means that more than 50% of it has dissolved but there is still some undissolved precipitate remaining.
+    - Sometimes it is possible that the precipitate's color can be described using more than one color name. In those cases the different given names will be separated by a slash '/'.
+    - Remember that the reported colors are qualitative and approximate. [/CONTEXTUAL]
+
+    [SYNTACTICAL] Usage examples:
+    [
+        `add_a_solution(test_label="test_2B", sol1_label="test_2A", sol2_label="HCl(0.02M)", sol2_vol=1)`, # adding 1 mL of the reagent 'HCl(0.02M)' to 'test_2A', labeling the result as 'test_2B'
+        `add_a_solution(test_label="A_B_HCl", sol1_label="A_B", sol2_label="HCl(6M)", sol2_vol=5)`, # adding 5 mL of the reagent 'HCl(6M)' to 'A_B', labeling the result as 'A_B_HCl'
+        `add_a_solution(test_label="test_3", sol1_label="test_1", sol2_label="test_2", sol2_vol=2)`, # adding 2 mL of 'test_2' to 'test_1', labeling the result as 'test_3'
+    ]
```
MrtinoRG, line 829: Again I would remove the comments

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+                    [ERROR_DETAILS] The given `sol2_vol` is not an integer or it is less than 1 mL [/ERROR_DETAILS]
+                    [ERROR_RECOVERY] Make sure you are adding at least 1 mL of the second solution [/ERROR_RECOVERY]
+        
+        VolumeError: [ERROR_WHEN] When `sol2_label` contains a precipitate or its remaining volume is less than the requested amount [/ERROR_WHEN]
+                     [ERROR_DETAILS] The given `sol2_label` was found in the Inventory and it is a solution but it either contains a precipitate or there is not enough of it remaining [/ERROR_DETAILS]
+                     [ERROR_RECOVERY] If the second solution contains a precipitate, filter it using the `filter_solution` tool. If the remaining amount of it is less that the requested amount, try doing the test will a smaller `sol2_vol` if possible, or make more of that solution by repeating the steps that lead to it. [/ERROR_RECOVERY]
+    [/RAISES]
+
+    [LIMITATIONS] Known Limitations:
+    - The host solution may contain precipitates but the second solution (the one being added) cannot contain any precipitates.
```
MrtinoRG: Here you switch into using host and second solution. Perhaps I would add "soln" labels within paretheses

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    - The minimum allowed volume for the second solution is 1 mL.
+    - The reported colors are qualitative and approximate.
+    - The perceived color of precipitates will depend on the composition of the precipitated solids. If more than compound co-precipitate at the same time, the color may be different from the color of pure precipitates.
+    - The perceived color of solutions will depend on the concentraion of species in that solution. Both the hue and the lightness of the perceived color can change as the concentration of species in the solution change.
```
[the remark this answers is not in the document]
MrtinoRG: @aaaghajani

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    - Any observations mentioning that an existing precipitate "partially" or "mostly" dissolves are qualitative and approxiamte statements.
```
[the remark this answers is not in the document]
MrtinoRG: @aaaghajani here is correct too :)

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+        sol2 = compositions[sol2_label]
+    except KeyError as e:
+        raise KeyError(f"Invalid sol2_label: {e}")
+    
+    if not isinstance(sol2, StockSolution):
+        raise RuntimeError(f"{sol2_label} is not a solution!")
+
+    if sol1.volume == 0:
+        raise VolumeError(f"The remaining volume of {sol1_label} is zero!")
+
```
MrtinoRG, line 910: I think you have repeated this code through several functions. Perhaps you can move it into a utils function

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    [CONTEXTUAL] How this tool works:
+    - It seprates the aqueous phase and the solid phase from the `label`, and removes it from the inventory.
+    - The supernatant is added to the Inventory, with the original label appended by '_filtrate'.
+    - The precipitate is added to the Inventory, with the original label appended by '_precipitate'. [/CONTEXTUAL]
+
+    [SYNTACTICAL] Usage examples:
+    [
+        `filter_solution(label="test_2B"), # 'test_2B' is removed from the Inventory and is replaced by 'test_2B_filtrate' and 'test_2B_precipitate'
+    ]
```
MrtinoRG, line 1029: Here again (I am commenting for the cases I believe more examples can be provided)

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    1. [PREREQUISITE] You need to have a separated precipitate in the Inventory before calling this tool. Separate the precipitates by using the `filter_solution` tool. [/PREREQUISITE]
+    2. [CURRENT] Use this tool to add the precipitate to a given volume of the solution. [/CURRENT]
+    3. [FOLLOW_UP] You can use your chosen `test_label` to refer to the resulting solution when performing further tests. If the added precipitate does not fully dissolve, you can add other solutions to the mixture by calling the `add_a_solution` tool or filter the precipitate again using the `filter_solution` tool. If the added precipitate fully dissolves, you can perform further tests on the resulting solution by calling tools such as `measure_pH`, `mix_two_solutions`, `add_a_solution` or `perform_flame_test`. [/FOLLOW_UP] [/WORKFLOW_INTEGRATION]
+
+    [CONTEXTUAL] How this tool works:
+    - It draws `sol_vol` mL from `sol_label`.
+    - Adds it to a new empty container labeled `test_label`.
+    - Adds all of the precipitate labeld `prec_label` to the container and stirs the mixture until equilibrium is reached. The resulting mixture is added to the Inventory.
```
MrtinoRG: Yeah, like the label stuff is also super confusing. I am not saying is wrong, but that it can be confusing and I do not know a better solution that the one that you used

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    - When an observation mentions partial dissolution, it roughly means that about 20-50% of the added precipitate has dissolved.
+    - When an observation mentions that a precipitate has mostly dissolved, it means that more than 50% of it has dissolved but there is still some undissolved precipitate remaining.
+    - Sometimes it is possible that the precipitate's color can be described using more than one color name. In those cases the different given names will be separated by a slash '/'.
+    - Remember that the reported colors are qualitative and approximate. [/CONTEXTUAL]
+
+    [SYNTACTICAL] Usage examples:
+    [
+        `add_precipitate_to_solution(test_label="test_3", prec_label="test_2_precipitate", sol_label="test_1", sol_vol=10)`, # adding all of the precipitate 'test_2_precipitate' to 10 mL of the solution 'test_1', labeling the result as 'test_3'
+        `add_precipitate_to_solution(test_label="prec3_HCl", prec_label="test3_precipitate", sol_label="HCl(1M)", sol_vol=20)`, # adding all of the precipitate 'test3_precipitate' to 20 mL of the reagent solution 'HCl(1M)', labeling the result as 'prec2_HCl'
```
MrtinoRG: Here again I would add more examples, and remove the comments

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+                    [ERROR_RECOVERY] Make sure you are using at least 4 mL of the solution and passing it as an integer value [/ERROR_RECOVERY]
+        
+        VolumeError: [ERROR_WHEN] When `sol_label` contains a precipitate or its remaining volume is less than the requested amount [/ERROR_WHEN]
+                     [ERROR_DETAILS] The given `sol_label` was found in the Inventory and it is a solution but it either contains a precipitate or there is not enough of it remaining [/ERROR_DETAILS]
+                     [ERROR_RECOVERY] If the solution contains a precipitate, filter it using the `filter_solution` tool. If the remaining amount of it is less that the requested amount, try doing the test will a smaller `sol_vol` if possible, or make more of that solution by repeating the steps that lead to it. [/ERROR_RECOVERY]
+    [/RAISES]
+
+    [LIMITATIONS] Known Limitations:
+    - The solution may not contain any precipitates.
+    - The minimum allowed volume for the solution is 4 mL.
```
MrtinoRG, line 1175: Why here is 4?

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    if type(prec) != Precipitate:
+        raise RuntimeError(f"{prec_label} is not a precipitate!")
+    if type(sol) not in [Solution, StockSolution]:
+        raise RuntimeError(f"{sol_label} is not a solution!")
+    if type(sol) == Solution:
+        if sol.has_precipitate: 
+            raise VolumeError(f"{sol_label} already has a precipitate. If you want to add a different precipitate, you must filter it first!")
+
+    if type(sol_vol) != int or sol_vol < 4:
+        raise ValueError(f"`sol_vol` must be an integer greater than or equal to 4")
```
MrtinoRG, line 1202: Yeah, this code is in the previous tool (I think), so I would move it to the utils

tasks/wetlab/wetlab/tools.py
```diff
@@ -0,0 +1,1303 @@
...
+    [DETAILED] Returns a string cotaining the details of the solutions and precipitates in the Inventory, including their remaining volumes in mL [/DETAILED]
+
+    [PROCEDURAL] When to use this tool:
+    - Typically used in the beginning to check the unknown samples.
+    - You can use this tool to check the Inventory, which is effectively a summary of the experiments performed so far. 
+    - Use this tool to check the remaining volumes of the samples and test solutions.
+    - Use this tool to check the correct label of solutions and precipitates to use in other tools. [/PROCEDURAL]
+
+    [WORKFLOW_INTEGRATION] Typical workflow integration:
+    1. [PREREQUISITE] None [/PREREQUISITE]
```
MrtinoRG: Here again perhaps I would add something indicating how to interpret the input data

MrtinoRG (commented):
I've missed (irony on) these long PRs

MrtinoRG (commented):
I've missed (irony on) these long PRs

tasks/wetlab/wetlab/tools.py
aaaghajani: yeah, I changed it to 1

tasks/wetlab/wetlab/tools.py
aaaghajani: Thank you for letting me know, I added more examples

tasks/wetlab/wetlab/tools.py
aaaghajani: I don't think this is "too many". It's just that this env is very open-ended and there are many paths that the agent can take. They still struggle even with this many hints :D

tasks/wetlab/wetlab/tools.py
aaaghajani: I added those, but why remove the comments?

tasks/wetlab/wetlab/tools.py
aaaghajani: They are a bit confusing to humans, but the agents don't seem to be confused at all, based on the test runs

tasks/wetlab/wetlab/tools.py
aaaghajani: to have enough liquid so that the precipitate can dissolve as much as it can and not be limited by the amount of solution. Since most tests are done with 1-2 mL of each solution, the precipitates are formed from ~4-5 mL of solution.

tasks/wetlab/wetlab/tools.py
aaaghajani: they are very slightly different, so I didn't bother with making it a util function

tasks/wetlab/wetlab/tools.py
aaaghajani: There is no "input data". I left it as None to encourage the agent to call this tool right away at the beginning

tasks/wetlab/wetlab/tools.py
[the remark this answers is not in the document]
kjappelbaum: @aaaghajani not sure if you already went through the copilot suggestions. A bunch of them seem quite reasonable.

tasks/wetlab/wetlab/tools.py
kjappelbaum: Sorry - i was not clear here - in this comment (and some others below) I did not ask for clarification for myself. But I asked if there is something we can clarify for the agent or via a comment/docstring for our future selves. 

Here, I wonder if the fact that you call it differently in different places might make it more confusing (in a minor way) than necessary

tasks/wetlab/wetlab/tools.py
kjappelbaum: This is an example of where a comment in the code would help - this relies on the fact that you have a special method implemented in. It would be useful for the reader of the code to be reminded of that and that this is the reason you do it. `1 *`

tasks/wetlab/wetlab/tools.py
kjappelbaum: This, again, would be a nice example for a comment as the question I had when reading the code. 

However, it also makes one wonder why you not always just take the elemental symbol. This would make the code future-proof and eliminate the need for special treatments.

tasks/wetlab/wetlab/tools.py
kjappelbaum: and here, too, a small comment about the fact that multiplication is a special thing here would help the reader of the code

tasks/wetlab/wetlab/env.py
kjappelbaum: no, the `_add_task_tools`

tasks/wetlab/wetlab/engine.py
kjappelbaum: In this case, It would be good to open an issue to track this

tasks/wetlab/wetlab/engine.py
kjappelbaum: My intuition would have been to add all the white ones. 

If you stick with this default (which is ok), I believe that we need to communicate prominently that there is such a default in the code.

kjappelbaum:
@aaaghajani, with these larger PRs, the reviewers will need a bit more of your help. If you can let us know what you've already implemented and when you need re-review, it would be very helpful.

If you implement a suggestion in a specific commit, it can be convenient to comment "implemented in <link to commit>". If you choose not to implement a change, please state that explicitly.

This way, reviewers can go through the comments and resolve the ones where they believe the issue has been addressed.

tasks/wetlab/wetlab/tools.py
MrtinoRG: Like I am suggesting this such that all the environments have a similar level of description. This would make the comparisons more robust, despite how difficult these comparisons are for different environments

tasks/wetlab/wetlab/tools.py
MrtinoRG: Again, to keep a similar level of description to the other environments

tasks/wetlab/wetlab/tools.py
MrtinoRG: Great!

tasks/wetlab/wetlab/tools.py
MrtinoRG: But you have two tools with `None` as prerequisite so perhaps is still not clear which one should be called first

tasks/wetlab/wetlab/tools.py
aaaghajani: changed the wording to "pH-indicator paper" to be clear

tasks/wetlab/wetlab/tools.py
aaaghajani: added clarification comment

tasks/wetlab/wetlab/tools.py
aaaghajani: added comment (commit not pushed yet)

tasks/wetlab/wetlab/tools.py
aaaghajani: changed 10 to 1 and added comment

tasks/wetlab/wetlab/tools.py
aaaghajani: removed the task-dependent bit from docstring

tasks/wetlab/wetlab/tools.py
aaaghajani: removed the * (commit not pushed yet)

tasks/wetlab/wetlab/tools.py
aaaghajani: added explicit "greater than or equal to 1" in all cases

tasks/wetlab/wetlab/tools.py
aaaghajani: added clarification comment

tasks/wetlab/wetlab/tools.py
aaaghajani: change the wording in all "main tools" to "This tool lets you perform ..."

tasks/wetlab/wetlab/tools.py
aaaghajani: changed to "in the beginning of a task"

tasks/wetlab/wetlab/tools.py
aaaghajani: I have already done some runs with the comments, I guess it does not make sense now to remove them for the rest of the runs because this will cause inconsistency. But the reason I added the comments is that I saw similar comments in the catalyst env

tasks/wetlab/wetlab/tools.py
aaaghajani: added clarification in parantheses

aaaghajani:
The force-pushes are due to me being a newbie with Git... I pulled main, did some changes and pushed but `react.py` and `utils.py` were still the old version. I reverted the commits and fixed those two files. But then I couldn't push because I had reverted the commits that I had already pushed and now I was behind origin, so I had to force-push. The changes in `src` are from Nawaf's last commit that introduced the hooks, I didn't make any changes.

tasks/wetlab/wetlab/QA/wetlab_qa_01.json
kjappelbaum, line 1: Is this the full QA?

kjappelbaum:
I guess the runs can be removed from this PR and should be on HuggingFace instead?

PR #250, opened 2025-12-24
fix: llm response object content in reasoning

n0w0f:
Added this because content in response object from GPT-OSS is null and the completion is under reasoning key

## Summary by Sourcery

Normalize LLM response messages to handle reasoning-model outputs where content may be null by extracting text from reasoning-specific fields before downstream usage.

Bug Fixes:
- Handle cases where LiteLLM responses from reasoning models provide text under reasoning or reasoning_content when content is null to avoid empty message content.

Enhancements:
- Introduce a helper utility to centralize extraction of message content from LiteLLM messages, including support for reasoning-specific fields.

### DIFF
--- a/src/corral/agents/utils.py
+++ b/src/corral/agents/utils.py
@@ -49,6 +49,42 @@ class LiteLLMMessage(TypedDict, total=False):
     name: str | None
 
 
+def extract_message_content(message: Message) -> str:
+    """
+    Extract content from a LiteLLM message, handling both regular content
+    and reasoning model responses.
+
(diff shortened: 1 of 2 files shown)


MrtinoRG:
The changes here are already present in #236 , so I would suggest to close this one

kjappelbaum:
@n0w0f do you agree with @MrtinoRG's suggestion or do you see any problems with doing it this way?

n0w0f:
> @n0w0f do you agree with @MrtinoRG's suggestion or do you see any problems with doing it this way?

I agree. I am closing this.

PR #249, opened 2025-12-23
Generate latex

chandan21gupta:
env_descriptions.py contains code to generate latex for environments : task and tool tables.

PR #262, opened 2026-01-08
feat: add code for `to_latex()` method

MrtinoRG:
## Summary by Sourcery

Add LaTeX documentation generation for tasks and tools across environments, HTTP API, and benchmarking, plus supporting utilities and small tooling fixes.

New Features:
- Introduce an env.to_latex() method to generate LaTeX task and tool documentation from environments using Code2Latex.
- Expose HTTP endpoints and router methods to generate LaTeX for a task and to clear LaTeX cache, and integrate LaTeX generation into the bench workflow behind a flag.

Enhancements:
- Add a Code2Latex utility module with caching and LaTeX rendering helpers for task colorboxes and tools longtables.
- Export the new code2latex utility from the corral.utils package.
- Refine tool description tags in spectra_elucidation to use RETURNS_* markers and update Modal function lookup usage.


MrtinoRG:
Hi, 

I created this as I found it easier than do it in the other PR (#249), mostly because it is very difficult at this point to merge that branch with `dev` (I guess at some point we will have to remove `q2label` and take the useful stuff from there to a new branch. 

The code works that first it is needed to be run only one task, to then run the subtasks corresponding to that task. It is not that automatic, but it is very difficult to link tasks and subtasks otherwise.

src/corral/backend/env.py
```diff
@@ -410,3 +411,268 @@ def configure_additional_apps(self):
...
+            level: Task level identifier (e.g., 1, 2, "advanced")
+            env_name: Environment name (e.g., "afm", "catalyst"). If not provided,
+                     will try to get from task_group.group_id
+            task_name: Optional custom name for the task (defaults to task_id)
+
+        Returns:
+            Tuple of (task_tex_path, tools_tex_path, scoring_tex_path) - paths to the generated .tex files
+        """
+        from corral.router.verbosity import ToolVerbosity, VerbosityConfig
+        from corral.utils.code2latex import CacheMetadata, Code2Latex
```
kjappelbaum: is this here and not at the top of the file to avoid circular imports?

src/corral/backend/env.py
```diff
@@ -410,3 +411,268 @@ def configure_additional_apps(self):
...
+
+        Args:
+            output_dir: Directory for output .tex files
+            level: Task level identifier (e.g., 1, 2, "advanced")
+            env_name: Environment name (e.g., "afm", "catalyst"). If not provided,
+                     will try to get from task_group.group_id
+            task_name: Optional custom name for the task (defaults to task_id)
+
+        Returns:
+            Tuple of (task_tex_path, tools_tex_path, scoring_tex_path) - paths to the generated .tex files
```
kjappelbaum, line 548: personally, I do not like tuples as return type. `namedtuples` or something like that seem to be way cleaner

tasks/spectra_elucidation/spectra_elucidation/tools.py
```diff
@@ -66,9 +66,9 @@ def get_formula_from_smiles(smiles: str) -> str:
 
     Returns:
         str:
-            [ARGS_BRIEF] The chemical formula in Hill notation (C, H, then alphabetical) [/ARGS_BRIEF]
-            [ARGS_DETAILED] The chemical formula of the molecule represented by the SMILES string, formatted in Hill notation. If the SMILES string is invalid or cannot be parsed, it returns an error message. [/ARGS_DETAILED]
-            [ARGS_EXAMPLES] '"C2H6O" for ethanol, "C6H6" for benzene, "C2H5NO" for acetic acid amide, "C7H6O3" for salicylic acid' [/ARGS_EXAMPLES]
+            [RETURNS_BRIEF] The chemical formula in Hill notation (C, H, then alphabetical) [/RETURNS_BRIEF]
+            [RETURNS_DETAILED] The chemical formula of the molecule represented by the SMILES string, formatted in Hill notation. If the SMILES string is invalid or cannot be parsed, it returns an error message. [/RETURNS_DETAILED]
+            [RETURNS_EXAMPLES] '"C2H6O" for ethanol, "C6H6" for benzene, "C2H5NO" for acetic acid amide, "C7H6O3" for salicylic acid' [/RETURNS_EXAMPLES]
```
kjappelbaum, line 67: might this have caused problems in some agent runs?

src/corral/run.py
```diff
@@ -494,6 +539,12 @@ def bench(
             configure_timeout: Timeout for configuring additional apps.
             run_name: Name for the benchmark run (used in report filename).
                      If None, defaults to "unknown_env".
+            generate_latex: If True, generates LaTeX documentation for all tasks
```
kjappelbaum: do we really need it here? is this a realistic use case that we want the latex docs before running the benchmark? It clutters the arguments of the function a lot for a use case that seems to not be super common

reports/tests/run_react_agent.py
kjappelbaum, line 1: should we keep this test in the PR?

kjappelbaum:
This PR is quite close to being mergeable. It would probably be good to see which of the comments need to be implemented and then merge it to avoid divergence.

src/corral/run.py
MrtinoRG: Yes I know that this is quite close to merging but I do not know exactly how to address this. 

Like what are the alternatives?

MrtinoRG:
@kjappelbaum now I added the generate_latex method in the runner such that one can simply create the Latex if desired without benching the envs. 

I do not know if this is what you had in mind but it is cleaner. Did not think about this before.

When I added in the bench method it was partially because in that way one would get the latex such that it is possible to verify if what is being bench is what one desires, but I guess in the current way this is also possible even without running 🙃

kjappelbaum:
> @kjappelbaum now I added the generate_latex method in the runner such that one can simply create the Latex if desired without benching the envs.
> 
> I do not know if this is what you had in mind but it is cleaner. Did not think about this before.
> 
> When I added in the bench method it was partially because in that way one would get the latex such that it is possible to verify if what is being bench is what one desires, but I guess in the current way this is also possible even without running 🙃

yes, I felt it would be good to generate LaTeX without benching the envs

reports/tests/run_react_agent.py
kjappelbaum: @MrtinoRG  those files can probably go

src/corral/utils/code2latex.py
```diff
@@ -0,0 +1,593 @@
+"""
+Code2Latex - Automatic LaTeX documentation generation from Corral task environments.
+
+This module provides utilities for generating formatted LaTeX files (with colorboxes)
+containing task descriptions, tools, and scoring information. All data is kept in memory
+and lost at the end of the run.
+"""
+
+from __future__ import annotations
```
kjappelbaum: I feel I asked this before, but what versions of Python do we want to support? Do we really need this futures stuff?

src/corral/utils/code2latex.py
```diff
@@ -0,0 +1,593 @@
...
+
+from dataclasses import asdict, dataclass
+from pathlib import Path
+from typing import TYPE_CHECKING, Any, ClassVar
+
+from loguru import logger
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
```
kjappelbaum: It is strange to hide this variable but not the others. Do we really need to protect it from always being imported?

reports/tests/run_react_agent.py
MrtinoRG: They are removed...

src/corral/utils/code2latex.py
MrtinoRG: But it does not hurt right? Like it simply makes the type annotations strings (to my understanding)

MrtinoRG: Ok, I see it is kind of a "failed experiment" the implementation of this feature into Python coding

reports/tests/run_react_agent.py
kjappelbaum: ups, sorry. I did not realize that is a deletion. Sorry!

src/corral/backend/env.py
```diff
@@ -410,3 +411,268 @@ def configure_additional_apps(self):
...
+
+    def to_latex(
+        self,
+        output_dir: str,
+        level: int | str,
+        env_name: str | None = None,
+        task_name: str | None = None,
+    ) -> tuple[str, str, str | None]:
+        """
+        Generate LaTeX documentation for this task.
```
kjappelbaum, line 524: minor: Perhaps we should add a note what docstring format this requires

src/corral/backend/env.py
```diff
@@ -410,3 +411,268 @@ def configure_additional_apps(self):
...
+
+        This method creates a TaskDefinition from the environment's task data
```
kjappelbaum:
```suggestion
        This method creates a `TaskDefinition` from the environment's task data
```

src/corral/backend/env.py
```diff
@@ -410,3 +411,268 @@ def configure_additional_apps(self):
...
+        and delegates to Code2Latex.colorbox() for generating formatted LaTeX files
```
kjappelbaum:
```suggestion
        and delegates to `Code2Latex.colorbox()` for generating formatted LaTeX files
```

src/corral/backend/env.py
```diff
@@ -410,3 +411,268 @@ def configure_additional_apps(self):
...
+        and Code2Latex.longtable() for generating tools documentation.
```
kjappelbaum:
```suggestion
        and `Code2Latex.longtable()` for generating tools documentation.
```

src/corral/backend/env.py
```diff
@@ -410,3 +411,268 @@ def configure_additional_apps(self):
...
+
+        If the environment has a `current_task` attribute (e.g., TaskGroupEnvironment),
```
kjappelbaum:
```suggestion
        If the environment has a `current_task` attribute (e.g., `TaskGroupEnvironment`),
```

src/corral/backend/env.py
```diff
@@ -410,3 +411,268 @@ def configure_additional_apps(self):
...
+        it will automatically detect:
+        - Whether this is a subtask (based on input_from_tasks)
```
kjappelbaum:
```suggestion
        - Whether this is a subtask (based on `input_from_tasks`)
```

src/corral/backend/env.py
```diff
@@ -410,3 +411,268 @@ def configure_additional_apps(self):
...
+        - Dependencies on other tasks
+        - env_name from task_group.group_id if not provided
```
kjappelbaum:
```suggestion
        - `env_name` from `task_group.group_id` if not provided
```

src/corral/backend/env.py
```diff
@@ -410,3 +411,268 @@ def configure_additional_apps(self):
...
+                     will try to get from task_group.group_id
+            task_name: Optional custom name for the task (defaults to task_id)
+
+        Returns:
+            Tuple of (task_tex_path, tools_tex_path, scoring_tex_path) - paths to the generated .tex files
+        """
+        from corral.router.verbosity import ToolVerbosity, VerbosityConfig
+        from corral.utils.code2latex import Code2Latex, LatexMetadata
+
+        # Get task description from prompt
```
kjappelbaum, line 568: to be honest, such comments do not help. I can read the same a line below. 
It is more interesting, for instance, to have comments about the "why". For example, why do I need to cast it to `str`?

src/corral/backend/env.py
```diff
@@ -410,3 +411,268 @@ def configure_additional_apps(self):
...
+        description = str(self.get_task_prompt())
+
+        # Get list of tool names
+        tools = list(self.tools.keys())
+
+        # Get detailed tool information for longtable with DETAILED verbosity parsing
+        # This parses the descriptions similarly to server.py, using the Detailed level
```
kjappelbaum: I am not sure if this comment would help anyone at a later point. With this clear link to one file it might also be easily outdated

src/corral/backend/env.py
```diff
@@ -410,3 +411,268 @@ def configure_additional_apps(self):
...
+        for tool in self.tools.values():
+            filtered_description = VerbosityConfig.filter_tool_description(
+                tool.description, verbosity
+            )
+
+            # Extract RETURNS section from the original description and filter it
+            sections = VerbosityConfig.extract_all_sections(tool.description)
+            returns_parts = [
+                sections.get("RETURNS_BRIEF", ""),
+                sections.get("RETURNS_DETAILED", ""),
```
kjappelbaum: perhaps I missed it - but why did you hardcode those two verbosity levels?

src/corral/backend/env.py
```diff
@@ -410,3 +411,268 @@ def configure_additional_apps(self):
...
+            pass
+
+        return {
+            "name": fn_name,
+            "description": description,
+            "arguments": structured_args,
+            "returns": returns_section,
+        }
+
+    def to_latex(
```
kjappelbaum, line 515: it is, of course, a bit annoying to ask for it, but the codebase would be stronger if we also had tests for things like that.

kjappelbaum:
@sourcery-ai dismiss

kjappelbaum (approved):
I only left some minor comments, and I trust your judgment to decide which ones are relevant, which ones should be dismissed, and which ones should be new issues. 

Thanks for pushing this and for being patient with the review!

PR #265, opened 2026-01-15
Fix issue #264: Pythoncom missing in the dependencies of AFM

imandal98:
Implemented safe `pythoncom` import (Windows only)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cross-platform stability by preventing Windows-only system initialization on non-Windows platforms while preserving behavior on Windows.
  * Applied guarded initialization and cleanup across scanning, analysis, visualization, optimization, and parameter-handling workflows to avoid platform-specific errors and ensure safe cleanup.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/tasks/afm/src/env.py
+++ b/tasks/afm/src/env.py
@@ -2,9 +2,10 @@
 import gc
 import json
 import os
+import platform
 from collections.abc import Callable
 from pathlib import Path
-import platform
+
 import nanosurf
 from loguru import logger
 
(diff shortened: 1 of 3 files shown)


MrtinoRG:
Yeah, now seems to work fine. Thanks @imandal98 
 
I understand that this env will only work when the equipment is available, and the equipment only works with Windows OS. 

This is fine, but when migrating to its own package I would add a note that reflects this.

PR #266, opened 2026-01-16
feat: llm response object and messages have id

n0w0f:
## Summary by Sourcery

Introduce a structured LLMResponse wrapper that carries message metadata (IDs, logprobs, and token usage) and propagate stable message IDs through agents, serialization, and benchmarking utilities.

New Features:
- Add LLMResponse wrapper object encapsulating raw LLM messages along with optional metadata such as ID, logprobs, and usage.

Enhancements:
- Change llm_call to always return an LLMResponse, capturing response IDs, optional logprobs, and usage metadata based on call parameters.
- Extend LiteLLMMessage and message serialization to support message IDs and ensure agents (tool-calling, ReAct, planner, base agent, reflection) propagate and store assistant message IDs for later use.
- Adjust benchmarking script to log logprobs with associated content, tweak sampling/benchmark parameters, support multiple verbosity levels, and rename output directories to encode verbosity and agent type.
- Update test tooling configuration to use an argument-list style schema for tools instead of a JSON-schema-like parameters object.
- Refine base agent and reflection logic to read token usage from LLMResponse metadata instead of tuple returns and to handle missing content safely.

Tests:
- Update and extend unit tests and mocks to use the LLMResponse wrapper, cover logprobs and usage metadata handling, and validate serialization of message IDs.

### DIFF
--- a/reports/resistor_networks/gpt_oss_single/run_react.py
+++ b/reports/resistor_networks/gpt_oss_single/run_react.py
@@ -45,6 +45,7 @@ def logprobs_hook(context: HookContext) -> None:
     lp_data = {
         "id": llm_response.id,
         "task_id": context.task_id,
+        "content": llm_response.content,
         "iteration": context.iteration,
         "logprobs": lp_dict,
     }
@@ -160,7 +161,7 @@ def setup_litellm():
 def run_benchmark(
     model: str = "claude-3-5-sonnet-20241022",
     _task_ids: list | None = None,
-    temperature: float = 0.7,
+    temperature: float = 0,
     run_name: str = "corral_benchmark_run",
     verbose: str = "brief",
     hooks=None,
@@ -176,7 +177,7 @@ def run_benchmark(
     agent = ReActAgent(
         model=model,
         logprobs=True,
-        top_logprobs=2,
+        top_logprobs=20,
         max_iterations=20,
         temperature=temperature,
         api_endpoint="https://api.helmholtz-blablador.fz-juelich.de/v1",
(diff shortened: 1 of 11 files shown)


src/corral/agents/utils.py
```diff
@@ -115,9 +182,16 @@ def llm_call(
                 # Ensure reasoning_content is a string
                 message.content = str(reasoning_content) if reasoning_content else None
 
+        metadata = LLMResponseMetadata()
+        metadata["id"] = response.id
+
+        # Include logprobs if requested via kwargs
```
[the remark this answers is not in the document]
n0w0f: This solution also rely on `logprobs` in LiteLLM ? @sourcery-ai

MrtinoRG (approved):

reports/resistor_networks/gpt_oss_single/run_react.py
```diff
@@ -160,7 +161,7 @@ def setup_litellm():
 def run_benchmark(
```
MrtinoRG, line 161: @n0w0f above, in line 76, why do you divide by the len(probs)? I guess it is to calculate the mean of the entropy for the entire generation? But you have that below and I agree more with how that (avg_top_entropy) is calculated. Like that `entropy` I do not know what it is and I would keep in mind that it is not useful at all.

Also, what do you think about adding the tail of logprobs that we do not have to not bias the calculation to much to the top that we have. This is add `probs_tail` as: `probs_tail = 1 - top_probs` and the entropy would end up being calculated as:

`top_ent = -np.sum(probs * np.log(probs + 1e-10)) - (probs_tail * np.log(probs_tail + 1e-10)`

Also, I would also calculate the perplexity as a measure of uncertenty. This is: `perplexity = np.exp(-avg_logprob)`. I would perhaps even also calculate the perplexity per token

This is not urgent as we can calculate this afterwards with the stored top_logprobs :)

n0w0f: I agree with you that `avg_top_entropy` make more sense, and currently we use that for the analysis. `entropy` here as  you said is for the entire generation, when i added it for the first time, i thought this could be useful for analysis, but ended up not using it. Without context, i agree one can be misled here.

For now i just again get all those metrics from the log prob file, not really use those metrics file. I am still getting intuitions on the metrics you suggested, feel free to add them already,  i think people are still yet to start the runs

PR #271, opened 2026-01-21
feat: load from trace and fix toolcalling intervention

n0w0f:
## Summary by Sourcery

Introduce trace-based intervention hooks and relax CORRAL_WORK_DIR requirements across tasks while improving CLI ergonomics for task servers.

New Features:
- Add support for creating intervention hooks from recorded trace files that can replay ReAct and tool-calling assistant steps with optional tool execution.

Bug Fixes:
- Fix tool-calling intervention behavior so recorded tool_calls are injected in an OpenAI-compatible format and executed reliably when requested.

Enhancements:
- Simplify BASE_WORK_DIR handling for catalyst, ML, and resistor network tasks by centralizing its definition and logging its usage.
- Improve task server CLIs to take the tasks JSON path and optional port as positional arguments and log the final host/port configuration instead of relying on CORRAL_TASKS_PATH defaults.
- Generate per-run fallback CORRAL_WORK_DIR locations for scoring scripts when the environment variable is unset, avoiding hard failures and work-dir collisions.


src/corral/agents/hooks/__init__.py
```diff
@@ -6,12 +6,16 @@
...
+    create_trace_intervention_hook,
+)
 
 __all__ = [
     "AgentHooks",
     "CriticalHookError",
     "HookContext",
     "HookPoint",
     "create_intervention_hook",
+    "create_trace_intervention_hook",
```
MrtinoRG, line 20: Why you need a new one?

tasks/catalyst/src/catalyst/score.py
```diff
@@ -1,5 +1,7 @@
 import json
```
MrtinoRG, line 1: And what are the reasons for these changes? I guess because in this case we do not have different trials? But would not be better to have something common instead of modying all the envs?

Adding datetime would solve the problem, right?

src/corral/agents/hooks/intervention.py
```diff
@@ -193,3 +234,211 @@ def intervention_hook(context: HookContext) -> None:
         _intervention_registry.execute(context, intervention_thought, execute_tools)
 
     return intervention_hook
+
+
+def _inject_react_step(
```
MrtinoRG, line 239: For this and the one for tool calling, why do you need such amount of code? 

I mean in principle you could simply pass the mesaages as they are and they should work, as the messages follow the same format

MrtinoRG: Like why you cannot simply take step 1 and all the messages following this that it is not assistant? Then for step or message 2 the same

MrtinoRG: Perhaps I am missing why you need to parse again the actions a nd everything

src/corral/agents/hooks/intervention.py
```diff
@@ -193,3 +234,211 @@ def intervention_hook(context: HookContext) -> None:
...
+            content,
+            flags=re.DOTALL,
+        ).strip()
+
+        context.messages.append(
+            LiteLLMMessage(role="assistant", content=stripped_content)
+        )
+
+
+def _inject_toolcalling_step(
```
MrtinoRG, line 297: Also, how did you run the initial experiments without these functions?

MrtinoRG (commented):
I do not know why you need that much overhead in parsing again the messages, but probably there is a good reason

src/corral/agents/hooks/__init__.py
n0w0f: this so that i could load from a trace file

tasks/catalyst/src/catalyst/score.py
n0w0f: this paths could have caused problem. because when i randomly pick traces and run the experiment. The path in those traces could be different from the current working dir. This is an interim solution

src/corral/agents/hooks/intervention.py
n0w0f: this is because in some traces the syntax model produced is wrong. and the observation after execution should be same to that of the trace

src/corral/agents/hooks/intervention.py
n0w0f: I ran only react in the initial experiment.
Now we plan to run both

MrtinoRG: Yeah, but in this PR there is also the function to add the ReAct...

src/corral/agents/hooks/intervention.py
MrtinoRG: But, then why you need to proccess them? Like you can simply pass the same from that trace right?

tasks/catalyst/src/catalyst/score.py
MrtinoRG: This path stuff is not ideal...

src/corral/agents/hooks/intervention.py
n0w0f: the additions in ReAct is of minimal difference functional wise.
In the initial way of doing things, I would give steps as a dict, and then it is wrapped accordingly(as thoughts and actions) but then there is possibility if issue with this as in original scenario there could be syntax error.
With the current implementation it takes as it is and agent see the same use role based observation about syntax not being correct

tasks/catalyst/src/catalyst/score.py
n0w0f: An ideal file system implementation for the framework would require discussion and some thinking.

src/corral/agents/hooks/intervention.py
n0w0f: There is also support to vary the number of steps to be shown from the trace. The only other processing here is to take assistant based steps from the trace.
Then later there is also implementation for stripping only thought if `execute_tools=False`

MrtinoRG (approved):

PR #275, opened 2026-01-27
AFM scoring function logic

imandal98:
## Summary by CodeRabbit

* **Refactor**
  * Standardized aggregate performance metrics and numeric types across many reports; several top‑level metrics updated to computed or uniform values.

* **Improvements**
  * Added broader tooling and usage aggregates (tool calls, token/tokenization totals, execution and benchmark durations).
  * Enhanced score normalization to better align reported magnitudes.

* **Chores**
  * Simplified experiment configurations by removing several tuning parameters and a deprecated tip entry.
  * Consolidated tool list formatting.


reports/afm/Task/task_1_claude-sonnet-4-5_ToolCalling/claude-sonnet-4-5-Tool_calling-task_1-workflow_verbosity.json
sajidmannan, line 1: I checked each file, and the agents seem to consistently return results in meters. Therefore, I’m not sure we need to explicitly ask for unit conversions in the prompt. This raises the question of whether the tool being used always returns outputs in meters or in some other default unit. If that is the case, then this may indicate an issue with the agent, as it appears to be performing an additional task beyond what was asked not the prompt.

reports/afm/Task/task_1_gpt_4o_ToolCallling/gpt-4o_Tool_calling-task_1-brief_verbosity.json
sajidmannan, line 1: @imandal98 Can you check the documentation to clarify unit things? If read function you use in tool has any standard unit of return ? I don't see any unit specific related things in tool.

sajidmannan (commented):
Overall, I can see that all the changes have been made, and everything looks fine to me. My only concern is about the units: if the read function used here always returns values in meters, then I don’t see the need to explicitly asked unit conversion in the prompt. The agent should return results in the default unit. If agent is changing the units, that means the agent is performing an additional task beyond the instruction, and this behavior should be penalized. At the moment, all the scores appear to be 1, and the environment seems trivial.

MrtinoRG:
Is this the final state @imandal98?



Can you change the branch of destiny above from `main` to `dev`?

imandal98:
Closing this PR because the target branch was incorrect (main).
A new pull request has been opened against dev: #288

Please continue the review there.

PR #285, opened 2026-02-09
fix: solve scoring for the spectra and retro subtasks

MrtinoRG:
## Summary by Sourcery

Fix scoring logic for spectra hydrogen environments and retrosynthesis subtask targets, and regenerate associated task and report JSONs.

Bug Fixes:
- Correct hydrogen environment counting in spectra scoring to distinguish non-exchangeable and diastereotopic proton environments.
- Fix retrosynthesis subtask generation so each build_complete_route task uses the intended set of final targets instead of an incorrect flattened list.

Chores:
- Regenerate retrosynthesis and spectra subtask definitions and model evaluation report JSONs to reflect the corrected scoring and target configuration.


n0w0f (approved):

n0w0f:
There seems to be some tests failing, probably from the scoring function changes.

PR #287, opened 2026-02-14
afm and md qa - final version, commends already addressed

chandan21gupta:
These are final QAs for MD and AFM. All the comments are addressed and hence should be directly used without any additional review. 
Made this new PR so that it is based on dev.


imandal98 (approved):

tasks/reasoning_qa/afm/tasks_json/knowledge_qa/tasks_json/task_afm_1.json
MrtinoRG, line 1: The requires key has to be always the second key in the list being the first the environment name

imandal98: Done

PR #290, opened 2026-02-17
Md corrected runs

chandan21gupta:
MD RUNS


MrtinoRG:
@chandan21gupta Do you think we can close this one?

chandan21gupta:
> @chandan21gupta Do you think we can close this one?

This contains runs in the specified structured format. You will need to merge this.

MrtinoRG:
What do you mean by runs? If it is the reports, they will go into HF, right so no need to have them here...

chandan21gupta:
> What do you mean by runs? If it is the reports, they will go into HF, right so no need to have them here...

Oh yes, in that sense this PR can be closed.

MrtinoRG (approved):

PR #291, opened 2026-02-20
feat: code sandbox

n0w0f:
Added a code execution tool sandboxed with docker.
- Would be helpful to have this as a common tool and also to run ML environment  without those optimized tools

## Summary by Sourcery

Introduce a configurable sandboxed code execution framework with Docker and subprocess backends, and expose it via Corral tools for Python code and shell command execution.

New Features:
- Add Docker-based and subprocess-based sandbox backends with persistent filesystem and optional Python state across executions.
- Provide Corral tools to execute Python code and shell commands within a chosen sandbox backend, with consistent JSON-formatted results.
- Introduce centralized configuration, resource limit, and network policy models for sandbox environments.

Enhancements:
- Refactor output capture and execution result parsing into reusable utilities shared between legacy code tools and new sandbox backends.

Tests:
- Add comprehensive tests for sandbox configuration, state persistence, output capture, execution results, Docker and subprocess backends, and the new sandbox tooling API.


kjappelbaum:
Did you consider using an external dependency for that? 

What added benefit would this bring if we make corral to default to docker use?

PR #297, opened 2026-02-22
feat: add agents that do not rely on JSON

MrtinoRG:
## Summary by Sourcery

Introduce compact YAML-based agents that avoid relying on raw JSON tool history and observations.

New Features:
- Add CompactHistoryAgent that records tool calls and results as compact YAML assistant messages instead of structured JSON tool-call/result pairs.
- Add CompactReActAgent variant that reformats ReAct tool observations and action inputs into clean YAML while preserving ReAct reasoning flow.

Enhancements:
- Export the new compact agent classes from the agents package for external use.


src/corral/agents/__init__.py
```diff
@@ -1,11 +1,15 @@
...
+from .compact_react import CompactReActAgent
 from .llm_planner import LLMPlanner
 from .react import ReActAgent
 from .reflexion_agent import ReflexionAgent
 from .tool_calling import ToolCallingAgent
 
 __all__ = [
     "BaseAgent",
+    "CompactHistoryAgent",
+    "CompactReActAgent",
```
kjappelbaum, line 12: Do you think the names should perhaps be even more similar? Right now - due to the additional `History` - it is not clear how much they are related by just looking at the name. 

But it would be really elegant if we could do that by just looking at the name

example.json
kjappelbaum, line 1: should this be committed?

src/corral/agents/compact_history.py
```diff
@@ -0,0 +1,266 @@
...
+
+from corral.agents.hooks import HookPoint
+from corral.agents.prompt_utils import create_prompt
+from corral.agents.schema import Action
+from corral.agents.tool_calling import ToolCallingAgent
+from corral.agents.utils import LiteLLMMessage, convert_to_openai_tool_format
+from corral.router.routes import CorralRouter
+
+
+class CompactHistoryAgent(ToolCallingAgent):
```
kjappelbaum: Should perhaps `YAML` be in the name, too?

src/corral/agents/compact_tool.py
```diff
@@ -0,0 +1,266 @@
...
+from corral.agents.prompt_utils import create_prompt
+from corral.agents.schema import Action
+from corral.agents.tool_calling import ToolCallingAgent
+from corral.agents.utils import LiteLLMMessage, convert_to_openai_tool_format
+from corral.router.routes import CorralRouter
+
+
+class CompactHistoryAgent(ToolCallingAgent):
+    """
+    Agent derived from ToolCallingAgent that stores past tool exchanges as clean
```
kjappelbaum, line 19: After the experiments, it would be worth adding motivation for this approach to the docstring, so users have clearer guidance on which agent to choose.

An alternative implementation would be to give agents a `serialization_format` argument accepting `json` (default), `yaml`, `toml`, etc. This might be preferable because the new classes suggest that new logic is being introduced, when that isn't necessarily the case — it's really just a variation on existing logic. From a structural standpoint, it's cleaner to keep such variations within existing implementations and reserve new classes for genuinely new logic.

src/corral/agents/compact_tool.py
```diff
@@ -0,0 +1,266 @@
...
+
+    All constructor arguments are identical to :class:`ToolCallingAgent`.
+    """
+
+    # ------------------------------------------------------------------
+    # Helpers
+    # ------------------------------------------------------------------
+
+    @staticmethod
+    def _try_parse(value: Any) -> Any:
```
kjappelbaum, line 46: this somehow very much reminds me of https://github.com/mangiucugna/json_repair 

if we would like `corral` as a framework to _build_ agents, having this json repair as an option could make sense. For benchmarking, this is less relevant.

src/corral/agents/compact_tool.py
```diff
@@ -0,0 +1,266 @@
...
+        args_parsed = cls._deep_parse(args)
+        result_parsed = cls._extract_result(result)
+
+        block = {"tool": name, "arguments": args_parsed, "result": result_parsed}
+        return yaml.dump(
+            [block],
+            default_flow_style=False,
+            allow_unicode=True,
+            sort_keys=False,
+        ).rstrip()
```
kjappelbaum, line 117: This is a bit of why I feel it should be generalized and a class argument. The actual formatting is only a small "flavor" on the same logic.

src/corral/agents/compact_tool.py
```diff
@@ -0,0 +1,266 @@
...
+                    for called_tool in tool_calls:
+                        function_name = str(called_tool.function.name)
+                        result: str | None = None
+
+                        try:
+                            raw_arguments = json.loads(called_tool.function.arguments)
+                            action = Action(
+                                tool_name=called_tool.function.name,
+                                arguments=raw_arguments,
+                            )
```
kjappelbaum, line 205: This is perhaps a case for later optimization. Perhaps `called_tool` should directly be an `Action` object? I do not see why this shouldn't be the case.

src/corral/agents/compact_tool.py
```diff
@@ -0,0 +1,266 @@
...
+
+                    final_answer_match = re.search(
+                        r"Final Answer: (.*)", content, re.DOTALL | re.IGNORECASE
+                    )
+                    self._execute_hooks(
+                        HookPoint.AFTER_ITERATION,
+                        interface,
+                        task_id,
+                        llm_response=full_llm_response,
+                    )
```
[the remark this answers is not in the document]
kjappelbaum: @MrtinoRG this seems like a good catch. What do you think?

src/corral/agents/compact_tool.py
```diff
@@ -0,0 +1,266 @@
...
+                        function_name = str(called_tool.function.name)
+                        result: str | None = None
+
+                        try:
+                            raw_arguments = json.loads(called_tool.function.arguments)
+                            action = Action(
+                                tool_name=called_tool.function.name,
+                                arguments=raw_arguments,
+                            )
+                            function_call = interface.execute_tool(
```
kjappelbaum, line 206: one question that is deeper - and might make stuff more complex - is whether we should also ask the model to create the action inputs as `yaml`. If it is shown `yaml` in the history, `json` in the docstring and  `json` as the required format, it might genuinely be confusing. 

It is clearly not easy to change all of that (including the docstrings, which might include some json) in a clean way. But it would be good to keep in mind if we see confusing results.

PR #298, opened 2026-02-23
feat: expose history

MrtinoRG:
## Summary by Sourcery

Introduce a flexible, registry-based metrics system with verbose agent history reporting and enhance CorralRunner and reporting pipelines to expose history, support custom metrics, and generate benchmark reports automatically.

New Features:
- Add support for capturing and exporting per-trial agent messages and tool call details in benchmark reports when verbose mode is enabled.
- Introduce a registry-based, pluggable metrics framework with overall and task-level core metrics, pass@k/pass^k metrics, and protocol-based context support.
- Expose metrics configuration and introspection APIs on CorralRunner, including listing, printing, registering, unregistering, and resetting metrics.
- Automatically generate and save JSON benchmark reports with informative, model-based filenames or a user-specified run name after each benchmark run.
- Allow passing conversation history into trials and prompts so agents can resume from prior context, and propagate this through run_agent and execute_single_trial.

Bug Fixes:
- Ensure reasoning-only responses with empty content still yield usable content by falling back to reasoning_content in llm_call.
- Fix exception handling in async context7_tools fetching to correctly surface ExceptionGroup sub-exceptions.
- Adjust tests and error message expectations to align with new run_agent return signature and error formatting.

Enhancements:
- Refactor BenchmarkResult to delegate all metric calculations to the new MetricRegistry, simplify reporting, and standardize display names for metrics.
- Track trial duration at the runner level instead of in the environment, and add per-trial tool execution duration as a property on TaskTrialResult.
- Improve prompt construction semantics so that history is inserted after the system prompt and initial user content is not duplicated when resuming from history.
- Enhance WandB logging to log only metrics defined in the BenchmarkResult registry and to support task-level breakdowns from task metrics.
- Sanitize model names for filesystem-safe benchmark report filenames and tweak default k-value handling for pass@k metrics.

CI:
- Update GitHub Actions workflows to use the latest actions/cache@v5 for docs and test-related caching.

Documentation:
- Add a detailed Custom Metrics Guide documenting the new metrics system, how to define and register custom metrics, configure k values, and integrate with CorralRunner and BenchmarkResult.

Tests:
- Add comprehensive tests for the new metrics core, base abstractions, registry behavior (including parallel execution), WandB logging, protocol compatibility, and reporting integration.
- Update existing agent and metrics tests to validate the new run_agent tuple return (answer, messages, usage) and the registry-based metrics outputs in reports.

## Summary by CodeRabbit

* **New Features**
  * Full metrics system with default & custom metrics, plus a registry that can compute metrics (including parallel calculation) and surface per-task and overall results.
  * Benchmark runner now persists richer reports (including optional per-trial messages) and exposes metric-listing/registration utilities.

* **Documentation**
  * New comprehensive guide on creating, registering, and using custom metrics.

* **Improvements**
  * Agents can be initialized from prior conversation traces; run flows surface message history and verbose reporting.
  * CI workflows updated to newer action versions.

* **Tests**
  * Extensive new tests covering metrics, registry, reporting, and logging.


AdrianM0 (approved):
checked the expose history commit, looks good to me

src/corral/run.py
```diff
@@ -24,17 +28,59 @@ def create_session_id() -> str:
...
+
+    Args:
+        model_name: The raw model name to sanitize
+        max_length: Maximum length for the sanitized name (default: 50)
+
+    Returns:
+        Sanitized model name safe for use in filenames
+    """
+    # Replace invalid filesystem characters with underscores
+    invalid_chars = r"[/\\:*?\"<>|\s\-]+"
```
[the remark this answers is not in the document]
AdrianM0: @MrtinoRG is this a concern?

kjappelbaum:
do you want to make this PR to `main` or to `dev`? I believe one reason we see such a big diff might be because you make a PR to `main`

src/corral/agents/prompt_utils.py
```diff
@@ -154,20 +156,20 @@ def create_prompt(
     if history is None:
         history = []
 
+    if system_prompt and not history:
+        messages.append(LiteLLMMessage(role="system", content=system_prompt))
```
kjappelbaum: This would be for the case that the `history` contains the system prompt. 

I guess this is the most common case. However, the docstring does not seem to match this implementation. You write in the docstring "these messages are inserted after the system prompt". Here, you only insert the `system_prompt` if there is no history.

src/corral/agents/prompt_utils.py
```diff
@@ -154,20 +156,20 @@ def create_prompt(
...
-
     # Add surrender instructions if enabled
     if enable_surrender and surrender_prompt:
         surrender_instructions = surrender_prompt.fill({})
         kwargs["surrender_instructions"] = surrender_instructions
 
-    user_content = build_user_content(user_prompt, task_guide=task_guide, **kwargs)
-
-    messages.append(LiteLLMMessage(role="user", content=user_content))
+    if not history:
```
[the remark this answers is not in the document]
kjappelbaum: I guess this follows a bit what I mentioned

src/corral/agents/prompt_utils.py
kjappelbaum: I am not sure if this code will still follow somewhere. But I am a big fan of "constructor classmethods". I feel they produce very clean and readable code. 

That is, in this case, I would handle the case where I would like to start the agent from a previous run with the `classmethod` `agent.from_trace`. And the `from_trace` could then also handle the details on whether the system prompt is kept or dropped. 

In the ideal case, this often also means that the `__init__` of the class does not explode with arguments, but that you handle the things that concern a specific construction method in the `classmethod` for this construction method.

src/corral/agents/utils.py
```diff
@@ -106,6 +106,13 @@ def llm_call(
 
         message = response.choices[0].message
 
+        # If message content is None, try to get reasoning_content
+        if message.content is None:
+            reasoning_content = getattr(message, "reasoning_content", None)
+            if reasoning_content is not None:
+                # Ensure reasoning_content is a string
+                message.content = str(reasoning_content) if reasoning_content else None
```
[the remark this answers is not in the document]
kjappelbaum: I agree with this assessment. 

Personally, I tend to write code without mutations because it can cause those side effects that are difficult to keep in mind sometimes. 

So, unless a method clearly communicates that it is doing something `in_place` I would work with copies. 
In most cases, this not cause meaningful performance problems for our code but improves readability and our ability to reason over code a lot

src/corral/agents/utils.py
```diff
@@ -428,6 +436,11 @@ def get_context_window(model: str) -> int:
     Returns:
         int: The max input tokens for the model, or None if unknown.
     """
+    if (
+        model
+        == "openai/1 - GPT-OSS-120b - an open model released by OpenAI in August 2025"
+    ):
+        return 131072
```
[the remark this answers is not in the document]
kjappelbaum: This suggestion, in principle, makes sense, too.

MrtinoRG:
@kjappelbaum I made to and from (I think) `main` because the branch that @AdrianM0 is working with is from `main` so it would be easier for him to merge.

About the dataclass, yeah, that is a good idea. 

I proposed the history because it is simpler

AdrianM0:
> @kjappelbaum I made to and from (I think) `main` because the branch that @AdrianM0 is working with is from `main` so it would be easier for him to merge.
> 
> 
> 
> About the dataclass, yeah, that is a good idea. 
> 
> 
> 
> I proposed the history because it is simpler 

I could also do the more manual approach you suggested, so if it makes more sense to merge into dev then probably you could go ahead

src/corral/agents/utils.py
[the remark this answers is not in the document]
MrtinoRG: No, it does not. 

Like I am aware of the hardcoded rule, but it is the only way we have to make it work for the Blablador models. 

For context this function is used to add the context aware piece in the system prompt such that the models know how much context they have consumed.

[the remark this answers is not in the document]
kjappelbaum: What I meant here is that constants are ideally not hidden in the code but configurable outside the code. It is ok for now, but bad if we extrapolate

src/corral/agents/prompt_utils.py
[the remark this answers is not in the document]
MrtinoRG: I guess the easiest solution would be to check if the first message is a `"role": "system"`

src/corral/agents/prompt_utils.py
MrtinoRG: Ok, I implement it now. 

Basically I added another bench method that run from traces, and then the classmethod you proposed in the Base Agent.

tasks/afm/src/env.py
MrtinoRG, line 1: This comes from `main` not being merged into `dev`

.github/workflows/docs.yaml
MrtinoRG, line 1: This is here because of `main` not being merged into `dev`

src/corral/agents/utils.py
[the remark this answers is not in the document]
MrtinoRG: This does not belong to this PR anymore.

Sorry for the mess with the two branches.

But perhaps the comment that I have here is that perhaps move such constants into a constants.py file or something similar

MrtinoRG:
@kjappelbaum I now implemented what you suggested, a `from_trace` class method in the base agent, that simply initializes the agent from a previous trace instead of building the prompt. 


For "exposing" this, I added the runner method `bench_from_traces` that takes the traces (pairs of task_id and trace) and simply benchmarks from there. 

A note for @AdrianM0 is that I have also added a method in the runner's `_load_traces_from_agent_logs` that will return you the traces in the desired format for the `bench_from_traces` method from a directory with the agent_logs files

MrtinoRG:
Also regarding the branchs, I merged into `dev` such that it looks much cleaner

src/corral/report/metrics/base.py
```diff
@@ -0,0 +1,220 @@
...
+        ...     def all_task_ids(self) -> set[str]:
+        ...         return set(self.tasks.keys())
+        ...
+        ...     def get_task_trials(self, task_id: str) -> TaskData | None:
+        ...         return self.tasks.get(task_id)
+        >>>
+        >>> # MyBenchmark automatically satisfies MetricContext!
+        >>> benchmark = MyBenchmark(tasks={...})
+        >>> metric = SomeTaskMetric()
+        >>> result = metric.calculate(benchmark)  # ✅ Works!
```
kjappelbaum, line 38: I guess this was not introduced in this PR, but I am not sure if the ` # ✅ Works!` comment is super helpful here

src/corral/report/metrics/base.py
kjappelbaum, line 1: In the documentation we must certainly explain those concepts. They are not necessarily intuitive

kjappelbaum: (But that is a separate issue)

src/corral/agents/base_agent.py
```diff
@@ -122,6 +123,33 @@ def __init__(
...
+    @classmethod
+    def from_trace(
+        cls,
+        trace: list[LiteLLMMessage],
+        **init_kwargs,
+    ) -> Self:
+        """Construct an agent pre-loaded with a previous conversation trace.
+
+        This classmethod creates a new agent instance whose message history is
+        initialised from ``trace``.  When the agent's ``run()`` method is
```
kjappelbaum, line 135: Perhaps I am unsure about the latest convention, but why do we have double backticks here?

src/corral/agents/base_agent.py
```diff
@@ -122,6 +123,33 @@ def __init__(
...
+        called it will use these messages instead of building a fresh prompt,
+        allowing benchmarks to be replayed from saved traces.
```
kjappelbaum, line 137: perhaps I would really add here a clarification about what happens with the system prompt

kjappelbaum: and how the history is inserted (e.g., in one concatenated message vs. one message per turn as it was serialized)

tests/agents/test_base_agent.py
kjappelbaum, line 1: it seems the tests are not adjusted to the use of the new `classmethod`

src/corral/run.py
```diff
@@ -535,6 +525,134 @@ def trial_executor(task_id: str, trial_index: int) -> TaskTrialResult:
...
+            task_ids=task_ids,
+            trial_executor=trial_executor,
+            trials_per_task=trials_per_task,
+            k_values=k_values,
+            verbose=verbose,
+            session_id=session_id,
+            tool_verbosity=tool_verbosity,
+            hooks=hooks,
+            run_name=run_name,
+            extra_wandb_config={"from_traces": True},
```
kjappelbaum, line 606: maybe it would be cool to expose those wandb runs. Then one could link the two runs or potentially even continue from the past run

src/corral/agents/llm_planner.py
```diff
@@ -106,24 +104,26 @@ def run(
...
-            system_prompt=self.system_prompt,
-            user_prompt=self.user_prompt,
-            task_guide=task_guide,
-            history=history,
-            max_iterations=self.max_iterations,
-            examples=examples,
-            tools=tools,
-        )
+        if self._initial_messages is not None:
+            self.messages = list(self._initial_messages)
```
kjappelbaum, line 112: Why can't this just happen upon initialization of the class? Then you wouldn't need the extra `_initial_messages` attribute

MrtinoRG: Because then you have the same messages for all the tasks...

src/corral/run.py
MrtinoRG: Sorry but I am not following you on this. What do you mean by expose the runs?

tests/agents/test_base_agent.py
MrtinoRG: Yeah, I did not add tests for this yet

src/corral/report/metrics/base.py
MrtinoRG: I do not know exactly what specific comments you mean, but some are explained in `docs/custom_metrics.md`. The thing is that this README is not indexed in the docs I think :(

src/corral/agents/llm_planner.py
kjappelbaum: > Because then you have the same messages for all the tasks...

But this is not necessarily bad? I believe it really comes down to how tightly we want to couple agents to a certain usage mode. You assume that you will always run multiple tasks that should be decoupled by the agent, not having any history. But we do not need to enforce this usage mode. 

We also need to rethink this also with the state discussion (it is not clear yet to me where the state should really live). 

It is probably ok for now, but I would like us to revisit this, because it feels more complex than it should.

AdrianM0:
@MrtinoRG the deepcopy fix worked, thanks

src/corral/agents/llm_planner.py
MrtinoRG: Right now they are adapted to both with this implementation (implementation that I think it is not the cleanest ever, but I do not know how to make it cleaner). 

If the user uses the agent class `from_trace(trace, **kwargs)`, there you need to provide a trace. So if one wants to run it always with the same trace (not using the runner), in this way works.

On the other hand, the way the runner uses the agent is initializing it one time and then using that instance for all the trials and tasks. For that case, the `_initial_messages` are modified before calling the run method. 

So with this the agent should not be tied to one usage mode, because then if one simply initialize the agents as normal, they will run from the beginning, or one can even use the `from_trace` class with a custom system and user prompt

MrtinoRG:
Merging this now as @AdrianM0 confirmed it worked.

@AdrianM0 if you have any other suggestions on how to improve Corral, just let us know :)

PR #302, opened 2026-02-27
Reports to hf

chandan21gupta:
Scripts to push traces and reports to HF.

## Release Notes

* **New Features**
  * Added data retrieval tools to download and cache Hugging Face datasets with support for configurable parameters and local storage.
  * Added dataset publishing tools to build and upload organized report and trace subsets to Hugging Face Hub with automatic metadata extraction and organization.
  * Introduced utilities for processing and normalizing trace and report data with validation and error handling.


scripts/pull_hf_data.py
```diff
@@ -0,0 +1,57 @@
...
+    )
+    parser.add_argument("--verbosity", required=True)
+    parser.add_argument(
+        "--type",
+        required=True,
+        choices=["traces", "overall_reports", "task_reports"],
+    )
+    parser.add_argument("--output_dir", required=True)
+
+    return parser.parse_args()
```
kjappelbaum: personally, I have a preference for building CLIs with `fire`. This makes things a bit easier to maintain

scripts/reports_hf.py
```diff
@@ -0,0 +1,243 @@
...
+
+    args = parser.parse_args()
+    path = args.path
+
+    p = Path(path)
+
+    env = p.parts[-1]
+    model = p.parts[-2]
+
+    if model == "claude":
```
kjappelbaum: This is perfect for now, but would not work anymore once we would add more claude variants

scripts/trace_hf.py
```diff
@@ -0,0 +1,352 @@
...
+AGENTS = {
+    "react": "ReActAgent",
+    "reactagent": "ReActAgent",
+    "tool calling": "ToolCallingAgent",
+    "toolcalling": "ToolCallingAgent",
+    "toolcallingagent": "ToolCallingAgent",
+}
+
+VERBOSITY = {"brief", "comprehensive", "workflow"}
+
```
kjappelbaum: Ideally, we would not repeat this code. This is in two files. (As some of the other code below)

scripts/reports_hf.py
```diff
@@ -0,0 +1,243 @@
...
+
+        for report_path in report_paths:
+            report, overall_report = get_scores(config, report_path)
+            all_reports.extend(report)
+            overall_reports.extend(overall_report)
+
+    DATASET_NAME = "jablonkagroup/corral-reports"
+    api = HfApi()
+
+    api.create_repo(
```
kjappelbaum: This means that if the repo already exists, stuff would be overridden?

scripts/reports_hf.py
```diff
@@ -0,0 +1,243 @@
...
+
+        logger.info(f"\nUploading subset: {subset_name}")
+
+        dataset = Dataset.from_pandas(subset_df.reset_index(drop=True))
+
+        dataset.push_to_hub(
+            DATASET_NAME,
+            config_name=subset_name,
+            private=False,
+        )
```
kjappelbaum: If we would write very clean code, we would also factor this out into a function to reduce repetition

scripts/reports_hf.py
```diff
@@ -0,0 +1,243 @@
...
+    DATASET_NAME = "jablonkagroup/corral-reports"
+    api = HfApi()
+
+    api.create_repo(
+        repo_id=DATASET_NAME,
+        repo_type="dataset",
+        private=False,
+        exist_ok=True,
+    )
+    trials_df = pd.DataFrame(all_reports)
```
kjappelbaum: would it make sense to add here some checks? For instance, if there are any `nan`? Or if we have lost some data on the way?

scripts/trace_hf.py
```diff
@@ -0,0 +1,352 @@
...
+        if len(trials) > len(traces):
+            logger.warning(
+                f"[WARNING] {task_name}: "
+                f"{len(trials)} trials but only {len(traces)} traces."
+            )
+            # for trial_id, trial_meta in enumerate(trials):
+            #     trace_path = None
+            #     messages = None
+            #     model_version = None
+            #     timestamp = None
```
kjappelbaum:
```suggestion
```

ideally, unused code is deleted

scripts/trace_hf.py
```diff
@@ -0,0 +1,352 @@
...
+
+        elif len(trials) < len(traces):
+            logger.warning(
+                f"[WARNING] {task_name}: "
+                f"{len(trials)} trials but {len(traces)} traces found. "
+                "Extra traces will be ignored."
```
kjappelbaum, line 243: but there is no clear algorithm with which we choose what the extra traces are that we ignore?

scripts/trace_hf.py
```diff
@@ -0,0 +1,352 @@
...
+    grouped = {}
+
+    for item in category_path.iterdir():
+        # ✅ Ignore logprobs directories
+        if item.is_dir() and item.name.lower().startswith("logprobs"):
+            continue
+
+        pair = extract_pair(item.name)
+        if pair is None:
+            continue
```
kjappelbaum, line 300: We should make it clearer if we have stuff that we could not match.

kjappelbaum:
@coderabbitai review

kjappelbaum:
Some feedback https://www.loom.com/share/2eb9d1d9d42b49198c46c251d6d46ebe

scripts/reports_hf.py
MrtinoRG, line 1: There are different keywords in the reports as the results of the PR with the metrics implementation. This is for GPT-oss the keywords in the reports are different while representing the same. Ideally the script should manage this in a way that all the configs have the same columns. This right now is not there and the columns in the HF have different names...

MrtinoRG: And the trials are not being pushed, like all the information about the different trials would get lost basically?!

MrtinoRG (changes requested):

scripts/reports_hf.py
chandan21gupta: > And the trials are not being pushed, like all the information about the different trials would get lost basically?!

Sorry I didn't understand. Did you mean trial_ids?

scripts/reports_hf.py
chandan21gupta: Have added predefined models in separate file. So if new models are to be added we can simply add model in the file. Otherwise error is thrown.

scripts/trace_hf.py
chandan21gupta: Done

scripts/reports_hf.py
chandan21gupta: If repo already exists, it would simply skip the push. 
But this can be better managed with versioning. I will change this behaviour once I start working on versioning.

scripts/reports_hf.py
chandan21gupta: Done

scripts/trace_hf.py
chandan21gupta: Removed

scripts/trace_hf.py
chandan21gupta: I am now raising errors even when number of trials != number of traces. 
I do not know how to map in case extra/less files are there, but even then I would not like to push anything in such cases. It would be better if user can simply be notified about this, as equality is a strict requirement.

chandan21gupta:
> Some feedback https://www.loom.com/share/2eb9d1d9d42b49198c46c251d6d46ebe

Thanks for the feedback. 
I have pushed the changes addressing the comments to the best of my abilities.

MrtinoRG (approved):

PR #305, opened 2026-02-28
chore: rm  older runs

n0w0f:
Removed the following files from old run

[MD](https://github.com/lamalab-org/corral/tree/dev/reports/md) 
[claude - Spectra, Catalyst, ML](https://github.com/lamalab-org/corral/tree/dev/reports/claude)
[gpt - Spectra, Catalyst, ML](https://github.com/lamalab-org/corral/tree/dev/reports/gpt4o)

Additionally remove "MD_Optimized" (https://github.com/lamalab-org/corral/tree/main/reports/corral_md_optimised)


chandan21gupta (changes requested):
There are many issues that still remain. 

I was only able to push 
gpt-4o : ml, catalyst
claude_45 : catalyst

Same problems still exist : extra traces (is OSS catalyst), missing files. There are also cases in report where trial_id contains some string ("attempt_"). Please check.

n0w0f:
As you can see, this PR is not meant to solve any of those. This is for deletinf unwanted files.

chandan21gupta:
> As you can see, this PR is not meant to solve any of those. This is for deletinf unwanted files.

Ok, but then you will have to push the reports and traces yourself with the scripts I have shared.

n0w0f:
> > As you can see, this PR is not meant to solve any of those. This is for deletinf unwanted files.
> 
> Ok, but then you will have to push the reports and traces yourself with the scripts I have shared.

I can push it, or open different PR, however you like it.
My concern here is I don't want to mix in changes from reports which would be used. This is to remove files which will not be part of analysis

chandan21gupta:
> > > As you can see, this PR is not meant to solve any of those. This is for deletinf unwanted files.
> > 
> > 
> > Ok, but then you will have to push the reports and traces yourself with the scripts I have shared.
> 
> I can push it, or open different PR, however you like it. My concern here is I don't want to mix in changes from reports which would be used. This is to remove files which will not be part of analysis

I see.
These are not the final MD runs as they were revised. But anyways, it will be helpful if you can push the runs for Catalyst, ML, and Resistors from your end so that I dont push wrong files. Thanks.

n0w0f:
> These are not the final MD runs as they were revised. 

so the optimized_md is not the correct files ? If not i would remove that as well

Are the correct files present somewhere ?

chandan21gupta:
> > These are not the final MD runs as they were revised.
> 
> so the optimized_md is not the correct files ? If not i would remove that as well
> 
> Are the correct files present somewhere ?

Present in my local. I did't push them as we had decided to push everything on HF. So they are currenly on HF. You can remove them from here.

n0w0f:
@chandan21gupta I have removed the MD optimized files as well.

Can you check if i made the right changes.
If everything looks okay, then you  can change the status from change requested.

After that i would work on fixing the issue you faced with the traces, in a different PR

n0w0f:
@imandal98 Are [these](https://github.com/lamalab-org/corral/tree/dev/reports/afm) the final version of AFM ?

chandan21gupta (approved):
Fine with me. AFM also seems to be updated one, but still would be better if Indrajeet confirms it.

PR #308, opened 2026-03-03
fix: correct that trace is not saved when agent fail

MrtinoRG:
## Summary by Sourcery

Introduce incremental Hugging Face dataset update support and ensure agent traces are persisted even on failures.

New Features:
- Add an incremental update mode to push_trace_to_hf that filters, merges, and repushes subsets of an existing Hugging Face dataset.
- Support processing an additional level_3 directory when collecting trace data.

Bug Fixes:
- Ensure agent message traces are saved even when the agent raises an exception by moving trace persistence into a finally block and recording full error context.

Enhancements:
- Change the default Hugging Face dataset name used for trace uploads and allow reusing existing configs via download.
- Improve error logging during agent execution by capturing and storing full tracebacks in both logs and messages.

Build:
- Update retrosynthesis task dependencies to include rxnmapper and pin setuptools to <80, and refresh the uv.lock file.


MrtinoRG:
Everything is already pushed, but I would recommend that we merge them such that we have a commit with all the reports and then at some point we remove all the reports

scripts/push_trace_to_hf.py
n0w0f, line 118: why environment over task_type ?
this is a nit but with environment i would expect md / catalyst etc

n0w0f (approved):

scripts/push_trace_to_hf.py
MrtinoRG: Because since I basically rerun entire envs for which some file was missing this had more sense

PR #309, opened 2026-03-05
feat: add script to pull data + snakefile

MrtinoRG:
## Summary by Sourcery

Add shared benchmarking utilities and scripts to aggregate, normalise, push, and download corral benchmark reports and QA data to/from HuggingFace, and wire them into a simple Snakemake workflow.

New Features:
- Introduce a push_reports_to_hf script to collect benchmark reports from multiple layouts, normalise metrics and task results, and push grouped model–environment configs to HuggingFace.
- Add a combine_md_reports script to merge MD sub-environment reports, recompute metrics, and write unified MD reports plus associated agent logs.
- Provide a download_reports_from_hf analysis script and Snakefile rule to fetch report configs from HuggingFace into a local JSONL dataset for analysis.
- Centralise shared constants and helper utilities for report/model/env handling, JSON I/O, and logprob entropy computation across scripts.

Enhancements:
- Refactor existing push_trace_to_hf and push_qa_to_hf scripts to reuse shared JSON and entropy utilities and shared constants for HF repositories and known models, with optional pandas handling for parquet output.


analysis/download_reports_from_hf.py
```diff
@@ -0,0 +1,258 @@
...
+
+When specific configs are requested each one is saved under its own name,
+making it easy to share or inspect a single model/environment slice:
+
+    df = pd.read_json("analysis/results/data/claude_4_5-afm.jsonl", lines=True)
+
+In both cases `Task Results` is decoded to a nested dict so it can be
+accessed without an extra `json.loads` call.
+
+Usage:
```
n0w0f, line 22: Do we have subtasks also in here or are those dealt differently ? There is no info about subtask in here

scripts/combine_md_reports.py
```diff
@@ -0,0 +1,381 @@
...
+    logger.info("")
+
+    # Assert no env directories exist in base_dir that are not covered by MD_ENVS.
+    # If a new MD sub-environment is added on disk but not to this list, its
+    # tasks would be silently excluded from every combined report.
+    actual_env_dirs = {
+        d.name
+        for d in base_dir.iterdir()
+        if d.is_dir() and not d.name.startswith("level_")
+    }
```
n0w0f, line 215: if we place the final reports at the same level, remember we cannot rerun this script as it is, as it would also consider the combined report next time

n0w0f (approved):

analysis/download_reports_from_hf.py
MrtinoRG: Yes, they are there. There is a `category` column that it is either tasks or subtasks.

scripts/combine_md_reports.py
MrtinoRG: Yeah, I think this and the hardcoded rules in the push script (hopefully) we do not have to use them again :)

PR #314, opened 2026-03-06
chore: utility shared across plots

n0w0f:
These are some utilities that i use across branches for different plots

## Summary by Sourcery

Introduce shared plotting utilities and centralise QA topic report downloading into a single combined workflow.

Enhancements:
- Refactor QA topic report downloader to discover all parquet files, infer env/model/qa_type from paths, normalise environment names, and always produce a single merged JSONL dataset.
- Add shared plotting utilities for loading, filtering, and ordering benchmark and QA data, including helpers for metric naming and environment ordering by QA scores.
- Add centralised plot configuration with display name mappings, colour palettes, and font-size settings for models, environments, and agents.


n0w0f:
@MrtinoRG I need this across different branches, so thought of keeping it in a live branch, that i can pull to different plotting branches,
We can keep updating the `plot_config` and `plot_utils` as we need. If we you have some settings finalized please add it in those, so I also use the same

analysis/download_qa_topic_reports_from_hf.py
```diff
@@ -59,6 +47,12 @@
 
 _OUTPUT_FILENAME = "qa_topic_reports.jsonl"
 
+# Environment name normalization to match reports dataset
+# QA data has some names that differ from the main reports
+ENV_NAME_NORMALIZATION = {
+    "corral_md": "md",  # QA has "corral_md", reports has "md"
+}
+
```
n0w0f, line 55: You see there is still this in cosistency in name in QA and Reports in the dataset

MrtinoRG (approved):
@n0w0f solve the CI prev to merging please

analysis/download_qa_topic_reports_from_hf.py
kjappelbaum: perhaps this should be fixed upstream in the dataset?

analysis/plot_config.py
```diff
@@ -0,0 +1,102 @@
...
+    "md": "Molecular\nDynamics",
+    "ml": "Machine\nLearning",
+    "resistor": "Resistor",
+    "retro": "Retro-\nsynthesis",
+    "spectra": "Spectra",
+    "wetlab": "Wetlab",
+}
+
+AGENT_NAMES = {
+    "react": "React",
```
kjappelbaum:
```suggestion
    "react": "ReAct",
```

analysis/plot_config.py
```diff
@@ -0,0 +1,102 @@
...
+MODEL_COLOURS = [
+    "#711c91",  # Claude 4.5
+    "#ea00d9",  # GPT-4o
+    "#7150e0",  # GPT-OSS-120B
+]
+
+AGENT_COLOURS = [
+    "#30292F",  # React
+    "#5D737E",  # Tool calling
+]
```
kjappelbaum: wouldn't  a dict be more robust than a list? 

Or have models be a dict with different properties. 
Same for agents.

analysis/plot_utils.py
```diff
@@ -0,0 +1,362 @@
...
+# Get repo root - this file is in analysis/
+REPO_ROOT = Path(__file__).resolve().parent.parent
+
+# Data paths
+REPORTS_PATH = REPO_ROOT / "analysis" / "results" / "data" / "reports.jsonl"
+QA_REPORTS_PATH = REPO_ROOT / "analysis" / "results" / "data" / "qa_topic_reports.jsonl"
+REASONING_PATH = REPO_ROOT / "analysis" / "reasoning.json"
+
+# Default per-environment level selection (used when level_strategy="default_map")
+DEFAULT_ENV_LEVEL_MAP = {
```
kjappelbaum: do we currently still use this for any figure/analysis?

analysis/plot_utils.py
```diff
@@ -0,0 +1,362 @@
...
+        agent_type_strategy: "average", "react", or "tool_calling"
+        agent_value: Specific agent value (overrides strategy)
+
+    Returns:
+        Filtered dataframe
+    """
+    if agent_value:
+        agent_type_strategy = agent_value
+
+    if agent_type_strategy == "average":
```
kjappelbaum: this is perhaps a bit counterintuitive for the API of a function. 
Why would you want to pass "average" and not simply skip the filtering?

kjappelbaum: and I guess the same question applies below

analysis/download_qa_topic_reports_from_hf.py
MrtinoRG: This should not be needed anymore as now all the configs are md (as in the reports)

analysis/plot_utils.py
n0w0f: This is not used in final version i had this while I was trying different plots

analysis/plot_config.py
n0w0f: yes, fixed now

PR #317, opened 2026-03-09
feat: add panel 5 plots

MrtinoRG:
## Summary by Sourcery

Add a new analysis pipeline to retrieve, filter, and analyse annotation data to produce marker vs scaffold impact figures.

New Features:
- Introduce Snakemake rules to fetch annotations from MongoDB, filter them, and generate marker vs scaffold plots for panel 5.
- Add a script to export annotations from MongoDB into a local JSON dump using existing environment configuration.
- Add a filtering script that deduplicates and annotates MongoDB exports with metadata derived from the accepted_questions directory structure.
- Add a plotting script that computes marker-level model vs scaffold impacts and generates PDF figures for raw count and regression-based decompositions.

Enhancements:
- Update project ignore rules to exclude new analysis outputs as needed.


analysis/filter_annotations.py
kjappelbaum, line 1: naiive question: Should we structure the analysis scripts even more? At least in folders (one folder per figure(type))?

MrtinoRG:
I will merge this as the figure in here it is mostly finalized I think

PR #323, opened 2026-03-20
Wetlab+Reports

aaaghajani:
This PR contains the latest version of the wetlab environment, as well as all the reports and agent logs for all the runs:
- models: {Claude, GPT-4o, GPT-OSS}
- levels: {1, 2, 3}
  - {task, subtask}:
    - ReAct - brief
    - ReAct - workflow
    - ReAct - comprehensive
    - ToolCalling - brief
    - ToolCalling - workflow
    - ToolCalling - comprehensive

This is the most important change to be reviewed:
https://github.com/lamalab-org/corral/commit/e4a00d5b31e8ef8d8f033f300d80b20cd0b2c3f1

the other commits are just moving files around

After merging, I will remove the "wetlab" branch.


src/corral/agents/react.py
```diff
@@ -115,6 +115,10 @@ def parse_llm_response(
                     "False", "false"
                 )
 
+                # Convert tuples to lists for color mixing in wetlab
+                if tool_name == "simulate_color_mixture":
+                    converted_input = converted_input.replace("(","[").replace(")", "]")
+
```
MrtinoRG: I am not sure if this changes should be merged. 

First I do not really understand why one need such changes. 

Second I think on the high level it is not very elegant to have such hard-coded rules

Third, the problem with the types is something we are aware (I believe there is an open issue for it), and we did not want to implement a hotfix because when we saw it it was already everything run. Changing it for one environments makes it less comparable

MrtinoRG: And I do not know how to proceed now, because some of the scores are coming from the corrected code (Wetlab) while everything else comes from a code with a known issue which is not ideal, but at least comparable

src/corral/agents/utils.py
```diff
@@ -286,6 +286,17 @@ def convert_dict_arg(arg: dict) -> dict:
         prop["description"] += (
             ' - Provide as an array of strings, e.g., ["item1", "item2"]'
         )
+    elif arg_type == ("list[tuple[str, float]]"):
+        prop["type"] = "array"
+        prop["items"] = {
```
MrtinoRG, line 291: And the same applies here I guess

MrtinoRG: @kjappelbaum do you have an opinion on this?

tasks/wetlab/wetlab/compositions.pkl
aaaghajani, line 1: This is an artifact of saving the inventory between subtasks. It can be removed

tasks/wetlab/pyproject.toml
aaaghajani, line 1: Why is reaktoro removed?

MrtinoRG: Because it goes into the conda env (check environment.yml)

MrtinoRG: Reaktoro can only be installed with conda, not with pip

src/corral/agents/react.py
kjappelbaum: and with "corrected code" you refer to the code that is here? 
At least for the release with the paper, I would have one consistent release of the corral package which corresponds to the version we used to run the experiments. Otherwise, we have 0 chance of reproducing anything

kjappelbaum:
Screenshot 2026-03-26 at 12 56 30
also: why does GitHub not show any change compared to `dev`?

kjappelbaum:
so what is the actual change that is then still supposed to be merged into dev?

MrtinoRG:
> Screenshot 2026-03-26 at 12 56 30 also: why does GitHub not show any change compared to `dev`?

This is a bug or what it is now shown when there are too many changes. If you try to see the files you will face an github status error

MrtinoRG:
> so what is the actual change that is then still supposed to be merged into dev?

The changes in the commit above 
[fix: correct JSON schema for simulate_color_mixture in wetlab](https://github.com/lamalab-org/corral/pull/323/commits/e4a00d5b31e8ef8d8f033f300d80b20cd0b2c3f1)
[e4a00d5](https://github.com/lamalab-org/corral/pull/323/commits/e4a00d5b31e8ef8d8f033f300d80b20cd0b2c3f1)

src/corral/agents/react.py
aaaghajani: That was the reason I included them in the PR, so the runs can be reproduced the way I ran them.

MrtinoRG: Yeah, but right now we will have this to reproduce wetlab but breaks everything else

MrtinoRG: I mean, this code is for the general react implementation meaning that it will be applied to all the environments, despite the correction in our runs was applied only when running wetlab. 

That's why I am asking

kjappelbaum: puh, I did not see that it was for this general file. This is annoying. Can't we have be same effect with a hook? Or some sanitization that the env does?

src/corral/agents/utils.py
kjappelbaum: this is a very special case here? Was this combination of types used anywhere else? What broke without this change?

src/corral/agents/react.py
aaaghajani: > breaks everything else

How? It only applies to `simulate_color_mixture`, or if tool argument has type `List[Tuple[str, float]]`. So this will not affect any other environments.

kjappelbaum: because it is a general code based that is supposed to be reused by others. 

(1) We should not encourage others to fill general code with special cases
(2) We do not know if someone will have a tool like that in the future

src/corral/agents/utils.py
aaaghajani: This specific combination only appears in `simulate_color_mixture` in wetlab and nowhere else. Without this, the agent writes the whole argument as a string.

aaaghajani: The issue is that any argument that is not found in the `TYPE_MAPPING` dictionary (which only contains primitives), defaults to string. There was one exception and that's `list[str]`, so I followed the same logic and added this exception.

MrtinoRG: Since it is only in WetLab, I am merging it. 



Then, we should merge dev into main and start thinking on doing a release/Zenodo. On the meanwhile I will coordinate with @n0w0f on how he thinks it is the best way to solve the function that converts to OpenAI tool format. I still believe that the best perhaps would be to use the decorator that they have.

PR #324, opened 2026-03-20
feat: landing page for Corral benchmark

n0w0f:
## Quick start

```bash
uv run python site/extract_data.py   # regenerate data from source
cd site && uv run python -m http.server 8090 # serve at http://localhost:8090
```

## Summary

Interactive landing page for the Corral benchmark at `site/index.html`. Showcases all environments, tools, tasks, and subtasks with a verbosity slider for tool descriptions.

### Features
- **Auto-discovered environments**: `extract_data.py` walks `tasks/*/environments/` and finds `tools.py`/`score.py` via glob — no hardcoded paths. Display names and descriptions in `site/env_meta.json`
- **Environment explorer**: Browse environments with Tools, Tasks, and Subtasks tabs
- **Tool verbosity slider**: 7 levels (Brief → Comprehensive) showing progressively more detail including Arguments, Returns, Raises, Limitations, and Examples as styled sections
- **Level selector**: Filter tasks/subtasks by level for multi-level environments (e.g. Retrosynthesis has 3 levels, AFM has 4)
- **Task/subtask capping**: Shows up to 25 chips with a "more" link to the [HuggingFace dataset](https://huggingface.co/datasets/jablonkagroup/corral-environment-tasks)
- **Dynamic stats**: Environment count, tool count, task count, and subtask count all computed from data at runtime
- **Scoring code viewer**: Each task/subtask shows its scoring function source code
- **Results & Leaderboard tabs**: Placeholder structure for benchmark results

### Environment names (matching paper)
| Directory | Display Name |
|---|---|
| `afm` | AFM Operation |
| `catalyst` | Surface Construction |
| `corral_md` | Molecular Simulation |
| `ml` | ML Property Prediction |
| `resistor_network` | Circuit Inference |
| `retrosynthesis` | Retrosynthetic Planning |
| `spectra_elucidation` | Spectra Elucidation |

### Adding a new environment

1. Create `tasks/<env>/environments/level_1/{tasks_json,subtasks_json}/` with standardized JSON files
2. Add a `tools.py` and `score.py` somewhere under `tasks/<env>/`
3. Optionally add an entry in `site/env_meta.json` for a custom display name and description
4. Run `uv run python site/extract_data.py` — the new environment auto-appears

### Files
| File | Purpose |
|---|---|
| `site/index.html` | Single-page app (Alpine.js + Tailwind) |
| `site/extract_data.py` | Extracts tools, tasks, subtasks, scoring fns from source |
[...]

PR #328, opened 2026-03-23
refactor: standardize task configs across all environments

n0w0f:
## Summary

Standardize all environment task configurations  to a common directory structure and JSON schema, upload to HuggingFace, and add a shared task loader utility.

### Directory standardization
- All environments now follow: `tasks/<env>/environments/level_<N>/{tasks_json,subtasks_json}/<task>.json`
- Migrated: catalyst, ml, resistor_network, corral_md, afm, retrosynthesis, wetlab, samplemath (spectra_elucidation was already compliant)
- Legacy paths (`config/`, `src/enviroment/`) removed.

### JSON schema standardization
- All task/subtask files converted to **array format** (top-level `[]`)
- Dict-keyed files: keys moved to `id` field on each entry
- Unified field names: `scoring_fn` → `scoring_function`, added `description` from `input.prompt` where missing
- Every entry now has a `uuid`

### Pre-commit enforcement
- `check-env-convention` hook (`scripts/check_env_convention.py`)
- Validates directory structure, required subdirs, JSON validity

### HuggingFace upload
- Script: `scripts/upload_tasks_to_hf.py` (dry-run by default, `--push` to upload)
- Repo: [`jablonkagroup/corral-environment-tasks`](https://huggingface.co/datasets/jablonkagroup/corral-environment-tasks)
- **34 subsets**, **909 entries** total across all environments.

### Shared task loader utility
- New: `src/corral/utils/task_loader.py`
  - `load_task_entries(environment, level, task_type)` — load from HuggingFace
  - `load_task_entries(local_dir=...)` — load from local JSON directory
  - `load_task_entries_from_env_package(Path(__file__).parent, level, subtask)` — resolve standard path relative to env package
- Refactored catalyst `env.py` as **reference implementation** using the shared loader
- Other environments to be migrated in follow-up (see #329)

## Scripts

| Script | Purpose |
|---|---|
| `scripts/upload_tasks_to_hf.py` | Upload tasks to HF (`--push` flag) |
| `scripts/check_env_convention.py` | Pre-commit hook for directory convention |

## Test plan

- [x] `scripts/check_env_convention.py` passes on all environments (including samplemath and wetlab)
- [x] `scripts/upload_tasks_to_hf.py` dry-run shows 34 subsets / 909 entries
- [x] HF dataset uploaded with all subsets
- [x] `task_loader` imports and loads catalyst tasks from local directory
- [x] Pre-commit hook catches invalid structures


n0w0f:
This PR would help in #324 
Note subtasks in MD level -2 seems to be missing !

MrtinoRG (commented):
At least in Retro and Spectra, you will need to change the paths in the environment.py file, as to run I define a level and subtask as args but not a path

kjappelbaum:
This is up for discussion: But would there be value in such refactoring cases to first commit the scripts and then commit the changes? This might make reviewing easier. I have no clear opinion here, and am happy to hear what you think

scripts/check_env_convention.py
```diff
@@ -0,0 +1,149 @@
+"""Pre-commit hook to validate that environment task configs follow the standard convention.
+
+Expected structure:
+  tasks/<env>/environments/level_<N>/tasks_json/<task>.json
+  tasks/<env>/environments/level_<N>/subtasks_json/<task>.json
+
+Rules:
+  1. Each environment must have an `environments/` directory with `level_<N>/` subdirs
+  2. Each level must contain `tasks_json/` (required) and optionally `subtasks_json/`
```
kjappelbaum, line 9: so tasks will not always go via JSON and we always need to have levels? 
Are there cases where this can cause problems? For example, when we want to use an env to solve a task ("corral is a framework for scientific agents") and not to benchmark an agent

scripts/check_env_convention.py
```diff
@@ -0,0 +1,149 @@
...
+import json
+import re
+import sys
+from pathlib import Path
+
+TASKS_ROOT = Path("tasks")
+
+# Environments that should follow the convention
+# (skip samplemath as it's a sample/demo environment)
+SKIP_ENVS = {"samplemath"}
```
kjappelbaum: why should the demo env not follow this convention? Perhaps there it is even most important to follow the best practices of the framework

scripts/check_env_convention.py
kjappelbaum: and this might be a long-term discussion: But I could also see a case where we would like to remove the tasks from the repo and move them to a dataset on HF

scripts/check_env_convention.py
```diff
@@ -0,0 +1,149 @@
...
+        if not env_dir.is_dir():
+            continue
+        if env_dir.name in SKIP_ENVS:
+            continue
+        # Only check envs that have an environments/ dir or config/ dir (legacy)
+        has_environments = (env_dir / "environments").is_dir()
+        has_legacy_config = (env_dir / "config").is_dir()
+        has_legacy_enviroment = (env_dir / "src" / "enviroment").is_dir()
+
+        if has_legacy_config:
```
kjappelbaum: do you want to deprecate this check at some point? what value will it have once everything is migrated?

scripts/standardize_task_schemas.py
kjappelbaum, line 1: do we want to keep this one-off script in the repository? Once the migration is done, this should no longer be needed, right?

scripts/standardize_task_schemas.py
```diff
@@ -0,0 +1,130 @@
...
+def standardize_group_b(data: dict | list) -> list[dict]:
+    """Standardize Group B (catalyst, ml, resistor, corral_md, afm) entries.
+
+    Converts dict-based format to array-based.
+    """
+    if isinstance(data, list):
+        # Already array format
+        entries = data
+    else:
+        # Dict format: {"task_key": {task_data}, ...} → [{...with id}, ...]
```
kjappelbaum: this is a helpful comment

scripts/upload_tasks_to_hf.py
kjappelbaum, line 1: if we already push stuff to HF, it might be even less clear why we then enforce the presence of the `json` files.

src/corral/utils/task_loader.py
```diff
@@ -0,0 +1,131 @@
...
+    # From local directory
+    entries = load_task_entries(local_dir="tasks/catalyst/environments/level_1/tasks_json")
+"""
+
+import json
+from pathlib import Path
+
+from loguru import logger
+
+HF_REPO = "jablonkagroup/corral-environment-tasks"
```
kjappelbaum, line 25: this is a longer-term discussion: but do we really want to keep them all in one repo on HF? how do we want to handle external contributions? WIth @MrtinoRG 's plugin system, this anyhow will look differently

scripts/standardize_task_schemas.py
n0w0f: This pruning and documentation about this standard could be a separate PR ?

src/corral/utils/task_loader.py
n0w0f: One option would be that along with tasks we also push an artifact that would help in registering the environment in plugin system. So only environment with the tasks and the artifacts in  HF would be auto-discovered registered.  Versioning is also easy this way.

We probably need to setup a  discussion covering all the points you raised.

1. multiple datasets (a collection) vs single dataset

2. How easy would be someone from outside to add an environment if we integrate HF a lot ? Or would this happen via PR ? 

3. How would someone develop and test and environment without adding to GH or to plugin. I feel it would still help to keep the possibility of loading tasks from local open. 


>  so tasks will not always go via JSON and we always need to have levels?
Are there cases where this can cause problems? For example, when we want to use an env to solve a task ("corral is a framework for scientific agents") and not to benchmark an agent

Some of those structure was enforced so that we could easily auto discover levels etc for the landing page. One option is we add level as a key to tasks json and keeping it flat. There would be clever ways, but this probably need discussion and iteration

n0w0f:
> At least in Retro and Spectra, you will need to change the paths in the environment.py file, as to run I define a level and subtask as args but not a path

I sort of intentionally did not do it at this point.  I have made an issue with some cotext #329 
Once the protocol is finalized I can make a different PR where i adapt all the env to work with that

scripts/check_env_convention.py
n0w0f: some of those scripts are still here because wetlab was not merged at this point.

kjappelbaum (approved):

PR #333, opened 2026-03-27
feat: possible solution for the tool types

MrtinoRG:
## Summary by Sourcery

Align tool definitions and usage with OpenAI function-calling and Agents SDK, simplifying tool metadata handling and removing legacy conversion utilities.

Enhancements:
- Refactor the Tool base class to store parameter metadata as JSON Schema, deriving ToolArgument views for backwards compatibility and providing OpenAI and MCP-compatible representations.
- Update the @tool decorator to use the OpenAI Agents SDK function_tool for schema generation, rely on type hints and pydantic.Field metadata, and support hidden arguments via schema filtering and default injection.
- Change the environment and server APIs to expose tools directly in OpenAI function-calling format, with verbosity-based description filtering applied to JSON Schema properties.
- Adjust tests and fixtures to the new Field- and Literal-based tool definitions and the new tool exposure format, removing legacy docstring-parsing and type-mapping tests.

Tests:
- Update tool tests to validate Field-based metadata handling, Literal choices, default injection, and JSON Schema-driven argument validation.
- Update agent and interface tests to expect tools already in OpenAI function-calling format from the backend and to drop reliance on the removed conversion helpers.

Chores:
- Remove legacy type-mapping, docstring-parsing, and tool-conversion utilities and their associated tests that are no longer needed with the new schema-based tooling pipeline.
- Relax Ruff linting for tests to allow use of pydantic.Field (B008).


MrtinoRG:
image


I would say it works for the Tool Calling agent (see image above). Let's see ReAct

MrtinoRG:
I like what I see: 

image

src/corral/backend/env.py
```diff
@@ -142,25 +142,9 @@ def add_tool(self, tool: Tool):
...
-                        "choices": arg.choices,
-                    }
-                    for arg in t.arguments
-                ],
-            }
-            for t in self.tools.values()
-        ]
+    def get_available_tools(self) -> list[dict[str, Any]]:
+        """Return tools in OpenAI function-calling format."""
+        return [t.get_openai_tool_format() for t in self.tools.values()]
```
kjappelbaum, line 147: it feels that there should probably be a test for this method that is added with this PR?

pyproject.toml
```diff
@@ -56,6 +56,7 @@ dependencies = [
   "datasets>=4.6.1",
   "snakemake[irt]>=7.32.4",
   "netcdf4>=1.7.3",
+  "openai-agents",
```
kjappelbaum: do we need to pin a certain version?

tasks/wetlab/environment.yml
```diff
@@ -5,6 +5,7 @@ channels:
 dependencies:
   - python>=3.10
   - reaktoro
+  - matplotlib
```
kjappelbaum, line 8: is this a required change?

tasks/retrosynthesis/retrosynthesis/env.py
```diff
@@ -64,7 +64,7 @@ def load_tasks_from_json(
                 name=task["name"],
                 description=task["input"]["prompt"],
                 tools=task.get("tools", []),
-                scoring_fn=SCORING_FUNCTIONS[str(task["scoring_fn"])],
+                scoring_fn=SCORING_FUNCTIONS[str(task["scoring_function"])],
```
kjappelbaum, line 67: would an invalid key fail silently here? I guess not?

kjappelbaum: I am asking this to see if this should be verified for other environments, too

MrtinoRG: No, that's why I found the error...

tasks/wetlab/environment.yml
MrtinoRG: Without it prints a warning when launching the environment

tasks/ml/src/ml/tools.py
```diff
@@ -1854,7 +1857,11 @@ def train_xgboost_model(
         }
 
         if hyperparameters:
-            default_params.update(hyperparameters)
+            default_params.update(
```
kjappelbaum, line 1860: how did you now spot/make those changes? 

Is this because those changes already have been done somewhere or because of you running tests?

tasks/retrosynthesis/retrosynthesis/env.py
MrtinoRG: And I think I review all the envs now

tasks/ml/src/ml/tools.py
MrtinoRG: Most of the ones regarding the envs were spotted when running the envs. 

So I was thinking in adding some tests that simply run the env and check if the server is correctly setup

tasks/ml/src/ml/tools.py
```diff
@@ -849,7 +849,7 @@ def select_polymorphs_with_strategy(
 
 @tool(hidden_args=["work_dir"])
 def consolidate_polymorph_datasets(
-    composition_files: dict[str, str],
+    composition_files: str,
```
kjappelbaum, line 852: it is a bit scray to see those things. How did you spot them?

tasks/ml/src/ml/tools.py
MrtinoRG: But also note that this one is a result of the current tool implementation.

And the ones about the keys are a results of #328 I think

tasks/ml/src/ml/env.py
```diff
@@ -64,42 +64,53 @@ def get_scoring_function(name: str, params: dict | None = None) -> Callable:
 def load_tasks_from_json(
     json_path: str | Path, work_dir: str | Path
 ) -> dict[str, TaskDefinition]:
-    """Load task definitions from a JSON file.
+    """Load task definitions from a directory of JSON files.
```
kjappelbaum: it is either `dir` or `file`?

src/corral/backend/tool.py
```diff
@@ -1,59 +1,126 @@
...
+    generated by the OpenAI Agents SDK `function_tool` and stored directly.
     """
 
     def __init__(
         self,
         name: str,
         description: str,
-        arguments: list[ToolArgument],
+        arguments: list[ToolArgument] | None = None,
+        params_json_schema: dict[str, Any] | None = None,
```
kjappelbaum: perhaps this would really benefit from a note on what the preferred way of using it would be. I wonder if this also impacts other parts of the documentation.

kjappelbaum: and does this mean that you would deprecate one route for deriving the tool metadata?

src/corral/backend/tool.py
```diff
@@ -1,59 +1,126 @@
...
                 try:
-                    # Basic type checking
-                    if arg.type == "int":
+                    if arg.type == "integer":
                         int(value)
-                    elif arg.type == "float":
+                    elif arg.type == "number":
                         float(value)
-                    elif arg.type == "bool":
+                    elif arg.type == "boolean":
```
kjappelbaum: sorry for being so ignorant: But why do we need this change now? Where in the code has the type annotation format been changed?

tasks/wetlab/environment.yml
kjappelbaum: a warning or an error? I.e. do we really need matplotlib?

tasks/retrosynthesis/retrosynthesis/env.py
kjappelbaum: should we have an integration test that at least tries to initalize all environments? I realize that AFM might only work on windows but the intialization might still work (?)

PR #357, opened 2026-05-18
feat: add blog draft

MrtinoRG:
## Summary by Sourcery

Add a new documentation blog entry describing scientific-agent benchmarking and reasoning failures, along with associated trace artifacts.

Documentation:
- Add a long-form blog post explaining the Corral scientific-agent benchmark, its environments, and observed epistemic failure modes in LLM agents.

Chores:
- Check in representative JSON trace files used as case studies for the new blog post.

## Summary by CodeRabbit

* **Documentation**
  * Added blog post exploring how AI agents approach scientific reasoning, examining inquiry patterns and evidence interpretation through case studies.

* **New Features**
  * Implemented shareable deep links for blog trace examples.
  * Trace explorer now auto-selects and displays traces based on URL parameters for easy sharing and bookmarking.

[Review Change Stack](https://app.coderabbit.ai/change-stack/lamalab-org/corral/pull/357?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


docs/blog_18_05/index.md
```diff
@@ -0,0 +1,423 @@
+Scientific agents are promising because they seem to move the hard part of science into a loop we can automate.
+
+Give the model a question. Give it tools. Let it run experiments, inspect outputs, revise its beliefs, and submit an answer. This is a beautiful idea. It is also very easy to fool yourself with, because the final answer is the easiest thing to look at and the least informative thing about what happened.
```
kjappelbaum: i like the tone :)

docs/blog_18_05/index.md
```diff
@@ -0,0 +1,471 @@
...
+
+Our project did not start as an epistemology project. It started as a more ordinary benchmark-building effort: create a set of scientific tasks, run agents on them, and measure how well they perform. The traces were supposed to be mostly infrastructure. They were there so we could debug the environments, inspect failures, and make sure the evaluation was doing what we thought it was doing.
+
+Then we started going through them.
+
+Some failures looked normal. The agent made a bad decision. It chose the wrong formula. It misunderstood a tool. But other failures had a different direction. The agent would gather evidence and then ignore it. It would notice a contradiction and then explain it away. It would never update its initial belief, but instead force the evidence to fit some weak validation. It would test one hypothesis, reject it, invent a new one, and submit the new one without testing. At that point the question became less like “how accurate are these agents?” and more like “what kind of reasoning process is this?”
```
kjappelbaum: Perhaps there is almost too much stylistic elements in here.

The reason I wondered about it is because "wrong formula" can be "not using evidence". 

I perhaps would have phrased that by going through the logs patterns emerged how the systems sometimes appear to rather sample, or to make up their mind before. And then later we added the formalization labels such as "evidence", "belief", ...

MrtinoRG: Note that I do not enter in the Reasoning Patterns or Breakdowns in the entire blog. "Belief" and "evidence" are just words not referencing the patterns

docs/blog_18_05/index.md
```diff
@@ -0,0 +1,471 @@
...
+
+That is how a benchmark project became, partly, an epistemology project: to show the concerning behaviors we observed in the traces.
+
+A lot of scientific work is not just the final answer. It is the process that makes the answer trustworthy. A hypothesis is proposed, tested, and compared with evidence. Evidence against it should count. When observations conflict with a belief, the belief may need to change. This is close to Popper’s view of science as conjectures exposed to possible refutation [[1]](#ref-1). It also needs a caveat: scientists do not usually reject a whole theory after one anomaly [[2]](#ref-2). But if a theory survives only by adding ad hoc explanations after each failure, it becomes less progressive [[3]](#ref-3).
```
kjappelbaum: I do not want to bias this too much. But to me, the Gettier problem/justified true belief are the deeper reasons why process matters.

docs/blog_18_05/index.md
```diff
@@ -0,0 +1,471 @@
...
+
+Without this process, an output can look scientific while missing an important part of science. Tests matter because they give a claim a chance to fail. In Mayo’s severe-testing view, evidence supports a claim only when the test could have revealed problems with that claim [[4]](#ref-4). Imagine Relativity without Mercury’s perihelion, light bending, or time dilation, or the double helix without X-ray diffraction. A theory without tests is not necessarily wrong, but it is not trustworthy.
```
kjappelbaum: I think people might now be allergic to patterns like "not X, buy Y", which are seen as patterns of LLM writing

docs/blog_18_05/index.md
```diff
@@ -0,0 +1,471 @@
...
+
+This matters for agents because they can reproduce the visible steps of inquiry without following the discipline behind them. They can propose hypotheses, call tools, and write explanations while still failing to let observations constrain the answer. Peirce described inquiry as a process that moves from doubt toward settled belief through its consequences [[5]](#ref-5). Merton emphasized that science also relies on organized skepticism and public scrutiny [[6]](#ref-6). Agent traces are useful because they let us inspect whether the process behind an answer had these features.
```
kjappelbaum: I'd consider separating the background of philosophy of science from the observations. 

I'd consider making a series where the frist post shows some of the observations and then we later dive into the implications according to epistimology.

docs/blog_18_05/index.md
```diff
@@ -0,0 +1,471 @@
...
+
+The headline is simple. Present-day LLM agents can produce outputs that look like scientific progress, while the process that produced them often does not look like scientific reasoning.
+
+However, most of the work was not the epistemology itself. It was building the environments and all the engineering needed to make these agents run, fail, and leave interpretable traces. So the rest of this post has three parts. First, we describe the practical effort behind the benchmark: the tasks, tools, agents, traces, and evaluation machinery. Then we summarize the main performance results. Finally, we turn to the behaviors hidden behind the aggregate numbers. Those behaviors are what made us realize that the benchmark was not only measuring performance, but also exposing the concerning ways in which these systems produce those results.
```
kjappelbaum: This paragraph could also say: This is the first post in a series of posts where the first will show the observations, then the practical efforts, and then dive into the epistimological implications.

docs/blog_18_05/index.md
```diff
@@ -0,0 +1,471 @@
...
+
+## **Learning from evaluation design**
+
+We did not start this project without experience in evaluation. As a team, we had already worked on scientific QA benchmarks such as ChemBench [[8]](#ref-8) and MaCBench [[9]](#ref-9). In those projects, much of the work was mostly to design good questions. That experience helped us, but evaluating agents turned out to be a different kind of problem.
```
kjappelbaum: But even though you say here that most the work was to design good questions, there was also tons of work on how to even pose them to models in comparable ways :)

docs/blog_18_05/index.md
```diff
@@ -0,0 +1,471 @@
...
+
+![Corral framework](figures/figure_corral_framewor.png)
+
+This became important very quickly. In the Spectroscopy and Inorganic Analysis environments, for example, the system needs to know the identity of the unknown sample, but the agent must not see it. Tool calls therefore have an internal side that is hidden from the agent. The environment can use the sample ID, the ground-truth molecule, the reagent state, or simulated instrument settings. The agent only receives the kind of observation that a scientist would get from an experiment.
```
kjappelbaum: this is a nice paragraph and very cool info that is not so clear in the paper

docs/blog_18_05/index.md
```diff
@@ -0,0 +1,471 @@
...
+
+For the molecular-dynamics tasks, this was especially challenging because the environment had to make remote LAMMPS runs behave like ordinary tool calls. We used [Modal](https://modal.com/) to host a custom LAMMPS image, persistent volumes for potentials, structures, and simulation outputs, and remote helpers for log parsing and scoring. Most of the engineering work was about state: ensuring that each tool call saw the right files, each trial wrote to an isolated workspace, failed simulations returned useful errors, and outputs persisted long enough for later tools and scorers without leaking into other runs.
```
kjappelbaum: I think solely on this paragraph it is not clear what makes LAMPPS distinct from the other environments. 
Also, for the general reader it might not be clear what LAMMPS and Modal are

docs/blog_18_05/index.md
```diff
@@ -0,0 +1,471 @@
...
+
+We decided to build the benchmark engine from scratch, keeping a strict separation between environments and agents inspired by Aviary [[10]](#ref-10). During the evaluation runs, environments are accessed through API endpoints: they own task state, hidden variables, tools, reset behavior, and scoring. Agents, in contrast, receive task prompts and tool observations, and then choose tool calls or final answers. This separation follows a more Markovian evaluation, and clearly allows us to ablate the impact of the different components of these systems. It also gives us control over which information is visible to the model and which information remains hidden.
```
kjappelbaum: somehow the writing style here (and below) is different from the one above that discusses the epistimological background. I like the one here more.

docs/blog_18_05/index.md
```diff
@@ -0,0 +1,471 @@
...
+
+The main lesson was that these details were not separate from evaluation design. Hidden state, reset behavior, remote execution, logging, parsing, tool schemas, error handling, trace recording, and context limits all affect what is being measured. To make the benchmark fair, stable, reusable, and interpretable, they had to be part of the framework itself. In the end, we built a system where new tasks, tools, agents, and evaluation metrics can be added without changing the core engine. This also made it possible to record full traces of the agents’ actions, which later became central to understanding how the agents reasoned.
```
kjappelbaum: I feel one big issue also had been to somewhat approximate comparability of environments in docstrings and co

kjappelbaum: and another issue had been checkpointing/restart as stuff fails all the time - this is perhaps relevant in the context of your state discussion

docs/blog_18_05/index.md
```diff
@@ -0,0 +1,471 @@
...
+
+## **What the results show**
+
+Across more than 25,000 agent runs, the broad result is that current scientific agents can often produce useful outputs, but their performance depends strongly on the kind of scientific work they are asked to do. In relatively procedural workflow tasks, such as Molecular Simulation and Adsorption Surface Construction, the best configurations approached ceiling performance. In broader tasks that required hypothesis generation, testing, and revision, performance dropped sharply; in the hardest spectroscopy and inorganic analysis settings, even the strongest configuration stayed below 60%.
+
+The strongest predictor of success was the base model, not the agent wrapper. We compared GPT-4o, Claude Sonnet 4.5, and GPT-OSS-120B under both ReAct and structured tool-calling scaffolds. Through IRT analysis of diagnostic question-answer items, followed by a latent-factor model of benchmark success, we found that model reasoning ability accounted for 41.4% of the explained variance. The interaction between environment and task scope explained 30.1%. By comparison, the scaffold explained only 1.5%, and tool-description verbosity explained only 0.1%.
```
kjappelbaum: not sure if you would like to mention this. But I feel this is also what we feel in practice when we use coding agents: Changing the model is way more relevant than changing the scaffolding

docs/blog_18_05/index.md
```diff
@@ -0,0 +1,471 @@
...
+```
+
+The final answer it submits is `CC1=CC=C(C=C1)CCBr`.
+
+The correct answer is 4-bromophenethylamine, `NCCc1ccc(Br)cc1`. Its formula is C₈H₁₀BrN. Its exact mass is 198.9997, which is the peak the agent was trying to explain. Its NMR is the spectrum the agent already had on screen: the two close-together CH₂ triplets are the classic phenethylamine pattern, and the missing protons in the count are the NH₂ protons, which exchange and are routinely absent from a brief summary. Its IR has primary-amine stretches right where they appeared.
+
+The mistake is not arithmetic. The agent's calculations are mostly right. The mistake is structural. A hypothesis formed early - *this molecule contains only C, H, and Br* - and stayed in place after three different observations should have refuted it. The IR evidence was relabelled as artifact. The mass evidence was reframed as an adduct. The fragment-tool evidence was simply not engaged with. The final move, invoking [M+H]⁺ to absorb a one-mass-unit gap, is a small piece of chemistry plausibility recruited to defend a belief the evidence had already broken.
```
kjappelbaum: I would drop the rhethorical "the mistake is structural". 

Up to here the walkthrough through the failure case was super nice

docs/blog_18_05/index.md
```diff
@@ -0,0 +1,471 @@
...
+
+This is one realisation of the 71% number.
+
+### **Sampling new candidates as a method**
+
+![Sampling candidates as a method](figures/image_4a.png)
+
+A second pattern is less about refusing one particular refutation and more about replacing scientific inference with search. The agent has a space of possible answers. It samples one, asks a validation tool whether it looks good, samples another, asks again, and keeps going. This can look superficially empirical because experiments are being run. But the role of the experiment has changed. Instead of testing an informed hypothesis derived from the evidence, the tool becomes a scoring function for guesses.
+
+This [resistor trace](traces/task_3_20251111_112807.json) makes this especially visible. The task gives the agent ten exact node-to-node resistance measurements between five nodes: A, B, N_b1, X1, and X2. The correct scientific move is to infer constraints on the topology from the full resistance matrix. The environment also provides `validate_measurements`, which can compare a proposed topology against all measurements.
```
kjappelbaum: Perhaps "resistor trace" could be explained with something like "an environment where the agent needs to infer a resistor network"

docs/blog_18_05/index.md
```diff
@@ -0,0 +1,471 @@
...
+
+What it is more concerning is that in the Scope 1 of this environment, the agent is given with the `return_possible_fragments` tool, which returns fragments of the actual sample. This tool can provide direct evidence about the presence of certain atoms or groups in the molecule. We observed that in some traces the agent would receive fragments that clearly indicate the presence of some heteroatom not considered in the initial formula, but it would ignore that evidence and stick to the original belief. The agent would then try to rationalize the mass spectrum as an adduct of a different formula, rather than updating the formula to include the heteroatom.
+
+The [example trace](traces/22_22222_orgsyn_222_2222_20251217_194546.json) illustrates this behavior. It corresponds to the model Claude Sonnet 4.5 with a ReAct scaffold. The objective is the molecule shown in the image.
```
kjappelbaum: I guess we cannot directly link to a trace in the `explainers` section?

docs/blog_18_05/index.md
kjappelbaum: I feel here one example of the "correct scientific move" would help. Perhaps @n0w0f could also provide some more context here.

docs/blog_18_05/index.md
```diff
@@ -0,0 +1,471 @@
...
+
+Across the run, `validate_measurements` is called **eighteen** times. The trace becomes a sequence of topologies and resistor values, each adjusted in response to the aggregate error. Some changes improve one measured pair and worsen another. The agent notices local improvements:
+
+```plaintext
+Better!
+Good progress!
+Excellent! Getting very close.
+```
+
+But the reasoning never consolidates into a defensible circuit model. It is hill-climbing in topology space. The run ends with:
```
kjappelbaum: perhaps here it would good to contrast it with how good efficient scientific reasoning would have looked like

docs/blog_18_05/index.md
```diff
@@ -0,0 +1,471 @@
...
+Close but not quite matching. Let me try another structure...
+```
+
+Then another class:
+
+```plaintext
+Let me try an aziridine with an allyl substituent:
+```
+
+Then an N-methyl version, an oxaziridine, an allyl oxaziridine, an N-methyl aziridine, an oxazolidine, an isoxazolidine, and so on. After five setup calls to collect or derive evidence, every remaining tool call in the run is another validation of a sampled candidate: `simulate_spectra` is called **seventeen** times. That is the magnitude of the problem. The loop is not a short sanity check at the end of an inference; it is the dominant computation in the trace. Many of the candidates have the wrong carbon count or reproduce one feature while breaking another. The agent sometimes notices this:
```
kjappelbaum: here, instead of "dominant computation" it would be stronger to quantify the percentage

docs/blog_18_05/index.md
kjappelbaum: and I feel that there might also be value in highlighting in the other hidden work: computing reproducible ground truths in MD, the simulator (color and thermodynamic) for wet lab, reference databases, templates, ...

docs/blog_18_05/index.md
```diff
@@ -0,0 +1,471 @@
...
+
+```plaintext
+Let me look up: FeCrO4 is reported as brick red to brown. So both Ag+ and Fe+2 could give brick red with chromate.
+```
+
+This is not standard. Ferrous chromate is usually described as yellow or yellow-brown, not brick red, and it is not a diagnostic test for Fe²⁺. The "look up" verb is doing real work here. Despite some tools in the environment allowing to consult the color of the different compounds and precipitates, there is no tool call, no reference, no database. The phrase is being used to manufacture a piece of evidence in support of an already-preferred hypothesis. The agent then commits:
```
kjappelbaum:
```suggestion
This is not standard. Ferrous chromate is usually described as yellow or yellow-brown, not brick red, and it is not a diagnostic test for Fe²⁺.  Despite some tools in the environment allowing to consult the color of the different compounds and precipitates, there is no tool call, no reference, no database. The phrase "let me look up" is being used to manufacture a piece of evidence in support of an already-preferred hypothesis. The agent then commits:
```

docs/blog_18_05/index.md
```diff
@@ -0,0 +1,471 @@
...
+
+```plaintext
+<final_answer>{"cation": "Fe+2", "anion": "HCO3-"}</final_answer>
+```
+
+A colorless solution, a dark brown hydroxide, a black sulfide, and a brick-red chromate are observations that point much more naturally to Ag⁺ than to Fe²⁺. The resolution is not to revisit the carbonate/bicarbonate commitment and ask whether the Ba²⁺/acid behavior could instead come from the actual anion, F⁻. It is to invent a property of FeCrO₄ that resolves the surface contradiction and lets the earlier commitment stand.
```
kjappelbaum:
```suggestion
A colorless solution, a dark brown hydroxide, a black sulfide, and a brick-red chromate are observations that point much more naturally to Ag⁺ than to Fe²⁺. The resolution of the agent, however,  is not to revisit the carbonate/bicarbonate commitment and ask whether the Ba²⁺/acid behavior could instead come from the actual anion, F⁻. It is to invent a property of FeCrO₄ that resolves the surface contradiction and lets the earlier commitment stand.
```

kjappelbaum: in this writing, the actual answer is a bit hard to parse.

docs/blog_18_05/index.md
```diff
@@ -0,0 +1,471 @@
...
+{"cation": "Fe+2",  "anion": "HCO3-"}
+
+{"cation": "Ag+",   "anion": "F-"}
+```
+
+Every run sees the same colorless solution, the same pH 8, the same flame test. The cation identification covers four different metals across the periodic table; the anion covers three different oxyanions and a halide. The same observation panel is being interpreted as four very different chemical situations.
+
+### **A single measurement became a scaling law**
+
+The AFM domain also contains a more structural version of the same problem. In `afm_experiment_level_4`, the agent is asked to determine the scale-dependent relationship between surface roughness, Rq, and scan area, A. The required answer is not just a roughness value. It is an equation of the form `Rb = expression_in_A`, together with the measured roughness and area values used to fit or justify that relationship.
```
kjappelbaum:
```suggestion
The AFM domain also contains another version of the same problem. In `afm_experiment_level_4`, the agent is asked to determine the scale-dependent relationship between surface roughness, Rq, and scan area, A. The required answer is not just a roughness value, but an equation of the form `Rb = expression_in_A`, together with the measured roughness and area values used to fit or justify that relationship.
```

docs/blog_18_05/index.md
```diff
@@ -0,0 +1,471 @@
...
+
+Then it checks the remaining trial directories. They are empty too. At this point the scientific state is clear: the agent has one usable AFM image, one roughness value, and no variation in scan area. The task asks for a scale-dependent relationship. The evidence does not support one.
+
+The agent says this explicitly:
+
+```plaintext
+Only one trial directory contained an image file, which I successfully analyzed to obtain an Rq value. With only one data point, it's not possible to determine a scale-dependent relationship between Rq and A. I will use this single data point to provide a minimal response, but it won't be sufficient to establish a meaningful relationship.
+```
+
+Then it submits anyway:
```
kjappelbaum: the criticism here could be that we did not let the agent give up.

kjappelbaum: we should probably make clear what we tried/think of this

docs/blog_18_05/index.md
```diff
@@ -0,0 +1,471 @@
...
+
+```plaintext
+{"equation": "Rb = 3.966211875104922e-08",
+
+"Rb": [3.966211875104922e-08],
+
+"A": [256*256]}
+```
+
+This is not just an incomplete answer. It changes the type of claim being made. The agent has evidence for one roughness measurement, but it submits an equation. It also treats the image matrix shape, `256 × 256`, as if it were the scan area, even though the task asks for the physical area of the AFM scan. The missing experiment is converted into a constant "relationship"; the missing area metadata is replaced by a pixel count.
```
kjappelbaum:
```suggestion
The agent has evidence for one roughness measurement, but it submits an equation. It also treats the image matrix shape, `256 × 256`, as if it were the scan area, even though the task asks for the physical area of the AFM scan. The missing experiment is converted into a constant "relationship"; the missing area metadata is replaced by a pixel count.
```

docs/blog_18_05/index.md
```diff
@@ -0,0 +1,471 @@
...
+
+This is the moment to re-run with tighter convergence tolerances, more vacuum padding, or a different EAM potential. The agent does not. It pivots from physics back to the submission format:
+
+```plaintext
+But the task asks for the numerical value without units. Let me check what precision is expected. The example shows "56.00" which suggests 2 decimal places.
+
+So my answer would be: 0.00
+```
+
+The submitted answer is `0.00`. Rounding 0.0042 to two decimal places does give 0.00, but the answer it produces is not just inaccurate — it is unphysical. A clean aluminium surface with literally zero surface energy would not be a surface. The "0.05-0.10" range that the agent had quoted as a sanity check, and that its own number missed by more than 10×, disappears from the workflow; the format example survives.
```
kjappelbaum:
```suggestion
The submitted answer is `0.00`. Rounding 0.0042 to two decimal places does give 0.00, but the answer it produces is unphysical. A clean aluminium surface with literally zero surface energy would not be a surface. The "0.05-0.10" range that the agent had quoted as a sanity check, and that its own number missed by more than 10×, disappears from the workflow; the format example survives.
```

kjappelbaum: Here, too, someone could argue something like: See, agents aren't that bad. They actually spot their own mistakes.

docs/blog_18_05/index.md
```diff
@@ -0,0 +1,471 @@
...
+Run 2: 0.00   ← this trace
+
+Run 3: 0.06
+
+Run 4: 0.06
+
+Run 5: 0.06
+```
+
+Underconvergence is recoverable — four of the five runs recovered from it. The behaviour that singles out run 2 is not that it underconverged; it is that, having noticed the underconvergence, the agent treated the format requirement as the binding constraint and let the answer collapse to zero.
```
kjappelbaum:
```suggestion
Underconvergence is recoverable — four of the five runs recovered from it. The behaviour that singles out it is that, having noticed the underconvergence, the agent treated the format requirement as the binding constraint and let the answer collapse to zero.
```

kjappelbaum:
Again: Thanks for the initative and for pushing it. 

I feel that you almost aimed for too much here in one post. 
The writing still is a bit inconsistent between sections and in one part reads way more authentic than in the other parts. 

I would consider splitting it up into a series of posts. Starting with a first post that either dives into the results or the work behind the scences and later dives deeper into the epistimology/broader context.

docs/blog_18_05/index.md
MrtinoRG: I do not want to promise something than perhaps then we do not deliver

MrtinoRG: So I won't add this, sorry

docs/blog_18_05/index.md
MrtinoRG: That's why it says much and not all.. 🙄

docs/blog_18_05/index.md
MrtinoRG: If I understood the questions correctly, you are asking if the link will work when published.

The answer is NO. The idea would be to merge this PR and, when publishing it elsewhere, replace such links with static hyperlinks that point to the GitHub traces. That's why I included those 5 report traces in the PR, to ensure that they are somewhere that can be more easily inspected than in HF.

MrtinoRG:
@kjappelbaum Thanks for the review, I really appreciate it! ❤️  

I have now reviewed the text (especially the intro) to better match the other sections.

I would rather remove the section about learning from evaluation design than publish several blogs. The reason is that I have added that section, trying to give some value to the codebase and the environment design (such environments are nowhere else). However, the main core of the blog, and the reason for writing it, was to show examples that illuminate why we did the epistemology analysis.

If removed, I can also prepare another blog that goes deeper first into all the efforts in the development to then show all the value that the codebase and environments have. 

You take the final decision here. Let me know

kjappelbaum:
Thanks for being patient with me!

Happy to go with your suggestion to remove the section.

MrtinoRG:
I think it is ready to merge now. And read it and review it entirelly again and I like it

docs/blog_18_05/index.md
kjappelbaum: the question was if we could update https://lamalab-org.github.io/corral/#explainers to show one of the traces. Might be possible if one passes the JSON as URL parameter (or adds there a drag and drop to visualize one of the JSON files)

kjappelbaum (approved):
Thanks for spending so much effort on it, @MrtinoRG!

docs/blog_18_05/index.md
MrtinoRG: Ohhhhhhh I see. 

Let me try to do it :)

MrtinoRG: This is a fantastic idea BTW

MrtinoRG:
I guess before merging everything that is not the site modification should be removed right? 

Like we do not need any blog files to be merged

MrtinoRG:
Done.

Now it should be ready to merge, the text and images are present in previous commits :)

PR #366, opened 2026-06-13
feat: unified runtime state, remove task graph, and add tool API

### LINKED ISSUE #293: Rethink state
In a discussion with @n0w0f we came to the point that we perhaps should reconsider how the state is handled. 
The resetting is a bit inelegant. We should consider if it is its own object, if it is generalized via a "state manager". 

This is relevant if we want to generalize the framework to also allow some learning from previous trials.

### LINKED ISSUE #318: In subtasks, when error in chained tasks pass "No Answer" instead of the error
@aaaghajani reported that it is not ideal that in the subtasks or chained tasks, when there was some error in the subtasks run, if there is another subtasks that depend on this failed one, those inputs should not contain the error but perhaps something like `No Answer`.

MrtinoRG:
## Core changes

The changes in this PR only include the `src` code. When the PR gets accepted I will push the `tasks` update. This is to keep the PR reviewable and focused on the core refactor, which is already quite large.

### Runtime state (CorralState)

- Introduce `src/corral/backend/state.py` with a single `CorralState` dataclass as the one mutable runtime object per run
- Remove scattered state from `Environment` (`trial_states`, `trial_counter`, workspace) and `TaskGroup` (`results`, `scores`) — all now live in `CorralState`
- `Environment`, tools, scoring, and the server all read/write through `CorralState`; task definitions remain immutable config

### Task graph & dependency handling

- Remove `TaskGroup` as a mutable runtime object; replace with an immutable `TaskGraph` / helper functions (`topological_order`, `validate_task_graph`, `ready_tasks`)
- Add `InputRef` + `input_map` on `TaskDefinition` for explicit, named dependency wiring (replaces the magic `result_from_<task_id>` key convention)
- `CorralState` resolves dependency inputs via `resolve_inputs(task)` and stores per-task outputs in `task_runs[task_id]`

### Execution order & correctness

- Runner now enforces topological execution order in chained mode (previously relied on authoring order being correct)
- Add `assert_dependencies_selected` — raises a clear error up front if a selected subset omits a required dependency
- Drop the eager `get_task_prompt()` call at env construction; prompts are now only resolved after dependencies are satisfied, making `NOT YET AVAILABLE` placeholders in prompts unnecessary and removed
- Broken chain short-circuit: when an upstream task fails, downstream tasks are recorded as `score=0.0 / unreachable` with 0 iterations — no agent call, no exception

### Tool API cleanup

- Replace four tool parameters (`available_tools`, `common_tools`, `file_tools`, `file_tool_factory`) with a single `Toolset` value object on `Environment.__init__` and `build_environments`
- `Toolset.resolve(task, workspace)` collapses `_add_task_tools` + `_setup_file_tools` into one call per trial reset
- Add `excluded_tools` to base `TaskDefinition`; wetlab's `_add_task_tools` override is deleted

### Other cleanups

- Single vs. chained tasks share one authoring path (`TaskDefinition` with optional `input_map`); `group_id` is no longer a public argument
- Grouping (connected components) derived automatically by the framework
- `prompt_fn` / `setup_fn` hooks on `TaskDefinition` replace per-benchmark `Environment` subclasses for custom prompts and hardware setup

## Summary by Sourcery

Unify runtime task state management around a new CorralState, replace mutable task groups with an immutable task graph and dependency API, and simplify tool configuration into a single Toolset while updating routing, server, and tests accordingly.

Enhancements:
- Refactor Environment to use a shared CorralState and TaskRunState for all mutable runtime data, including trials, tool calls, and submissions.
- Replace TaskGroup and implicit dependency wiring with TaskDefinition-based dependency graphs, explicit InputRef/input_map, and topological ordering utilities, including connected components and validation.
- Introduce a Toolset value object to centralize tool configuration, including common, pool, and workspace-bound tools, and update Environment and tool resolution to use it.
- Ensure chained runs execute in dependency order with upfront dependency-closure checks, and short-circuit downstream tasks as unreachable when dependencies fail, without invoking the agent.
- Simplify LaTeX generation and server APIs to derive grouping and dependencies from TaskDefinition and CorralState instead of TaskGroup, including a new dependency_graph endpoint.

Tests:
- Add tests for dependency graph validation, topological ordering, and chained short-circuit behaviour, and update existing environment and configure_additional_apps tests to the new APIs.

## Release Notes

* **New Features**
  * Added dependency graph support, including validation and dependency-closed workflow execution.
  * Chained workflows now short-circuit: if an upstream task fails or is missing output, downstream tasks are marked unreachable instead of running.

* **Refactor**
  * Updated the task model to use immutable task definitions and explicit input wiring between tasks.
  * Centralized runtime/session state handling and improved prompt/export behavior for chained runs.

* **Bug Fixes**
  * Trial listing and status responses now reflect the latest completed trial and consistent tool statistics.

* **Documentation / Tests**
  * Updated docs and added/adjusted tests for dependency ordering and chained short-circuit behavior.


kjappelbaum:
Sorry for the delay. I just now reached this part of my inbox and will try to look at it tomorrow.

PR #382, opened 2026-08-18
feat: refactor the codebase

### LINKED ISSUE #9: async methods
As we are doing tons of API calls, we will benefit from implementing things async (this can be done in a refactoring, once we have a clearer idea of how things work).

### LINKED ISSUE #22: set up and document logging framework
we should use some logging framework. As we add more complexity to the system, we want to be able to set up handlers that redirect the output from different subsystems to different files/consoles

_Originally posted by @kjappelbaum in https://github.com/lamalab-org/mat-agent-bench/pull/21#discussion_r1913825738_

### LINKED ISSUE #98: agent result should be a pydantic models or dataclasses
currently we unpack a tuple

### LINKED ISSUE #111: checkpointing/process management
Following up on the discussion with @n0w0f and @MrtinoRG today. We seem to have a problem with restarting from crashes for large environments. This is linked to our checkpointing not working as expected and us not having automatic restart capabilities implemented. 

While there are some checkpointing packages in Python (or interfaces to stuff like [`supervisord`](https://supervisord.org/running.html)), it almost feels as if we would benefit from having a sort of workflow manager. For example, with [`prefect`](https://www.prefect.io): 

At least in prefect, we have a good mapping between tasks and flows (each task being one task in an environment and the entire flow, e.g., the run over one environment). 

```
from prefect import flow, task
import wandb

@task(retries=3, retry_delay_seconds=[1, 10, 100])
def substep_1():
    # Your substep logic here
    wandb.log({"substep_1": "complete"})
    return result

@task(retries=3, retry_delay_seconds=[1, 10, 100])
def substep_2(input_data):
    # Your substep logic here
    wandb.log({"substep_2": "complete"})
    return result

@flow
def main_workflow():
    wandb.init(project="my-long-process", resume="allow")
    
    result1 = substep_1()
    result2 = substep_2(result1)
    # Continue with additional steps
    
    return final_result

if __name__ == "__main__":
    main_workflow()
```

Similar things could be done with `aiida`, `fireworks`. They already handle checkpointing, restarts, dashboarding. 
Downside would be an additional dependency and that we somehow tie us a bit to one of those frameworks.

### LINKED ISSUE #256: Improve output type of agent
> This is a personal taste preference. But I started preferring datastructures with fields (named tuples, pydantic, ...)  over returning such complex things. This makes it easier to avoid misuse and also makes the documentation easier.  

 _Originally posted by @kjappelbaum in [#236](https://github.com/lamalab-org/corral/pull/236/changes#r2617587426)_

### LINKED ISSUE #294: Rethink workspace
In a discussion with @n0w0f we also touched on workspaces being a bit inelegant, and workspaces were also confusing to @AdrianM0 when he started using the framework. 

Right now, they are also limited a bit due to the use of the local filesystem. 

An overkill solution would be to have a "workspace service" where the runner could communicate via an ID to have access to its workspace. The "workspace service" could then control quite carefully what operations are allowed. It could also hide what filesystem is actually being used.

### LINKED ISSUE #377: Preserve native agent status and metadata in task trial reports
## Summary

Preserve Corral-native agent status, error detail, and run metadata in every `TaskTrialResult` and generated JSON report.

Corral agents already return `AgentRunResult.metadata`, but `execute_single_trial()` does not propagate it into `TaskTrialResult`. The generated report also omits enough native terminal detail that consumers cannot always distinguish a scored answer from an execution failure.

This is a Corral result-fidelity issue. Corral should report what it executed; external orchestrators remain responsible for their own IDs, idempotency, experiments, splits, and workflow state.

## Current gaps

- `AgentRunResult.metadata` is dropped after `agent.run_agent()`.
- `TaskTrialResult` has no explicit field for preserved agent or harness metadata.
- JSON reports omit `error_message`.
- The agent terminal status is not preserved consistently after submission and scoring.
- A valid score of zero and some execution failures are difficult to distinguish from the exported report alone.
- Success and exception paths do not expose an equivalent result shape.

## Proposed result additions

Add backward-compatible Corral-native fields to `TaskTrialResult` and JSON report trials:

- `execution_status`
- `termination_reason`
- `error_message`
- `run_metadata`, populated from `AgentRunResult.metadata`
- native timestamps when available

The exact types can be refined, but they should distinguish:

- submitted and scored
- surrendered
- maximum turns or budget exhausted
- timed out
- agent or harness failure
- tool infrastructure failure
- submission failure
- scorer failure
- cancelled

Do not encode these distinctions only in human-readable error strings.

## Metadata behavior

- Preserve agent-specific harness metadata without flattening or renaming unknown fields.
- Record actual values known by Corral, not external workflow metadata.
- Keep secrets, access tokens, credential-bearing URLs, and authorization headers out of reports.
- Hash large native tool schemas or prompt/configuration content when embedding full values would be unsafe or excessive.
- Keep report serialization backward compatible where possible through optional additive fields.

Corral may report native details it already owns, such as agent class, model, harness version, tool-schema digest, configured limits, and native seed. It does not need to accept or echo FLEET experiment IDs, request IDs, idempotency keys, phases, split manifests, or task-role assignments.

## Implementation sketch

1. Add typed status and termination values or constrained strings.
2. Extend `TaskTrialResult` with run metadata and complete terminal detail.
3. Propagate `AgentRunResult.metadata` through success, surrender, and exception paths.
4. Preserve the original agent terminal status before submission and scoring.
5. Include the new fields in `BenchmarkResult._prepare_report_data()`.
6. Add secret-redaction tests for native metadata.
7. Document the additive report fields.

## Acceptance criteria

- [ ] `AgentRunResult.metadata` survives into `TaskTrialResult`.
- [ ] Generated JSON reports contain the preserved run metadata.
- [ ] Reports expose explicit execution status, termination reason, and error detail.
- [ ] A valid score of zero is distinguishable from timeout, harness failure, submission failure, and scorer failure.
- [ ] Success, surrender, budget exhaustion, timeout, tool failure, and unexpected exception paths have tests.
- [ ] Agent, model, harness, native tool-schema, and configured-budget metadata can be recorded when Corral knows them.
- [ ] Report serialization does not expose configured secrets.
- [ ] Existing report consumers have an additive compatibility path.

## Non-goals

- external request or experiment correlation
- external idempotency enforcement
- benchmark split manifests or task-role assignment
- FLEET experiment, trial, episode, or memory models
- workflow retries
- a message broker
- distributed execution
- changing task scoring semantics

External consumers can wrap the synchronous Corral invocation and attach the returned native result to their own request and workflow records.

MrtinoRG:


src/corral/core/_immutable.py
```diff
@@ -0,0 +1,123 @@
...
+from collections.abc import Mapping
+from typing import Any
+
+from pydantic import BaseModel, ConfigDict, model_validator
+
+SHA256_HEX_LENGTH = 64
+_SHA256_HEX_CHARACTERS = frozenset("0123456789abcdef")
+
+
+class FrozenDict(dict[str, Any]):
```
n0w0f, line 14: New python seems to have this [inbuilt](https://docs.python.org/3.15/whatsnew/3.15.html#whatsnew315-frozendict)

I am surprised ,pydantic or other libs do not have this.

src/corral/core/state.py
```diff
@@ -0,0 +1,372 @@
...
+)
+
+
+def new_state_id() -> str:
+    """Mint a globally unique identifier for a State revision chain."""
+    return str(uuid4())
+
+
+class StateMetadata(FrozenModel):
+    """Durable configuration needed to understand or restore an execution."""
```
n0w0f: execution here means a trial or a step within a trial ?

src/corral/core/state.py
```diff
@@ -0,0 +1,372 @@
...
+
+    model: Mapping[str, JSONValue] = Field(default_factory=dict)
+    scaffold: Mapping[str, JSONValue] = Field(default_factory=dict)
+    environment: Mapping[str, JSONValue] = Field(default_factory=dict)
+    task: Mapping[str, JSONValue] = Field(default_factory=dict)
+    extra: Mapping[str, JSONValue] = Field(default_factory=dict)
```
n0w0f:
```suggestion
```

This field might get exploited later

src/corral/core/state.py
```diff
@@ -0,0 +1,372 @@
...
+from pydantic import JsonValue as JSONValue
+
+from corral.core._immutable import FrozenModel, validate_sha256_hex
+from corral.core.action import Action
+from corral.core.workspace import WorkspaceState
+
+if TYPE_CHECKING:
+    from datetime import datetime
+
+STATE_SCHEMA_VERSION = 2
```
n0w0f: Why is this 2 ?

src/corral/core/state.py
```diff
@@ -0,0 +1,372 @@
...
+    llm_calls: int = Field(default=0, ge=0)
+    tool_calls: int = Field(default=0, ge=0)
+    agent_steps: int = Field(default=0, ge=0)
+    metadata: Mapping[str, JSONValue] = Field(default_factory=dict)
+
+
+class RuntimeState(FrozenModel):
+    """Serializable progress needed by the next execution step."""
+
+    status: Literal[
```
n0w0f, line 35: This literal can be defined separately I believe

src/corral/core/state.py
```diff
@@ -0,0 +1,372 @@
...
+
+class UsageState(FrozenModel):
+    """Cumulative resource use recorded by committed transitions."""
+
+    input_tokens: int = Field(default=0, ge=0)
+    output_tokens: int = Field(default=0, ge=0)
+    llm_calls: int = Field(default=0, ge=0)
+    tool_calls: int = Field(default=0, ge=0)
+    agent_steps: int = Field(default=0, ge=0)
+    metadata: Mapping[str, JSONValue] = Field(default_factory=dict)
```
n0w0f: I am thinking if we should keep those metadata fields in the data classes here and below (`RuntimeState` and `TaskOutput` ) ? Concern again being we might start storing junks here.

src/corral/core/state.py
```diff
@@ -0,0 +1,372 @@
...
+        separators=(",", ":"),
+        sort_keys=True,
+    )
+
+
+def _sha256(payload: str) -> str:
+    return hashlib.sha256(payload.encode("utf-8")).hexdigest()
+
+
+class State(FrozenModel):
```
n0w0f: I am thinking with this if it would be possible for to recreate a tree of states like this ?

Image

src/corral/core/state.py
```diff
@@ -0,0 +1,372 @@
...
+
+def checkpoint_state(parent: State, source: State) -> State:
+    """Create one durable child containing the complete `source` value.
+
+    Agents may create several immutable in-memory forks while composing a model
+    message or updating their namespaced runtime data. Persistence checkpoints
+    only the semantically important boundaries. This helper squashes those
+    intermediate forks into one direct child of the latest durable `parent`
+    while preserving the complete environment, agent history, workspace,
+    usage, runtime, and dependency state.
```
n0w0f: This exist because parallel tool call could be part of one revision ?

src/corral/core/task.py
n0w0f, line 34: Do we need this anymore ? This was for subtasks ?

src/corral/core/task.py
```diff
@@ -7,7 +7,19 @@
...
+    from corral.core.state import State
+
+
+@dataclass(frozen=True)
+class EnvironmentSetup:
+    """Serializable values produced while configuring one task execution."""
+
+    hidden_arguments: Mapping[str, JsonValue] = field(default_factory=dict)
+    values: Mapping[str, JsonValue] = field(default_factory=dict)
+    status: str = "Additional apps/services configured for this task."
```
n0w0f, line 22: not sure about the variable name status here

src/corral/core/task.py
```diff
@@ -24,23 +36,24 @@ class InputRef:
...
 class TaskDefinition:
-    """Definition of a task with its requirements and scoring.
+    """Definition of a task with its requirements and evaluation hook.
 
     Tasks are immutable configuration describing the static DAG: dependencies
-    are declared per input field via `input_map`. All runtime data (outputs,
-    scores, resolved dependency values) lives in
-    `corral.backend.state.CorralState`.
+    are declared per input field via `input_map`. Runtime inputs and outputs
+    live in immutable :class:`corral.core.state.State` revisions.
```
n0w0f: Here and below some of the methods is for chaining tasks, do we need them anymore ?

src/corral/core/tool_description.py
```diff
@@ -0,0 +1,87 @@
...
+        sections["BASIC"] = untagged
+    return sections
+
+
+def default_tool_description(description: str) -> str:
+    """Return the one fixed runtime description for a tool."""
+    if not description:
+        return "No description available"
+    sections = _sections(description)
+    return sections.get("BRIEF") or sections.get("BASIC") or description.splitlines()[0]
```
n0w0f, line 67:
```suggestion
    return sections.get("BRIEF") or sections.get("BASIC")
```

I do not know if we should have fallbacks like this, probably it is better to raise an error so we know docstrings are not correct

src/corral/core/tool_description.py
```diff
@@ -0,0 +1,87 @@
...
+
+def _sections(description: str) -> dict[str, str]:
+    sections = {
+        name.upper(): _clean_nested_tags(content)
+        for name, content in _TAG.findall(description)
+        if content.strip()
+    }
+    untagged = _TAG.sub("", description)
+    untagged = _ARGS.sub("", untagged)
+    untagged = _RETURNS.sub("", untagged).strip()
```
n0w0f, line 56: below you take for args , but do we want to remove the returns description entirely ??

src/corral/core/tool_catalog.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+class ToolCatalogMismatchError(ToolCatalogBindingError):
+    """The persisted and currently executable tool catalogs differ."""
+
+
+def _tool_name(tool: Mapping[str, Any], *, index: int) -> str:
+    if tool.get("type") != "function":
+        raise ValueError(f"tool catalog entry {index} must have type='function'")
+    function = tool.get("function")
```
n0w0f, line 41: Mmm, I am thinking if this would always be the case, for native function calling by providers, the type is not always function, but could be interpreter etc right ?

src/corral/core/tool_catalog.py
```diff
@@ -0,0 +1,219 @@
...
+
+def canonical_tool_catalog_json(
+    tools: Sequence[Mapping[str, Any]],
+) -> str:
+    """Return a stable, order-independent representation for fingerprinting.
+
+    Tool order is intentionally excluded from compatibility: the persisted
+    snapshot remains authoritative for presentation order, while Environment
+    execution dispatches by unique tool name. Every schema value remains part
+    of the fingerprint.
```
n0w0f, line 58: Assumption here would be that the order in which tool descriptions are provided lets say in the prompt do not matter ?

src/corral/core/tool.py
n0w0f, line 1: There seems to be still some inconsistency around Python Type Hints with JSON Schema Types around tools

Image


For defining tools we might also have to enforce 
 1. All arguments and returns must be JSON-native.
2. Nullability must be handled via Pydantic, not strings.


And then also 
1. we replace `validate_arguments` with https://github.com/python-jsonschema/jsonschema
2. `preprocess_arguments` becomes much simpler
3. enforce that the execute wrapper always outputs JSON

n0w0f, line 120: I am thinking if this `hidden_args`  can silently make tools stateful. We have cases where hidden argument passes a mutable Python objects ?
The pattern below seems like a good pattern but introduces a new concept here and maybe need longer discussion

Image

src/corral/core/transition.py
```diff
@@ -0,0 +1,273 @@
...
+# as an API/instrument idempotency key. It is runtime-owned and never supplied
+# by the agent or persisted in the general hidden-argument namespace.
+CORRAL_ACTION_ID_ARGUMENT = "corral_action_id"
+
+
+@dataclass(frozen=True, slots=True)
+class ToolExecutionResult:
+    """A tool observation plus an optional complete environment namespace.
+
+    Most tools only return content. A stateful tool returns this value with a
```
n0w0f: stateful tool here means a tool that would change state of the environment ?

src/corral/backend/executors.py
n0w0f, line 219: This seems like might not terminate the entire process group that was spwaned

n0w0f (commented):
I think it is not possible to review this properly as it is very very modular and many changes at once.

Maybe to simplify a bit,

1. There is also two independent concurrency/orchestration systems doing conceptually similar jobs. temporal for task execution and customs code for background tool calls. We probably need to simplify this.
2. We can get rid of the slurm stuff.
3. Modal's own SDK supports async waiting on a spawned call, we can get rid of the custom polling mechanism.
4. AgentSession spinning up a real HTTP/MCP server per session also seems very complex
5. We should also try to find some std library to get rid of some of the custom code we have for sqlite.

n0w0f:
Screenshot 2026-09-04 at 17 22 38

Based on the definition here for an activity, i am thinking it make more sense for us to define an action itself as an activity.

- Then temporal would take care of the concurrency and retry mechanism.
- It also make sense than an entire agent iteration loop is then a workflow.

n0w0f:
> Screenshot 2026-09-04 at 17 22 38
> Based on the definition here for an activity, i am thinking it make more sense for us to define an action itself as an activity.
> 
> * Then temporal would take care of the concurrency and retry mechanism.
> * It also make sense than an entire agent iteration loop is then a workflow.

Now I think about another suboptimal thing here which is that, in the case of harness, the entire harness loop would be activity as it might not be able to define natively their tool call as temporal activity.

MrtinoRG:
Thanks for the comments @n0w0f 

I have now added some changes on the points that you have mentioned. I explain below:

I have removed the Slurm and Modal executors.

On the two different executions I have removed the one powered by temporal. 

I saw your comments about making each action as an temporal activity. However, I have not implemented this. On the tool side, this would create a problem around the timeouts and I do not think that the temporal adds that much value. On the agent side, that we discussed at some point, the SDKs would need a different management, and this would add overhead.

On the MCP servers, I have moved all the complexity into the env deploying one server per task. This only partially adresses the issue, but I am not sure how to further reduce the complexity

On the SQL point, I have now moved the database commands to a dedicated package. However, it is still quite complex, but I do not think it is possible to further reduce the complexity as the current code is mostly around the Corral database schema.

Some things to keep in mind is to update the docs, and to have a look on the AFM, but this later can be difficult without access to the experiments and software

src/corral/core/environment.py
```diff
@@ -0,0 +1,869 @@
...
+        *,
+        created_by_action: str | None = None,
+    ) -> WorkspaceState:
+        """Capture the task's materialized directory as a logical workspace."""
+        if not self.workspace_path:
+            return previous
+        if self.workspace_manager is None:
+            raise RuntimeError("a bound workspace requires a WorkspaceManager")
+        return asyncio.run(
+            self.workspace_manager.snapshot(
```
n0w0f, line 380: Could this already be inside an event loop ?

src/corral/cli.py
```diff
@@ -0,0 +1,816 @@
...
+            "corral.agents.reflexion_agent",
+            "ReflexionAgent",
+        ),
+        "terminus": AgentDefinition("corral.agents.terminus", "TerminusAgent"),
+        "tool-calling": AgentDefinition(
+            "corral.agents.tool_calling",
+            "ToolCallingAgent",
+            # Provider-native function tools and the reasoning mode enabled by
+            # default for some GPT models cannot be combined reliably.
+            default_kwargs=(("reasoning_effort", "none"),),
```
n0w0f, line 80: It looks like Anthropic's LiteLLM provider rejects the literal string "none"

That exception also gets swallowed somewhere upstream.

n0w0f (approved):
I think we can merge now @MrtinoRG

docker/benchmark.Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM python:3.12-slim
```
n0w0f: This is critical before merge. 
base image (python:3.12-slim) has no git, but corral's pyproject.toml pulls promptstore from a git
so we might need git in this image

docker/benchmark.Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+ENV PYTHONDONTWRITEBYTECODE=1 \
+    PYTHONUNBUFFERED=1
+
+RUN groupadd --gid 10001 corral-terminal \
+    && useradd --uid 10001 --gid 10001 --no-create-home --home-dir /workspace corral-terminal
+
+WORKDIR /opt/corral
+COPY . /opt/corral
+RUN python -m pip install --no-cache-dir .
```
n0w0f: This is also critical before merge
I think in `runtime/environment_loader.py` 

you resolve path as below 
```
def _repository_root() -> Path:
    return Path(__file__).resolve().parents[3]
```

This would work only if we pip install -e . / uv sync from the monorepo.
In current format copies the package to `site-packages/corral/...`

PR #393, opened 2026-09-10
fix: solve issues #390, #391 and #392

### LINKED ISSUE #390: Could this already be inside an event loop ?
Could this already be inside an event loop ?

_Originally posted by @n0w0f in https://github.com/lamalab-org/corral/pull/382#discussion_r3966063387_

### LINKED ISSUE #391: It looks like Anthropic's LiteLLM provider rejects the literal string "none"
It looks like Anthropic's LiteLLM provider rejects the literal string "none"

That exception also gets swallowed somewhere upstream.

_Originally posted by @n0w0f in https://github.com/lamalab-org/corral/pull/382#discussion_r3966175514_

### LINKED ISSUE #392: benchmark.Dockerfile changes
- [ ] base image (python:3.12-slim) has no git, but corral's pyproject.toml pulls promptstore from a git
so we might need git in this image
- [ ] in runtime/environment_loader.py

we resolve path as below
```
def _repository_root() -> Path:
    return Path(__file__).resolve().parents[3]
```
This would work only if we pip install -e . / uv sync from the monorepo.
In current format copies the package to site-packages/corral/...

MrtinoRG:
- Solve issues #390, #391, and #392
- Simplify the `llm_call` function so that no arguments are set unexpectedly
- The issue #392 stripped the failing isolation in the Docker container: the agent could look in the files in `/opt` for the codebase. Therefore, I tried to solve this by banning the agents' access to all the directories inside the Docker container except in `/workspace`. The permissions right now are managed by Unix permissions. This required defining a worker for the agents and ensuring that the spawned subprocesses of the agents also do not have those permissions.

## Summary by Sourcery

Harden Docker execution around workspace-only access while correcting agent branching, LLM configuration, SDK timeouts, and failure reporting.

New Features:
- Enforce workspace-root isolation for Docker agents, tools, subprocesses, and background jobs through restricted worker identities and filesystem boundaries.
- Provide isolated per-node workspaces for AI Scientist branches with explicit artifact promotion to the owning trial workspace.
- Apply configurable OpenHands MCP timeouts across both transport and executor layers, including native-tool collision handling.
- Add Playwright support to the OpenHands dependency extra and prepare Docker images for the permission policy.

Bug Fixes:
- Stop agents from accessing private code, checkpoints, task state, and sibling workspaces inside Docker containers.
- Preserve caller-supplied LLM configuration without injecting model-, temperature-, token-, or tool-selection arguments.
- Preserve workspace roots and private runtime directories during Modal synchronization.
- Expose agent and trial failure messages in CLI output and benchmark results.

Enhancements:
- Broker restricted agent session operations through a narrow JSON interface while keeping private controller state out of worker processes.
- Refine session projections and agent APIs to expose only public model, conversation, evaluation, and submission data.
- Classify tools as trusted or workspace-bound and fail closed when untrusted tools request private inputs.
- Update the Docker runtime protocol and launch configuration to require the new permission policy.

Build:
- Prepare benchmark and wetlab Docker images with required packages and protected image permissions.

Deployment:
- Strengthen Docker isolation using a root bootstrap, dropped worker privileges, private mount namespaces, and workspace-only mounts.

Tests:
- Add coverage for Docker permission boundaries, worker descendants, isolated node workspaces, artifact promotion, OpenHands timeout enforcement, LLM argument preservation, CLI failures, and safe Modal synchronization.

## Summary by CodeRabbit

* **New Features**
  * Added permission-controlled execution for agents, tools, background jobs, and Docker workers.
  * Added isolated workspaces for branched executions, with support for promoting branch artifacts.
  * Added configurable OpenHands MCP and tool execution timeouts.
  * Added workspace-aware tool metadata and trusted-tool execution.
  * Added clearer reporting of task and trial failures.

* **Bug Fixes**
  * Improved workspace path protection, snapshot safety, and preservation during remote updates.
  * Prevented private runtime files and sensitive state from crossing execution boundaries.
  * Standardized forwarding of language-model request parameters.

### DIFF
--- a/tests/runtime/test_permissions.py
+++ b/tests/runtime/test_permissions.py
@@ -1,0 +1,1403 @@
+"""Exercise real Unix permission checks in the trial image (no model API calls).
+
+Run with CORRAL_PERMISSION_TESTS=1 inside the Docker image. Opting in makes
+missing root privileges or an incorrectly protected image a failure, not a skip.
+"""
+
+# Exercise imports after privilege dropping; SDK extras are optional locally.
+# ruff: noqa: PLC0415
+
+import os
+import threading
+
+import pytest
+
+from corral.agents.ai_scientist.agent import AIScientistAgent
+from corral.agents.session import AgentSessionCapabilities
+from corral.core.environment import Environment
+from corral.core.tool import tool
+from corral.runtime import permissions
+
+pytestmark = pytest.mark.skipif(
+    os.environ.get("CORRAL_PERMISSION_TESTS") != "1",
+    reason="requires the real Docker trial permission boundary",
+)
+
+
+@pytest.fixture(autouse=True)
+def restricted_runtime(tmp_path):
+    assert os.geteuid() == 0
+    checkpoints = tmp_path / "checkpoints"
+    checkpoints.mkdir(mode=0o700)
+    permissions.configure(checkpoints)
+    yield
+    permissions._enabled = False
+
+
+@pytest.fixture()
+def workspace():
+    import tempfile
+
+    return tempfile.mkdtemp(prefix="permission-test-", dir="/workspace")
+
+
+@tool
+def permission_probe() -> str:
+    """Attempt reads and writes both directly and through detached descendants."""
+    import json
+    import os
+    import subprocess
+    import sys
+    from pathlib import Path
+
+    result = {"uid": os.getuid(), "gids": os.getgroups(), "denials": {}}
+    for target in (
+        "/opt/corral/pyproject.toml",
+        "/opt/corral/tasks",
+        "/corral-state",
+        "/etc/passwd",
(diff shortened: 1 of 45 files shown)


tests/runtime/test_permissions.py
```diff
@@ -0,0 +1,1403 @@
+"""Exercise real Unix permission checks in the trial image (no model API calls).
+
+Run with CORRAL_PERMISSION_TESTS=1 inside the Docker image. Opting in makes
```
n0w0f, line 3: Should we set this in ci or is it meant for only testing locally ?

n0w0f (approved):
I understood that a lot of code is around allocating permissions within containers, maybe we can think about maybe later about running those tools that require hidden arguments in 'Remotesession` or a different container.

PR #395, opened 2026-09-11
feat: new version of resistor network environment 

n0w0f:
- Added a sampler to sample more circuits. 
- Added another level with more complexity (bridges and more number of resistance present)
- Added another validation tool

## Summary by Sourcery

Replace the resistor-network benchmark with reproducible multi-level circuit generation, stronger validation, and topology-aware conductance scoring.

New Features:
- Add procedurally generated resistor-network benchmark levels with reproducible static task generation, including more complex bridge-based circuits.
- Add circuit visualization support for inspecting sampled task topologies.
- Add conductance-map scoring that accepts functionally equivalent resistor naming and parallel-group representations while enforcing the expected network structure.
- Add circuit-topology validation as an agent tool.

Bug Fixes:
- Correct delta-to-wye transformation corner mappings and strengthen circuit simulation and resistor validation for malformed, non-finite, disconnected, or dangling networks.
- Prevent generated tasks from containing resistors that are effectively unobservable from the published measurements.
- Ensure proposed bridge topologies connect all declared resistors.

Enhancements:
- Refactor benchmark loading and execution to select level-specific task directories through the runtime CLI.
- Centralize nodal-analysis simulation and tighten topology and measurement input validation.

Build:
- Update the resistor-network environment dependencies and lockfile for the new sampler and visualization functionality.

Documentation:
- Document the two benchmark levels, sampler constraints, task regeneration workflow, visualization commands, and updated benchmark execution and scoring behavior.

Tests:
- Add coverage for sampling, circuit composition, measurement selection, task determinism, load-bearing resistor guarantees, conductance scoring, topology validation, simulator validation, and transformation correctness.

Chores:
- Replace the previous hand-authored/chained task assets with generated level-specific benchmark tasks.

## Summary by CodeRabbit

* **New Features**
  * Added procedurally generated resistor-network challenges across two difficulty levels.
  * Added circuit visualization with labeled resistor networks and highlighted terminals.
  * Added topology validation and expanded circuit-analysis tools.
  * Added command-line workflows for task generation, visualization, and benchmark execution.

* **Improvements**
  * Scoring now evaluates equivalent conductance maps with tighter tolerance.
  * Circuit calculations provide stronger input validation and corrected transformation results.
  * Tasks now include sampled, partially observed measurements for realistic inference challenges.

* **Documentation**
  * Updated guidance for task generation, visualization, scoring, and benchmark usage.


tasks/wetlab/compositions.pkl
MrtinoRG, line 1: This file should not be here... Can you remove it before merging, please?

tasks/resistor_network/src/resistor_network/sampler.py
```diff
@@ -0,0 +1,615 @@
...
+        for neighbor in adjacency[node]:
+            if neighbor not in reachable:
+                reachable.add(neighbor)
+                frontier.append(neighbor)
+    if reachable != nodes:
+        return False  # no disconnected/open component
+
+    quality = _sampled_topology_quality(topology)
+    if quality[1] < min_cycle_rank:
+        return False
```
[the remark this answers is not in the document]
MrtinoRG: I think that the bots are right here @n0w0f...  `_sampled_topology_quality()` returns `(cycle_rank, parallel_groups, branch_nodes, node_count)`, so `quality[1]` is the number of parallel groups. Should this be `quality[0]`?

tasks/resistor_network/src/resistor_network/sampler.py
```diff
@@ -0,0 +1,628 @@
...
+        degrees[a] += 1
+        degrees[b] += 1
+        pair = tuple(sorted((a, b)))
+        pairs[pair] = pairs.get(pair, 0) + 1
+    cycle_rank = len(edges) - len(nodes) + 1
+    parallel_groups = sum(count > 1 for count in pairs.values())
+    branch_nodes = sum(
+        degree >= 3 for node, degree in degrees.items() if node not in {"A", "B"}
+    )
+    return cycle_rank, parallel_groups, branch_nodes, len(nodes)
```
MrtinoRG, line 366: Could we calculate complexity using the same graph representation as the scorer? The scorer combines parallel resistors into a single conductance, while this function counts each resistor as a separate edge. As a result, a simple parallel pair can increase the cycle rank even though the scorer sees only one connection. Using the same canonical graph in both places would make the difficulty measure more meaningful.

tasks/resistor_network/src/resistor_network/score.py
```diff
@@ -616,3 +576,115 @@ def check_valid_circuit_json(json_path: str) -> float:
...
+
+def conductance_map(topology: dict[str, Any]) -> dict[tuple[str, str], float]:
+    """Return merged conductance in siemens for each node pair."""
+    conductances: dict[tuple[str, str], float] = {}
+    for node_a, node_b, resistor_id in topology["connections"]:
+        pair = (node_a, node_b) if node_a < node_b else (node_b, node_a)
+        conductances[pair] = (
+            conductances.get(pair, 0.0) + 1.0 / (topology["resistors"][resistor_id])
+        )
+    return conductances
```
MrtinoRG, line 637: Here is where I would change the scorer to follow the same complexity metrics as in the sampler I think

tasks/resistor_network/src/resistor_network/sampler.py
```diff
@@ -0,0 +1,628 @@
...
+    if allow_bridge and target_resistors >= 5 and rng.random() < bridge_prob:
+        bridge = leaf_bridge(rng, ids)
+        remainder = target_resistors - bridge.num_resistors
+        if remainder <= 0:
+            return bridge
+        extra = build_random_block(
+            rng, ids, remainder, max_depth - 1, allow_bridge, bridge_prob
+        )
+        pair = [bridge, extra] if rng.random() < 0.5 else [extra, bridge]
+        return compose_series(pair)
```
MrtinoRG, line 232: Here, instead of allowing a bridge and adding one in such an artificial way, would it be better to enforce one or more bridges by enhancing the `build_random_block` function?

In the current code, the bridge is always attached to the remaining network as a separate series block.

tasks/resistor_network/src/resistor_network/sampler.py
```diff
@@ -0,0 +1,628 @@
...
+    """
+    nodes = sorted({n for conn in topology["connections"] for n in conn[:2]})
+    topology_json = json.dumps(topology)
+    measurements = []
+    for node_a, node_b in itertools.combinations(nodes, 2):
+        resistance = get_resistance_between_nodes(topology_json, [node_a, node_b])
+        measurements.append(
+            {"node_a": node_a, "node_b": node_b, "resistance": round(resistance, 3)}
+        )
+    return measurements
```
MrtinoRG, line 288: I am concerned that giving the resistance between every pair of visible nodes makes the inverse problem much easier than intended. 

As it is right now, the graph (which I assume you introduced for complexity control) is also penalizing, as there are very clearly defined structures in the tasks. Would it be a way to have more fused graphs instead of the current linear ones (especially for level 2)?

MrtinoRG (commented):
I think that right now, apart of the bug that the bots flagged, the major issue is that circuits are quite fixed in shape, which reduce a lot the potential complexity. I assume that this was to better fix the complexity, but I would not fix it on the upper bound for the level 2. 

So please let me know if you have major issues on making more freely difficult circuits for the level 2 instead of following such fixed graph structure? 

Also, have you done some run to check the performance?

n0w0f:
> I think that right now, apart of the bug that the bots flagged, the major issue is that circuits are quite fixed in shape, which reduce a lot the potential complexity. I assume that this was to better fix the complexity, but I would not fix it on the upper bound for the level 2.
> 
> So please let me know if you have major issues on making more freely difficult circuits for the level 2 instead of following such fixed graph structure?
> 
> Also, have you done some run to check the performance?

I have now increased the cycle ranks in both levels, this means there are more cycles than linear chains that we had previously.

The reason for capping the number of resistors was to reduce the number of iteration that might be required to solve this task. I have now increased this cap. Additionally now i don't show all the measurements, which further increase the complexity.

I haven't checked the performance yet, but the level 1 circuits now almost start from the complex circuits we had in our previous pre-print.

tasks/resistor_network/src/resistor_network/sampler.py
n0w0f: I have now removed 15 % of the measurement such that it is still possible to recover

tasks/resistor_network/src/resistor_network/sampler.py
n0w0f: The sampler is now update to allow for more complex circuits

tasks/resistor_network/src/resistor_network/score.py
n0w0f: This is now updated

tasks/wetlab/compositions.pkl
n0w0f: done

MrtinoRG:
So we discussed that simply removing the upper difficulty constraint would give circuits with a lot of nesting and very difficult topologies that would not be solvable. 

We also discussed that the scoring should at least check that the number of resistances matches. 

Last, we discussed that each task would receive only the required info of the measurements. This is, for example, when only 60% of the measurements are needed to solve the task; only that 60% will be given.

I see that you implemented the last point partially by having a fixed 85% of the info given. However, I am not sure if this is ok with all the tasks; for some tasks, this might be too much info, while for others it might be too little, making the task not solvable. One possibility, not super efficient, would be to iterate, removing different parts until one gets the minimum that makes the task solvable. For the solvable part, this could be done with the rule of linear algebra that the number of equations has to be greater than or equal to the number of independent variables. This would mean there are at least as many independent measurements as unknown resistor values. I think this would already give a better approximation than the fixed 85% that you have right now. 

Then are you still planning to add the number of resistors as a check that has to be met during scoring, even if this is with an equivalent system?

n0w0f:
> So we discussed that simply removing the upper difficulty constraint would give circuits with a lot of nesting and very difficult topologies that would not be solvable.
> 
> We also discussed that the scoring should at least check that the number of resistances matches.
> 
> Last, we discussed that each task would receive only the required info of the measurements. This is, for example, when only 60% of the measurements are needed to solve the task; only that 60% will be given.
> 
> I see that you implemented the last point partially by having a fixed 85% of the info given. However, I am not sure if this is ok with all the tasks; for some tasks, this might be too much info, while for others it might be too little, making the task not solvable. One possibility, not super efficient, would be to iterate, removing different parts until one gets the minimum that makes the task solvable. For the solvable part, this could be done with the rule of linear algebra that the number of equations has to be greater than or equal to the number of independent variables. This would mean there are at least as many independent measurements as unknown resistor values. I think this would already give a better approximation than the fixed 85% that you have right now.
> 
> Then are you still planning to add the number of resistors as a check that has to be met during scoring, even if this is with an equivalent system?


I am not sure if i should add scoring based on equal number of resistors, because I think it is fair to propose equivalent circuit even if that mean they have less number of resistance. This could be done in level 1, but for level 2 this would make the task impossible, as they would need to do those transformation to reduce those circuit anyway to solve them.

The optimization you suggested could be implemented for reducing the number of resistance values shown, but then this would force the model to using mathematical methods to solve this task than inquiry based ones. 

I think the circuits are already complex enough , and increasing complexity would now mean, need for mathematical methods or large number of iterations. One option is that i keep circuits not complex and show very minimal resistance.

Let me know what changes would you still like me to have implemented based on this.

n0w0f:
@MrtinoRG , based on our discussion I now made the following changes

- The scorer now requires the submitted circuit to contain the same number of physical resistors as the target. Functional equivalence is still allowed when the resistor count is preserved.

- The fixed 85% measurement rule was replaced with task-specific selection. During generation, we compute a numerical measurement Jacobian and select node pairs until the measurements have full local rank relative to the number of resistor values.

- added number of resistors in the tasks. in notes

tasks/resistor_network/src/resistor_network/sampler.py
```diff
@@ -0,0 +1,598 @@
...
+        low = json.loads(topology_json)
+        high = json.loads(topology_json)
+        low["resistors"][resistor_id] = value * math.exp(-step)
+        high["resistors"][resistor_id] = value * math.exp(step)
+        low_json = json.dumps(low)
+        high_json = json.dumps(high)
+        for row, pair in enumerate(pairs):
+            low_measurement = get_resistance_between_nodes(low_json, list(pair))
+            high_measurement = get_resistance_between_nodes(high_json, list(pair))
+            jacobian[row, column] = (high_measurement - low_measurement) / (2 * step)
```
MrtinoRG, line 267: I think here you have to be careful about how you manage the parallel resistors between nodes, as parallel resistors are not independent, but rather proportional to each other

tasks/resistor_network/src/resistor_network/sampler.py
```diff
@@ -0,0 +1,598 @@
...
+
+    # Add pairs greedily by rank increase.
+    while current_rank(chosen) < num_unknowns:
+        candidates = [pair for pair in shuffled if pair not in chosen]
+        if not candidates:
+            break
+        ranked = [(current_rank(chosen | {pair}), pair) for pair in candidates]
+        best_rank = max(rank for rank, _pair in ranked)
+        best_pairs = [pair for rank, pair in ranked if rank == best_rank]
+        chosen.add(rng.choice(best_pairs))
```
MrtinoRG, line 356: And the issue that I pointed out above was silent because here all candidates are added, and then it breaks, returning the measurements anyway, even if the number of unknowns is still lower than the chosen

MrtinoRG (commented):
I think that the issue that I pointed to is the only thing. Everything else seems to be in a greate shape. Thanks for pushing this ❤️

MrtinoRG (approved):

tasks/resistor_network/src/resistor_network/sampler.py
n0w0f: The measurement-selection loop now verifies the final rank after exhausting candidates. If full effective rank cannot be reached, it returns failure and the sampler retries the circuit .

tasks/resistor_network/src/resistor_network/sampler.py
n0w0f: For parallel resistors, the Jacobian now uses the number of unique node-pair conductances as the target rank rather than the number of physical resistors. Parallel resistors therefore contribute one observable electrical parameter, while the separate resistor-count check still enforces the required number of  physical components.

Since resistor values are now restricted to the disclosed pool, the generator also checks that parallel-group assignments are unique.

PR #396, opened 2026-09-13
feat: add Claude Code bare mode and defer agent SDK imports

MrtinoRG:
The `--bare` argument makes the Claude Code agents not look for SKILLs, Agents.md, skills, and any other similar files. It seems that the power of this argument is quite powerful. 

I also modified the Docker images (general and wetlab) to be equivalent to each other and install the SDKs only when needed. 

Finally, I modified the CLI such that the SDKs are not loaded there, avoiding their installation for Docker-based runs.

## Summary by Sourcery

Run benchmark agents with isolated scratch workspaces and environment-specific Docker runtimes while deferring SDK loading and enabling Claude Code bare mode by default.

New Features:
- Enable Claude Code bare mode by default, with an option to restore the full native tool set.

Bug Fixes:
- Prevent Docker-based benchmark runs from loading agent SDKs on the host.

Enhancements:
- Select benchmark images and runtime dependencies per environment and agent harness, with shared installation and build-argument handling.
- Support isolated per-execution scratch workspaces that can be used without exposing workspace tools or private task state.
- Improve workspace and WetLab state persistence, restoration, and isolation across executions.

Build:
- Update benchmark and WetLab Docker images to share runtime setup and install harness-specific dependencies only when required.

Tests:
- Add coverage for Claude bare-mode behavior, environment-specific Docker builds, deferred SDK loading, isolated workspaces, and WetLab state restoration.

## Summary by CodeRabbit

* **New Features**
  * Added configurable Claude agent “bare” mode with restricted native tools and API-key authentication.
  * Added environment-specific Docker image selection and optional build extras.
  * Added isolated per-execution workspaces that preserve files across restore and replay.
  * Added support for forwarding custom Docker build arguments.

* **Bug Fixes**
  * Improved Docker image setup, including Playwright, Git, and compatible Python environment configuration.
  * Strengthened workspace isolation and protection of private task data.


docker/benchmark.Dockerfile
```diff
@@ -1,16 +1,18 @@
...
 RUN apt-get update \
     && apt-get install -y --no-install-recommends git \
     && rm -rf /var/lib/apt/lists/*
 
 WORKDIR /opt/corral
 COPY . /opt/corral
-RUN python -m pip install --no-cache-dir -e . \
-    && python -m corral.runtime.permissions --prepare-image
+ARG CORRAL_EXTRAS=""
+ARG CORRAL_TASK="samplemath"
```
n0w0f, line 14: Are those intentional ? why do we need to set those arg ?

src/corral/cli.py
```diff
@@ -402,27 +401,43 @@ async def run_benchmark(
...
+    canonical_agent = normalise_agent_name(agent_name or args.agent)
+    runtime_options = {
+        **dict(getattr(args, "agent_kwargs", {}) or {}),
+        **dict(agent_kwargs or {}),
+    }
+    harness = canonical_agent
+    if harness == "reflexion":
+        harness = normalise_agent_name(runtime_options.get("actor", "tool-calling"))
+    extra = AGENT_DEFINITIONS[harness].extra or ""
+    image_kind = "wetlab" if args.environment == "wetlab" else "benchmark"
```
n0w0f, line 413: why hardcode wetlab here ? all other environment would with benchmark ?

n0w0f (approved):

docker/benchmark.Dockerfile
MrtinoRG: The argument will come from the runner so this is just the default

src/corral/cli.py
MrtinoRG: Wetlab requires a different dockerfile because it requires conda. Since Wetlab it is the only env that runs with conda I called the dockerfile "wetlab" but we can also change the name to "conda" for example

n0w0f: oh yes! now i remember

PR #397, opened 2026-09-13
Revised CORRAL MD environment

### LINKED ISSUE #399: Ensure MD follows permissions
Since the MD env relies heavily on the Modal infra, it is necessary that while using it, the permissions and the sync correctly happen such that the files are synced into the `/workspace` and the agents can inspect them

chandan21gupta:
## Summary by Sourcery

Revise the CORRAL MD environment with expanded tasks, new scientific scoring capabilities, and GPU-enabled Python simulation support.

New Features:
- Expand CORRAL MD task definitions across both levels with new simulation, phonon, trajectory, spectrum, and model-evaluation tasks.
- Add optional remote GPU execution for Python workflows, including MACE/ASE simulations and workspace synchronization through Modal.
- Introduce scoring support for phonon coefficients, trajectory temperatures, spectral similarity, and regression-model R² validation.

Enhancements:
- Revise CORRAL MD prompts and execution guidance to select appropriate simulation engines and resource usage for different potentials and workloads.

Build:
- Update project lockfiles and add scientific and machine-learning dependencies required by the revised environment.

Deployment:
- Extend the Modal LAMMPS image and workspace integration to support GPU-backed Python execution.

Chores:
- Refresh level 1 and level 2 task configurations and add corresponding ground-truth datasets.

## Summary by CodeRabbit

* **New Features**
  * Added GPU-accelerated execution for Python simulation scripts, including MACE and ASE workflows.
  * Added workspace-aware result and diagnostic synchronization for remote runs.
  * Added materials-science benchmark tasks covering simulations, phonons, regression, and machine-learning workflows.
  * Added scoring support for trajectories, spectra, phonon results, and regression models.
  * Added secure asset preparation and model setup workflows.
* **Improvements**
  * Updated simulation guidance with absolute potential paths and engine-selection instructions.
  * Preserved local CPU execution for Python scripts.
  * Improved submission handling for numeric, file-based, and JSON results.
  * Expanded documentation for worker deployment and benchmark workflows.


tasks/corral_md/environments/level_1/tasks_json/task_10.json
```diff
@@ -0,0 +1,22 @@
...
...he MACE-MP-0 calculator and perform a constrained geometric relaxation of the atomic positions and unit cell. Allow the in-plane cell components to relax while keeping the out-of-plane cell dimension and cell components fixed, using ASE's ExpCellFilter with the mask [True, True, False, False, False, True]. Optimize with BFGS until the maximum force is below 0.005 eV/\u00c5. After relaxation, calculate the equilibrium in-plane cell area as A_0 = |a_1 \u00d7 a_2|. Generate 11 biaxially strained configurations by scaling both relaxed in-plane lattice vectors by factors from 0.98 to 1.02 in uniform increments of 0.004. Scale the atomic coordinates proportionally in the xy-plane while keeping the 15 \u00c5 z-direction vacuum unchanged. Do not perform further relaxation of the strained configurations. For each strained configuration, calculate the total potential energy with MACE. Fit the resulting energy-area data to E(A) = aA\u00b2 + bA + c. Calculate the 2D in-plane stiffness as K_2D = 2aA_0. Convert the resulting value from eV/\u00c5\u00b2 to N/m using 1 eV/\u00c5\u00b2 = 16.0217662 N/m, and return K_2D as a single floating-point value in N/m. Use ASE, MACE, and the BFGS optimizer.",
```
n0w0f: This task seems to be of much higher complexity compared to Task 1 and 2 for instance

tasks/corral_md/environments/level_1/tasks_json/task_4.json
```diff
@@ -0,0 +1,22 @@
+[
+  {
+    "name": "Level_1 - Task_4",
+    "description": "Construct an initial state consisting of a 3 \u00d7 3 \u00d7 1 periodic Graphene sheet with a 10 \u00c5 vacuum layer. Place a Lithium (Li) adatom positioned vertically at z = max(z_carbon) + 1.8 \u00c5 and horizontally at the center of the cell (x = cell[0, 0] \u00d7 0.5, y = cell[1, 1] \u00d7 0.5) directly above a hollow site. Create a final state by displacing this Li adatom by the vector cell[0] / 3.0 from its initial position, with no change in its z coordinate, to reach the adjacent hollow site. Load the MACE-MP-0 foundation model as an ASE-compatible atomic calculator. Generate a linear interpolation of exactly 5 images (including endpoints) for a Nudged Elastic Band (NEB) path. Calculate the single-point potential energy of all 5 images using the MACE calculator without performing any structural optimization or relaxation. Return the resulting unrelaxed activation barrier (defined as the maximum energy across the path minus the initial state energy) as a float value in eV units.",
```
n0w0f: Return the resulting unrelaxed activation barrier (defined as the maximum energy across the path minus the initial state energy)

Do we need to define how to compute activation barrier etc. in the brackets ? Or is this because it is level 1 ?

tasks/corral_md/environments/level_1/tasks_json/task_5.json
```diff
@@ -0,0 +1,22 @@
+[
+  {
+    "name": "Level_1 - Task_5",
+    "description": "Construct a Platinum (Pt) (111) surface slab with a 2 \u00d7 2 surface cell, 4 atomic layers, and a 10 \u00c5 vacuum layer. Constrain the bottom 2 atomic layers of the slab so they are fixed during optimization. Add a Carbon Monoxide (CO) molecule oriented with the Carbon atom directly above a surface Pt atom (at the atop site), with its initial z-coordinate 1.8 \u00c5 above the z-coordinate of the top-layer Pt atoms. Load the MACE-MP-0 foundation model as an ASE-compatible atomic calculator. Perform geometry optimization using a quasi-Newton optimizer (e.g., BFGS) until the maximum force is below 0.05 eV/\u00c5 for three configurations: the combined slab-CO system (with bottom slab layers constrained), the clean reference slab (with bottom slab layers constrained), and an isolated reference CO molecule centered in a simulation box providing 5 \u00c5 of vacuum padding on all sides.",
```
n0w0f: Perform geometry optimization using a quasi-Newton optimizer (e.g., BFGS)

I think here as well we should maybe say just quasi-Newton optimizer or BFGS. And the score should not depend upon the optimizer in former case.

n0w0f: (with bottom slab layers constrained) 
Do we repeat this because it is level 1 ? Because in the task description you already mention `Constrain the bottom 2 atomic layers of the slab so they are fixed during optimization.`

tasks/corral_md/environments/level_1/tasks_json/task_1.json
```diff
@@ -1,29 +1,22 @@
...
     ],
-    "scoring_function": "check_numerical",
-    "submission_format": "numerical value without units (eg. 56.00)",
-    "initial_input": {
-      "materials project id": "mp-149",
-      "atom style": "full",
-      "units": "metal",
-      "pbc": "true"
-    },
+    "scoring_function": "check_trajectory_temperature",
```
MrtinoRG, line 14: It seems that this scoring function and some of the ones present in the tasks right now lead to a value error because they are not in the list in the env. file

tasks/corral_md/src/corral_md/env.py
MrtinoRG, line 49: This dict should contain more functions

tasks/corral_md/environments/level_1/tasks_json/task_1.json
```diff
@@ -1,29 +1,22 @@
...
+    "name": "Level_1 - Task_1",
+    "description": "Load the MACE-MP-0 foundation model as an ASE-compatible atomic calculator. Construct a bulk Silver (Ag) conventional face-centered cubic (FCC) unit cell with a lattice parameter a = 4.09 \u00c5, and expand it into a 4 \u00d7 4 \u00d7 4 supercell to yield exactly 256 Ag atoms. Run a molecular dynamics (MD) simulation using a Langevin thermostat at a target temperature of 300 K with a timestep of 1.0 fs and a friction coefficient of 0.01 fs\u207b\u00b9. Execute the simulation for 5000 steps and write the system state to an ASE trajectory file every 100 steps. Return the absolute path to the resulting trajectory file. ",
     "tools": [
-      "get_potential_metadata",
       "get_structure_from_mp_text",
-      "convert_structure_to_lammps_data",
-      "run_lammps",
+      "run_python_script",
```
MrtinoRG: This tool name is not correct

tasks/corral_md/environments/level_1/tasks_json/task_10.json
```diff
@@ -0,0 +1,22 @@
...
+    "tools": [
+      "get_structure_from_mp_text",
+      "run_python_script",
+      "get_nth_run_log",
+      "keyword_log_extractor",
+      "visualisation_tool"
+    ],
+    "scoring_function": "check_numerical",
+    "submission_format": "numerical value without units (eg. 206.38)",
```
n0w0f:
```suggestion
    "submission_format": "numerical value without units (eg. 106.12)",
```
Do not give the target value in example submission_format

tasks/corral_md/environments/level_1/tasks_json/task_10.json
```diff
@@ -0,0 +1,22 @@
...
+    "initial_input": {},
+    "scoring_params": {
+      "target": 206.38,
+      "tolerance": 0
```
MrtinoRG: I think the tolerance should be set the same for all the tasks. Otherwise it is very inconsistent

tasks/corral_md/environments/level_1/tasks_json/task_4.json
```diff
@@ -0,0 +1,22 @@
...
+    "tools": [
+      "get_structure_from_mp_text",
+      "run_python_script",
+      "get_nth_run_log",
+      "keyword_log_extractor",
+      "visualisation_tool"
+    ],
+    "scoring_function": "check_numerical",
+    "submission_format": "numerical value without units (eg. 0.5049)",
```
n0w0f:
```suggestion
    "submission_format": "numerical value without units (eg. 1.5)",
```
Here again the target appear in the submission format

tasks/corral_md/environments/level_1/tasks_json/task_8.json
```diff
@@ -0,0 +1,22 @@
...
+      "get_nth_run_log",
+      "keyword_log_extractor",
+      "visualisation_tool"
+    ],
+    "scoring_function": "check_numerical",
+    "submission_format": "numerical value without units (eg. 1332.60)",
+    "initial_input": {},
+    "scoring_params": {
+      "target": 1332.5977,
+      "tolerance": 0.5
```
MrtinoRG: This tolerance is too big. 50% is too much

tasks/corral_md/environments/level_1/tasks_json/task_5.json
```diff
@@ -0,0 +1,22 @@
...
+    "tools": [
+      "get_structure_from_mp_text",
+      "run_python_script",
+      "get_nth_run_log",
+      "keyword_log_extractor",
+      "visualisation_tool"
+    ],
+    "scoring_function": "check_numerical",
+    "submission_format": "numerical value without units (eg. -1.7855)",
```
n0w0f:
```suggestion
    "submission_format": "numerical value without units (eg. -2.7)",
```

removed the target value in submission format.

tasks/corral_md/environments/level_1/tasks_json/task_5.json
MrtinoRG, line 1: Perhaps for this kind of tasks that require a scalar one could ask for the files as well and do some checks that the files were actually compiled such that the runs took place right?

tasks/corral_md/environments/level_1/tasks_json/task_6.json
```diff
@@ -0,0 +1,22 @@
...
... it into a 3 \u00d7 3 \u00d7 3 supercell. Construct a separate bulk conventional FCC unit cell of Nickel (Ni) with a lattice parameter a = 3.52 \u00c5. Load the MACE-MP-0 foundation model as an ASE-compatible atomic calculator and attach it to both structures. Perform a geometry optimization (e.g., using BFGS) on both the pure Cu supercell and the pure Ni unit cell until the maximum atomic force is below 0.00005 eV/\u00c5, and calculate their respective energies per atom. Next, identify the Cu atom closest to the geometric center of the Cu supercell and replace it with a Ni atom. Relax this defective supercell using the same optimization criteria. Calculate the impurity substitution energy using E_sub = E_defected - E_pure_Cu + E_Cu_bulk_per_atom - E_Ni_bulk_per_atom, where E_defected is the relaxed total energy of the Ni-substituted Cu supercell, E_pure_Cu is the relaxed total energy of the pure Cu 3 \u00d7 3 \u00d7 3 supercell, and E_Cu_bulk_per_atom and E_Ni_bulk_per_atom are the relaxed energies per atom of the corresponding bulk Cu and Ni unit cells. Return the impurity substitution energy as a floating-point value in eV units. USE ASE FOR THIS SIMULATION, DO NOT USE LAMMPS.",
```
n0w0f:
```suggestion
    "description": "Construct a bulk conventional face-centered cubic (FCC) unit cell of Copper (Cu) with a lattice parameter a = 3.615 \u00c5, and expand it into a 3 \u00d7 3 \u00d7 3 supercell. Construct a separate bulk conventional FCC unit cell of Nickel (Ni) with a lattice parameter a = 3.52 \u00c5. Load the MACE-MP-0 foundation model as an ASE-compatible atomic calculator and attach it to both structures. Perform a geometry optimization (e.g., using BFGS) on both the pure Cu supercell and the pure Ni unit cell until the maximum atomic force is below 0.00005 eV/\u00c5, and calculate their respective energies per atom. Next, identify the Cu atom closest to the geometric center of the Cu supercell and replace it with a Ni atom. Relax this defective supercell using the same optimization criteria. Calculate the impurity substitution energy using E_sub = E_defected - E_pure_Cu + E_Cu_bulk_per_atom - E_Ni_bulk_per_atom, where E_defected is the relaxed total energy of the Ni-substituted Cu supercell, E_pure_Cu is the relaxed total energy of the pure Cu 3 \u00d7 3 \u00d7 3 supercell, and E_Cu_bulk_per_atom and E_Ni_bulk_per_atom are the relaxed energies per atom of the corresponding bulk Cu and Ni unit cells. Return the impurity substitution energy as a floating-point value in eV units. ",
```

This is some strong prompt optimization here

tasks/corral_md/environments/level_1/tasks_json/task_5.json
MrtinoRG: The task does not specify what it is asking. It simply asks for the 3 geometry optimizations, while the submission format asks for a floating-point number...

n0w0f (changes requested):

tasks/corral_md/environments/level_1/tasks_json/task_2.json
```diff
@@ -1,7 +1,7 @@
...
+    "name": "Level_1 - Task_2",
+    "description": "Calculate the surface energy of Aluminum for the surface whose normal is along the given direction by removing periodic boundary conditions in the given direction and computing the energy difference per unit area between the bulk and surface configurations. Use the Embedded Atom Method (EAM) potential and the following simulation parameters. The simulation should replicate the unit cell 5 times in each of the x, y, and z directions. Return the surface energy in eV/\u00c5\u00b2 units.",
```
MrtinoRG, line 4: Is it not required to have a potential file? I am unsure, but based on the old task I think so

tasks/corral_md/environments/level_1/tasks_json/task_8.json
```diff
@@ -0,0 +1,22 @@
...
+    "tools": [
+      "get_structure_from_mp_text",
+      "run_python_script",
+      "get_nth_run_log",
+      "keyword_log_extractor",
+      "visualisation_tool"
+    ],
+    "scoring_function": "check_numerical",
+    "submission_format": "numerical value without units (eg. 1332.60)",
```
n0w0f:
```suggestion
    "submission_format": "numerical value without units (eg. 1232.80)",
```

Removed the target value

tasks/corral_md/environments/level_1/tasks_json/task_9.json
```diff
@@ -0,0 +1,22 @@
+[
+  {
+    "name": "Level_1 - Task_9",
+    "description": "Load the MACE-MP-0 foundation model as an ASE-compatible atomic calculator. Construct a primitive unit cell of Silicon (Si) in the diamond crystal structure with a lattice parameter a = 5.43 \u00c5. Attach the MACE calculator to the system. First, perform a full geometry optimization of the atomic positions using a quasi-Newton optimizer (e.g., BFGS) until the maximum atomic force is below 0.01 eV/\u00c5. Next, perform a finite-displacement harmonic analysis on the relaxed crystal to evaluate the vibrational frequencies. Programmatically calculate and return the Zero-Point Energy (ZPE) of the system at the \u0393-point as a floating-point value in eV. (Ensure you clean up or overwrite any temporary displacement files generated during the finite-displacement analysis).",
```
n0w0f: (Ensure you clean up or overwrite any temporary displacement files generated during the finite-displacement analysis)

Why is this comment required here ?

tasks/corral_md/environments/level_1/tasks_json/task_9.json
```diff
@@ -0,0 +1,22 @@
...
+    "tools": [
+      "get_structure_from_mp_text",
+      "run_python_script",
+      "get_nth_run_log",
+      "keyword_log_extractor",
+      "visualisation_tool"
+    ],
+    "scoring_function": "check_numerical",
+    "submission_format": "numerical value without units (eg. 0.0714)",
```
n0w0f:
```suggestion
    "submission_format": "numerical value without units (eg. 0.012)",
```

tasks/corral_md/environments/level_2/tasks_json/task_10.json
```diff
@@ -0,0 +1,30 @@
...
+  {
+    "name": "Level_2 - Task_10",
+    "description": "Load the MACE-MP-0 foundation model as an ASE-compatible atomic calculator from the local path ./models/2023-12-03-mace-128-L1_epoch-199.model. Construct a bulk conventional face-centered cubic (FCC) unit cell of Aluminum (Al) with a lattice parameter a = 4.05 \u00c5, and expand it into a 3 \u00d7 3 \u00d7 3 supercell. Initialize atomic velocities for a target temperature of 300 K following a Maxwell-Boltzmann distribution. Set up a Langevin thermostat (NVT) simulation with a friction coefficient of 0.01 fs\u207b\u00b9 and a 1.0 fs timestep. Implement a control loop that increases the target temperature from 300 K to 900 K in increments of 100 K (evaluating 300, 400, 500, 600, 700, 800, and 900 K). For each temperature plateau, run a total of 500 MD steps: use the first 400 steps for thermal equilibration, then use the final 100 steps for production sampling, recording the potential energy at every step. Calculate the average potential energy over these final 100 steps for each temperature plateau. Return a Python dictionary mapping each temperature integer (in K) to its corresponding average potential energy (in eV).",
```
n0w0f: `./models/2023-12-03-mace-128-L1_epoch-199.model`

There is a hardcoded path here ? But this is not shipped with the environment

n0w0f (changes requested):

tasks/corral_md/environments/level_2/tasks_json/task_2.json
MrtinoRG, line 1: Since the scoring is based on a ground truth, shouldn't the prompt explicitly mention the method for selecting the intervals?

tasks/corral_md/environments/level_2/tasks_json/task_10.json
```diff
@@ -0,0 +1,30 @@
...
+    "tools": [
+      "get_structure_from_mp_text",
+      "run_python_script",
+      "get_nth_run_log",
+      "keyword_log_extractor",
+      "visualisation_tool"
+    ],
+    "scoring_function": "check_dict",
+    "submission_format": "Python dictionary (eg. {300: -398.81, 400: -397.74})",
```
n0w0f: I would remove the exact values from examples

tasks/corral_md/environments/level_1/tasks_json/task_4.json
```diff
@@ -0,0 +1,22 @@
...
+    "initial_input": {},
+    "scoring_params": {
+      "target": 0.5049,
+      "tolerance": 0.005
```
MrtinoRG: Since here it uses a model that might be stochastic, shouldn't the prompt include the seed with which these values were obtained?

There are several tasks like this, and I think without the seed being specified it is very unlikely that the model arrives at the same values as the ground truth

tasks/corral_md/environments/level_2/tasks_json/task_1.json
MrtinoRG, line 1: The level 2 tasks are quite different than level 1 so it is very difficult to clearly see how the difficulty increases and in which layers

tasks/corral_md/environments/level_2/tasks_json/task_5.json
```diff
@@ -0,0 +1,23 @@
...
+    "tools": [
+      "get_structure_from_mp_text",
+      "run_python_script",
+      "get_nth_run_log",
+      "keyword_log_extractor",
+      "visualisation_tool"
+    ],
+    "scoring_function": "check_numerical_array",
+    "submission_format": "Python list or tuple of two numerical values (eg. [-0.2672, -0.1108])",
```
n0w0f: "submission_format": "Python list or tuple of two numerical values (eg. [-0.2672, -0.1108])",

Again the answer is here in the example

tasks/corral_md/environments/level_2/tasks_json/task_8.json
```diff
@@ -0,0 +1,23 @@
...
...h a conventional cubic unit cell of diamond Silicon using a lattice parameter of a = 5.43 \u00c5 and expand it into a 2 \u00d7 2 \u00d7 2 supercell. Load the MACE-MP-0 foundation model from the local path ./models/2023-12-03-mace-128-L1_epoch-199.model as an ASE-compatible calculator. Generate a dataset of 100 perturbed configurations. For each configuration i, where i ranges from 0 to 99, copy the base supercell and apply random atomic displacements drawn from a normal distribution with a standard deviation of 0.01 + 0.001 \u00d7 i \u00c5. Evaluate the total potential energy of each configuration using the MACE calculator. Compute SOAP descriptors for all 100 configurations using the dscribe library with species=['Si'], r_cut=5.0 \u00c5, n_max=6, l_max=6, periodic=True via DSCRIBE api, and average='inner', using n_jobs=1. Use configurations 0\u201379 as the training set and configurations 80\u201399 as the held-out test set. Train a Ridge regression model with Scikit-learn api with alpha=1e-3 using the SOAP descriptors and corresponding MACE potential energies. Save the trained Ridge regression model in pickle format and return the absolute path to the saved model's pickle file.",
```
n0w0f: Do we need packages like `DSCRIBE` installed in the python tool environment ? current we do not have this correct ?

tasks/corral_md/environments/level_2/tasks_json/task_7.json
```diff
@@ -0,0 +1,24 @@
...
+    "tools": [
+      "get_structure_from_mp_text",
+      "run_python_script",
+      "get_nth_run_log",
+      "keyword_log_extractor",
+      "visualisation_tool"
+    ],
+    "scoring_function": "check_numerical",
+    "submission_format": "numerical value without units (eg. 9.56e-05)",
```
n0w0f: target value in example

tasks/corral_md/modal_app/lammps_app.py
```diff
@@ -34,13 +35,9 @@
...
-    )
-    .env(
-        {
-            "LAMMPS_POTENTIALS": (
-                "/potentials/EAM:/potentials/EAM_FS:/potentials/TERSOFF"
-            )
-        }
+        "mace-torch",
+        "scikit-learn",
+        "dscribe",
```
n0w0f: okay I see dscribe here.
So the SOAP is also via this ?

tasks/corral_md/modal_app/lammps_app.py
```diff
@@ -34,13 +35,9 @@
...
-        {
-            "LAMMPS_POTENTIALS": (
-                "/potentials/EAM:/potentials/EAM_FS:/potentials/TERSOFF"
-            )
-        }
+        "mace-torch",
+        "scikit-learn",
+        "dscribe", 
     )
     .run_commands(
```
n0w0f, line 33: Can't we have this container or image built and stored somewhere ?

tasks/corral_md/src/corral_md/modal_workspace.py
```diff
@@ -268,5 +231,96 @@ def run_lammps_in_modal(
...
+    script_file: str,
+    args: list[str] | None = None,
+    *,
+    volume: Any | None = None,
+    remote_function: Any | None = None,
+    job_id: str | None = None,
+) -> int:
+    """Run a Python script on Modal (GPU) and sync its workspace outputs back locally.
+
+    Same upload -> remote execute -> download -> atomic publish pattern as
```
n0w0f, line 245: This workspace sync setup seems to be different from what @MrtinoRG  had for the core codebase

n0w0f (changes requested):

kjappelbaum:
@claude review this please

kjappelbaum:
@claude review

tasks/corral_md/environments/level_1/tasks_json/task_4.json
chandan21gupta: Thanks for pointing this out. No we don't need to. I will remove this.

tasks/corral_md/environments/level_1/tasks_json/task_5.json
chandan21gupta: Yeah I will specify BFGS optimiser

chandan21gupta: I see. Yeah this task prompt is a bit rough. I shall update it.

tasks/corral_md/environments/level_1/tasks_json/task_10.json
chandan21gupta: Thanks, will remove them.

tasks/corral_md/environments/level_1/tasks_json/task_9.json
chandan21gupta: Not required. I shall remove this.

tasks/corral_md/environments/level_2/tasks_json/task_10.json
chandan21gupta: I was thinking of a design choice here. Now my tool is designed in such a way it doesn't need to be specified the path, it should be able to extrapolate the paths from existing tools themselves. I shall remove this from the prompt.

tasks/corral_md/modal_app/lammps_app.py
chandan21gupta: yes

tasks/corral_md/environments/level_2/tasks_json/task_8.json
chandan21gupta: Yes, for MD environment this would be needed.

tasks/corral_md/environments/level_2/tasks_json/task_1.json
chandan21gupta: The level 2 examples have more workflow steps. 

In level 1 we do at most simulations or some energy based calculations. 
In level 2, we increase workflow steps by additional classical model training, finetuning, phonon calculations, or more complex simulation workflows.

tasks/corral_md/environments/level_1/tasks_json/task_4.json
chandan21gupta: Yes, I will refactor the prompt and remove the seed info.

tasks/corral_md/environments/level_1/tasks_json/task_2.json
chandan21gupta: Its mentioned in the task prompt to use EAM potential file. 
For mace based, I specified the absolute path, but would update the prompt now so that agent itself extrapolates the path from potential file tool.

tasks/corral_md/environments/level_2/tasks_json/task_2.json
chandan21gupta: Can you explain please? 

Also this task is directly taken from previous benchmark without any changes, so it should be correct.

tasks/corral_md/environments/level_1/tasks_json/task_5.json
chandan21gupta: Yes, I would need to update it. Thanks

tasks/corral_md/environments/level_2/tasks_json/task_2.json
MrtinoRG: The point was that if you change the intervals, I think that the final prediction can change, so you would need to provide the intervals used to get the ground truth

tasks/corral_md/environments/level_1/tasks_json/task_2.json
MrtinoRG: True, my bad

chandan21gupta:
Have updated the prompts and missing scoring functions.

MrtinoRG:
I think it is ready. 

What I did: 

- Took the level 2 tasks that Chandan had and made from them level 1 after enhancing tasks 8 and 9, as Sajid suggested. 
- Made the scoring check everything that is specified in the prompt and left room for everything else. For this, each task will now have a submission example in the workspace so that we can check during scoring.
	- This is a bit different than the other envs, but I would not say it makes the env not comparable, as I think this env, because of how heavy it is in IO-related work, it is ok to have some given files. 
- Worked a bit on the infra to make the tools more robust to the latest changes in permissions. 
	- Also related to the infra, worked a bit so that the base image also defines a Modal volume that allows for a fast initialization of each task.
	- In this way, we avoid having to sync heavy files at the beginning, such as the models or the potentials
- For scoring now, some calculations are run; most of them are single-point, simple stuff. The exception is task 4, which has one Hessian calculation that I do not know how to simplify.
	- Also, on the scoring, the light checks are run first so we reject early instead of running everything.
- The workspace that the model sees is `/input/`, `/output`, `/potentials` and `models`
	- Each of these has different permissions, with the latter read-only. 
- There is a slightly heavy, or at least heavier than I would like, instruction on what the workspace looks like right now, for which I would like some thoughts about whether we should keep it, make it shorter, or even remove it to make it more comparable with other envs.

src/corral/core/task.py
```diff
@@ -82,6 +83,8 @@ class TaskDefinition:
     # Interactive benchmarks can score their committed trajectory. The final
```
n0w0f: Would they also see the score ?

tasks/corral_md/environments/level_1/tasks_json/task_1.json
```diff
@@ -1,7 +1,7 @@
 [
   {
-    "name": "Melting of Silicon - Level 1",
-    "description": "Calculate the average self diffusion coefficient of liquid silicon at 2500K using molecular dynamics simulations with LAMMPS. Start from a crystalline silicon structure, which is replicated three times in each direction. The Stillinger-Weber (SW) potential should be used to describe interatomic interactions. Perform the heating and equilibration under zero external isotropic pressure control using NPT ensemble. Heat the system from 300K to 2500K at a rate of 2.2 K/ps, then equilibrate the liquid at 2500K for 500 ps. Use the Einstein relation to compute the diffusion coefficient from the mean squared displacement (MSD) data collected during the equilibration phase. Report the diffusion coefficient in units of m\u00b2/s.",
+    "name": "Level_1 - Task_1",
```
n0w0f: There is no tool for repeating unitcell ?

tasks/corral_md/environments/level_1/tasks_json/task_1.json
```diff
@@ -11,19 +11,18 @@
       "keyword_log_extractor",
       "visualisation_tool"
     ],
-    "scoring_function": "check_numerical",
-    "submission_format": "numerical value without units (eg. 56.00)",
```
n0w0f: Is this the manifest file you are referring to [here](https://github.com/lamalab-org/corral/pull/397/changes#diff-7f554c3cdaefbd15fe9048449372922534266e16214bd440e1fbadd825eff216)

It might not be clear  `Level 2 parts empty` which fields correspond to which level. I also see that you have separate templates for each of the task. Cant we have only the fields that is required for this task there ?

tasks/corral_md/environments/level_1/tasks_json/task_1.json
```diff
@@ -11,19 +11,18 @@
...
+    "scoring_function": "check_level1_workflow",
+    "submission_format": "Submit the path to your completed manifest.json file. Use the blank example as a guide and fill in only the parts needed for this task. Leave the Level 2 parts empty. Paths inside the manifest may be relative to the manifest file.",
```
n0w0f: Is this the manifest file you are referring to [here](https://github.com/lamalab-org/corral/pull/397/changes#diff-7f554c3cdaefbd15fe9048449372922534266e16214bd440e1fbadd825eff216)


It might not be clear Level 2 parts empty which fields correspond to which level. I also see that you have separate templates for each of the task. Cant we have only the fields that is required for this task there ?

tasks/corral_md/environments/level_1/tasks_json/task_4.json
```diff
@@ -0,0 +1,23 @@
+[
+  {
+    "name": "Level_1 - Task_4",
+    "description": "Prepare the force constants for bulk palladium. Use the supplied MACE-MP-0 checkpoint, a one-atom primitive FCC cell with conventional lattice parameter 3.89 Å, and a 5 x 5 x 5 supercell. Keep the geometry fixed. Choose a suitable force-constant method and document its numerical settings and symmetry treatment. Save the evaluated structures with their energies and complete force arrays, the displacement mapping or analytical derivative data, and the force constants. Report basic symmetry and acoustic checks, and record the model identity.",
+    "tools": [
+      "get_potential_metadata",
+      "get_structure_from_mp_text",
```
n0w0f: I think they would need mp id for those tool.

tasks/corral_md/environments/level_1/tasks_json/task_6.json
```diff
@@ -0,0 +1,23 @@
+[
+  {
+    "name": "Level_1 - Task_6",
+    "description": "Prepare an equilibrated aluminum state near 300 K for a later vibration calculation. Build a periodic primitive FCC cell with conventional lattice parameter 4.05 Å and repeat it 4 x 4 x 4 to make 64 atoms. Use the supplied MACE-MP-0 model and keep the cell fixed. Initialize Maxwell-Boltzmann velocities at 300 K with a recorded random seed, then remove center-of-mass momentum once. Choose a suitable thermostat, timestep, run length, and sampling interval. Do not reinitialize or rescale the velocities during the run. Save the thermal trace, sampled trajectory, and restartable final state. Use the saved data to show that the state is ready for the later NVE run.",
```
n0w0f: `300 K with a recorded random seed`.
Those seeds seems to be not part of the manifest here and above

tasks/corral_md/environments/level_2/tasks_json/task_3.json
```diff
@@ -0,0 +1,26 @@
...
...ration. Clearly distinguish fitting diagnostics on training data from evidence of generalization. Save a loadable trained checkpoint and training history sufficient to establish that optimization occurred and to reproduce the reported diagnostics.\n\nBuild periodic FCC silver using a conventional four-atom cubic cell with lattice parameter 4.09 Å repeated 3 x 3 x 3, giving 108 atoms. Keep the cell fixed and use the fine-tuned model for a 5 ps simulation targeting 300 K. Initialize velocities at 300 K with a recorded random seed, remove center-of-mass momentum once, and do not reinitialize velocities during the run. Choose and justify the thermostat, integration timestep, and sampling. Quantify temperature behavior and assess physical stability using energies, forces, and structural evolution. Discuss the limitations of transferring a dimer-trained model to bulk silver; report instability or poor agreement honestly rather than assuming successful transfer.\n\nRecord which input checkpoints you used and where the trained model came from. For the MD run, save the cell, positions, momenta, time, potential energy, and forces often enough to repeat the temperature and stability checks.",
```
n0w0f: `/workspace/models,` why give the path here in task description unlike how you have done in other tasks ?

tasks/corral_md/environments/level_2/tasks_json/task_5.json
```diff
@@ -0,0 +1,23 @@
...
...25 pairs of isotropic_strain and uniaxial_strain drawn from [-0.02, -0.01, 0.00, 0.01, 0.02]. Construct each strained structure from the same reference: first scale the complete cell and positions by 1 + isotropic_strain, then apply diag(1, 1, 1 + uniaxial_strain) in Cartesian coordinates to every cell vector and atomic position. Do not relax the structures. For each pair, determine the largest finite, real, strictly positive Gamma-point phonon energy in eV.\n\nChoose a suitable force-constant method and its numerical settings. Document displacement sizes where applicable, symmetry treatment, acoustic-sum-rule treatment, and the conventions used to construct the mass-weighted dynamical matrix. Fit all 25 energies by unweighted ordinary least squares with an intercept: phonon_energy_eV = intercept_eV + isotropic_coefficient x isotropic_strain + uniaxial_coefficient x uniaxial_strain. Report both coefficients in eV per unit strain, the intercept in eV, and fit residuals.\n\nKeep the evaluated structures, raw forces or analytical derivative results, force constants, and all mode energies. Also record which strain pair produced each result, so the calculation and fit can be repeated.",
```
n0w0f: should we also be giving python tools here

tasks/corral_md/src/corral_md/workflow_scoring/task_1.py
```diff
@@ -0,0 +1,930 @@
...
+            any(
+                _close(actual, expected, rtol=0.002, atol=0.002)
+                for actual in logged.get(row.step, [])
+            )
+        ):
+            raise ValueError("Thermo table does not match raw LAMMPS log")
+    return table
+
+
+def _protocol(e, end_time_ps=1500):
```
n0w0f: In the task prompt it is not clear that the system need to be equilibrated until 1500ps. or am i missing something ?

tasks/corral_md/src/corral_md/workflow_scoring/task_1.py
```diff
@@ -0,0 +1,930 @@
...
+    # Valid upstream physical evidence is required for scientific analysis credit.
+    valid = geometry and raw and continuous
+    for name, points, check in (
+        ("msd_reconstruction", 12, _msd_check),
+        ("diffusion_estimator_and_units", 10, _estimate),
+        ("reported_diffusion", 8, _reported),
+        ("equilibrated_production", 9, _equilibrium),
+        ("liquid_and_diffusive_evidence", 5, _liquid),
+        ("diffusive_regime", 4, _diffusive),
+    ):
```
n0w0f: How did you assign points here ?
I am comparing against 1 for each or yes/no

tasks/corral_md/environments/level_1/tasks_json/task_6.json
n0w0f: Would it be important to say how long it needs to be equilibrated ?

tasks/corral_md/environments/level_1/tasks_json/task_7.json
```diff
@@ -0,0 +1,23 @@
+[
+  {
+    "name": "Level_1 - Task_7",
+    "description": "Measure the 300 K reference volume for the aluminum expansion study. Start from the supplied periodic 108-atom FCC aluminum cell and use the supplied MACE-MP-0 model. Initialize velocities once with a recorded random seed, remove center-of-mass momentum, and run isotropic NPT at 300 K and 1.01325 bar. Choose suitable numerical settings, equilibration time, production time, and sampling. Report the mean measured temperature, pressure, and volume from the production data. Save the thermal trace, production trajectory, and restartable final state so the heating sequence can continue without resetting positions or velocities.",
```
n0w0f: Are we supplying this  alumnium cell file ?

n0w0f: choose suitable numerical setting, sounds very open ended, how would we score a valid range ?

tasks/corral_md/src/corral_md/workflow_scoring/level1.py
```diff
@@ -0,0 +1,1111 @@
...
+            delta = atoms.positions[:, None, :] - ideal.positions[None, :, :]
+            length = ideal.cell.lengths()
+            delta -= np.rint(delta / length) * length
+            mapping = np.argmin(np.linalg.norm(delta, axis=2), axis=1)
+            if len(set(mapping)) != 64:
+                raise EvidenceError("Distortions do not preserve distinct ideal sites")
+            result.append(delta[np.arange(64), mapping] / scale)
+        return np.asarray(result)
+
+    def random_distribution():
```
n0w0f: seems like this might not guarantee every displacement came from an independent Gaussian sample ? but should be okay I guess

tasks/corral_md/environments/level_2/tasks_json/task_3.json
n0w0f: `Save a loadable trained checkpoint and training history sufficient to establish that optimization occurred and to reproduce the reported diagnostics.` 
The loadability part seems to be not scored in the scorer

n0w0f (approved):
Looks good to me. 
But i think we should broadly discuss about the rubric based scoring.

src/corral/core/task.py
MrtinoRG: No, the scoring happens after the submission

tasks/corral_md/environments/level_1/tasks_json/task_1.json
MrtinoRG: No, it is not needed. I think LAMMPS allows replication

tasks/corral_md/environments/level_1/tasks_json/task_1.json
MrtinoRG: I will correct this

tasks/corral_md/environments/level_1/tasks_json/task_4.json
MrtinoRG: So you would suggest removing this tool for this task?

tasks/corral_md/environments/level_1/tasks_json/task_6.json
MrtinoRG: Check tasks/corral_md/src/corral_md/submission_templates/task_6.json

tasks/corral_md/environments/level_2/tasks_json/task_3.json
MrtinoRG: I can remove it, but it does not matter much I would say

tasks/corral_md/environments/level_2/tasks_json/task_5.json
MrtinoRG: It is already there as all tasks get the python tools but perhaps we can add it to the list

tasks/corral_md/src/corral_md/workflow_scoring/task_1.py
MrtinoRG: The final scoring at the end is 0 or 1, so if a single of those checks fail, the task will get scored 0

tasks/corral_md/environments/level_1/tasks_json/task_6.json
MrtinoRG: The agent has to decide and report at the end the time

tasks/corral_md/environments/level_1/tasks_json/task_7.json
MrtinoRG: Good point

MrtinoRG: Again the agent can decide and must report the range back

PR #401, opened 2026-09-15
feat: add AFM 2.0 questions

imandal98:
## Summary

Add a separate `tasks/afm_2.0` environment with 20 acquisition benchmarks: ten single-image tasks at level 1 and ten ordered three-image tasks at level 2.

- Register an acquisition scorer that validates submitted artifacts, parameter records, roughness and friction calculations, and statistical summaries.
- Include the AFM tools, retrieval database, Windows helper scripts, and documentation for processing conventions and scoring limitations.
- Add hardware-independent regression tests using synthetic NID-reader outputs.

## Validation

- Passed both acquisition regression tests, covering all 20 environment definitions and invalid submissions.
- Passed Python syntax checks, JSON parsing, and `git diff --cached --check`.
- Full hardware integration was not run; it requires a Windows workstation with the Nanosurf instrument and vendor software.

## Summary by Sourcery

Add a standalone AFM 2.0 benchmark environment with acquisition tasks, instrument workflows, artifact-based scoring, and synthetic-data regression coverage.

New Features:
- Add 20 AFM acquisition benchmark tasks spanning single-image level 1 workflows and ordered three-image level 2 workflows.
- Provide AFM acquisition tooling for instrument control, document retrieval, scan optimization, grain targeting, and image analysis.

Enhancements:
- Add acquisition scoring that validates submitted NID artifacts, acquisition records, channel data, roughness and friction metrics, and multi-image statistical summaries.
- Document AFM processing conventions, scoring limitations, supported workstation requirements, and environment usage.

Build:
- Add a standalone AFM environment project configuration and Windows activation/report helper scripts.

Documentation:
- Document hardware, software, processing, scoring, and regression-test requirements for the AFM environment.

Tests:
- Add hardware-independent regression tests covering all environment definitions, valid acquisition reports, and invalid submissions.

## Summary by CodeRabbit

* **New Features**
  * Added a Windows AFM 2.0 benchmark environment for Nanosurf hardware.
  * Added 20 acquisition tasks: 10 single-image tasks and 10 ordered three-image tasks.
  * Added tools for AFM image analysis, grain detection, scan optimization, document retrieval, and code execution.
  * Added scoring that checks acquisition files, parameters, channels, and reported metrics.
  * Added setup documentation and Windows helper scripts.

* **Tests**
  * Added hardware-independent regression tests for acquisition scoring and validation.
  * Hardware integration test results are not available.


tasks/afm_2.0/src/score.py
```diff
@@ -0,0 +1,647 @@
...
+            images = submission['images']
+            if not isinstance(images, list) or len(images) != len(acquisition_sequence):
+                return 0.0
+            paths = [Path(item['path']).resolve() for item in images]
+            if len(set(paths)) != len(paths):
+                return 0.0
+            # Detect hard links as well as repeated paths/symlinks.
+            identities = [(p.stat().st_dev, p.stat().st_ino) for p in paths]
+            if len(set(identities)) != len(identities):
+                return 0.0
```
MrtinoRG, line 591: Could we strengthen this check to ensure these are actually different acquisitions? At the moment, copying the same `.nid` three times gives different inodes and would pass, I believe. I think we could hash the raw required channel arrays from each NID and reject exact duplicates. If the NID metadata contains an acquisition timestamp/ID, we could also require three unique ordered acquisition IDs in addition to the different hashing.

tasks/afm_2.0/src/score.py
```diff
@@ -0,0 +1,647 @@
...
+            measured = []
+            previous = [0, 0]
+            for image, expected in zip(images, acquisition_sequence):
+                for field, target in (
+                    ('sequence_number', expected['sequence_number']),
+                    ('recorded_params', expected['params']),
+                    ('physical_settings', expected['physical_settings']),
```
MrtinoRG, line 598: Could we also check these parameters directly from the NID metadata? 

Right now, we only compare the parameters reported by the agent with the expected ones, so the agent could report the correct values even if the image was acquired with different settings. If the NID stores these parameters, I think it would be better to read them from the file and check that they match both the expected values and what the agent reports.

So this would imply not only comparing with the expected values but also with the reported ones to see cheat potential

tasks/afm_2.0/src/tools.py
```diff
@@ -0,0 +1,735 @@
...
+        # Calculate RMS Roughness if requested
+        if calculate_rms_roughness:
+            z = data["Image"]["Forward"]["Z-Axis"]
+            z_mean = np.mean(z)
+            squared_differences = (z - z_mean) ** 2
+            total_sum = np.sum(squared_differences)
+            M, N = z.shape
+            rms_roughness = np.sqrt(total_sum / (M * N))
+            logger.info(f"RMS Roughness: {rms_roughness}")
+
```
MrtinoRG, line 667: Could we reuse exactly the same analysis functions here and in the scorer? Right now, `Image_Analyzer` and `score.py` calculate roughness/friction differently, so an agent following the provided tool can get a value that the scorer rejects. 

Is there a reason for this different scoring? 

If not, I would move plane leveling, `Ra/Rq` and friction calculation into shared functions in a different module and call them from both places.

tasks/afm_2.0/environments/level_2/task_7.json
```diff
@@ -0,0 +1,217 @@
...
+            "dgain": 0,
+            "times_per_line": 0.075,
+            "points_per_line": 256,
+            "lines_per_frame": 256,
+            "centre_x": 0,
+            "centre_y": 0,
+            "rotation": 0,
+            "image_height": 5000,
+            "image_width": 5000,
+            "mode": 2
```
MrtinoRG, line 28: I think that mode 2 is Static Force, while the task is about lateral force. Can you check? This would apply to tasks 7 to 10 in both levels

tasks/afm_2.0/src/env.py
```diff
@@ -0,0 +1,378 @@
...
+event(
+    "DEBUG",
+    "environment.configuration",
+    subsystem="runtime",
+    benchmark="afm",
+    llm_model=LLM_MODEL,
+)
+ENVIRONMENT = "enviroment"
+TASK_TYPE = "subtasks_1"  # "single_task" or "subtasks"
+BASE_WORK_DIR = rf"C:\Users\Admin\Desktop\corral\corral\tasks\afm\src\afm\{LLM_MODEL}\{ENVIRONMENT}\{TASK_TYPE}"
```
MrtinoRG, line 55: As the bot says, there are several paths that point to the afm task dir. But I would suggest that you move this one there and remove the old afm task. As a result, we would have only the afm task that would correspond to what you currently have here in afm_v2. Or do you have a reason to keep the old task as well? Note that if you want to consult anything, the git history contains the old tasks and the code.

tasks/afm_2.0/environments/level_1/task_2.json
```diff
@@ -0,0 +1,101 @@
...
+            "lines_per_frame": 256,
+            "centre_x": 0,
+            "centre_y": 0,
+            "rotation": 0,
+            "image_height": 5000,
+            "image_width": 5000,
+            "mode": 2
+          },
+          "physical_settings": {
+            "normal_deflection_setpoint_v": 0.1
```
MrtinoRG, line 31: Could you double-check what this setpoint represents and which units are actually used by the Nanosurf API? Here it is defined as a normal-deflection setpoint in volts, but depending on the operating mode the controller setpoint may use different units. It would be good to make this explicit in the task and, if possible, also check the value against the metadata stored in the NID file.


This also applies to the level 2 task 2

tasks/afm_2.0/environments/level_1/task_4.json
MrtinoRG, line 1: I would change the description here to “calculate mean roughness (Ra) and RMS roughness (Rq).”

MrtinoRG: Right now, only the latter is mentioned, while the schema includes both. It would simply be clearer in the prompt

tasks/afm_2.0/environments/level_1/task_5.json
MrtinoRG, line 1: The task explicitly asks for roughness at 1×1 µm and 2×2 µm scales, but neither the submission format nor the scorer represents those measurements. The scorer calculates one Ra/Rq pair over the entire 5×5 µm image after plane subtraction. Is this for a reason?

tasks/afm_2.0/environments/level_1/task_7.json
```diff
@@ -0,0 +1,109 @@
+[
+  {
+    "name": "AFM level 1 task 7: Load-Dependent Average Friction",
+    "description": "Acquire one contact-mode lateral-force image over a 5 µm × 5 µm area at 256 × 256 pixels. Use a time per line of 0.075 s, normal-force setpoint of 10 nN, and P/I/D gains of 100/50/0. Record the height, lateral trace, and lateral retrace channels. Calculate the average friction.",
```
MrtinoRG, line 4: Would you like to add the standard deviation in the prompts as well?

tasks/afm_2.0/environments/level_1/task_8.json
```diff
@@ -0,0 +1,109 @@
+[
+  {
+    "name": "AFM level 1 task 8: Scan-Time Effect On Average Friction",
+    "description": "Acquire one contact-mode lateral-force image over a 5 µm × 5 µm area at 256 × 256 pixels. Use a time per line of 0.075 s, normal-force setpoint of 10 nN, and P/I/D gains of 100/50/0. Calculate the average friction.",
```
MrtinoRG, line 4: Same as in the previous issue, would you like to mention the SD as well as the average friction?

tasks/afm_2.0/environments/level_2/task_5.json
```diff
@@ -0,0 +1,199 @@
+[
+  {
+    "name": "AFM level 2 task 5: Scale-Dependent Topographic Roughness",
+    "description": "Acquire exactly three tapping-mode topography images centered at the same coordinates using scan sizes of 10 µm × 10 µm, 5 µm × 5 µm, and 2 µm × 2 µm, in that order. Keep the resolution at 512 × 512 pixels, time per line at 0.075 s, P/I/D gains at 100/50/10, and amplitude setpoint at 70%. Before acquisition. Calculate the rms and mean roughness for every image.",
```
MrtinoRG, line 4: The schema asks for percent_change, but neither the description nor the schema says relative to what. 

Is this percentage clear to domain experts? If yes, no changes required, otherwise I would mention what it is about

tasks/afm_2.0/environments/level_2/task_8.json
```diff
@@ -0,0 +1,217 @@
+[
+  {
+    "name": "AFM level 2 task 8: Scan-Time Effect On Average Friction",
+    "description": "Acquire exactly three contact-mode lateral-force images at the same location using times per line of 0.1, 0.075, and 0.05 s, in that order. Keep the scan size at 5 µm × 5 µm, resolution at 256 × 256 pixels, normal-force setpoint at 10 nN, and P/I/D gains at 100/50/0. Calculate the average friction magnitude and spatial standard deviation for every image.",
```
MrtinoRG, line 4: The same percentage issue as in task 5, so the same applies. If it is something obvious, no changes are required.

tasks/afm_2.0/environments/level_2/task_9.json
```diff
@@ -0,0 +1,217 @@
+[
+  {
+    "name": "AFM level 2 task 9: Friction Repeatability And Surface-Check Measurement",
+    "description": "Acquire exactly three consecutive contact-mode lateral-force images of the same region without changing any acquisition parameter. Use a scan size of 5 µm × 5 µm, resolution of 256 × 256 pixels, time per line of 0.075 s, normal-force setpoint of 10 nN, and P/I/D gains of 100/50/0. For every acquisition, calculate average friction magnitude and its spatial standard deviation from the lateral channels, and calculate rms and average roughness from the height channel.",
```
MrtinoRG, line 4: Again do you want to mention the SD?

tasks/afm_2.0/environments/level_2/task_7.json
```diff
@@ -0,0 +1,217 @@
+[
+  {
+    "name": "AFM level 2 task 7: Load-Dependent Average Friction",
+    "description": "Acquire exactly three contact-mode lateral-force images at the same location using normal-force setpoints of 5, 10, and 20 nN, in that order. Keep the scan size at 5 µm × 5 µm, resolution at 256 × 256 pixels, time per line at 0.075 s, and P/I/D gains at 100/50/0. Calculate the average friction magnitude and spatial standard deviation for every image.",
```
MrtinoRG, line 4: Here the description never asks the agent to fit the friction versus normal load

tasks/afm_2.0/environments/level_2/task_10.json
```diff
@@ -0,0 +1,217 @@
+[
+  {
+    "name": "AFM level 2 task 10: Spatially Resolved Roughness And Friction",
+    "description": "Treat the initial scan center as the origin. Acquire exactly three contact-mode lateral-force images centered at (0 µm, 0 µm), (+3 µm, 0 µm), and (0 µm, +3 µm), in that order. The coordinates are absolute positions relative to the initial origin, and all scan areas are assumed to lie within the calibrated scanner range. Use a scan size of 2 µm × 2 µm, resolution of 256 × 256 pixels, time per line of 0.075 s, normal-force setpoint of 5 nN, and P/I/D gains of 100/50/0. Before each move, calculate the displacement from the current center; the expected displacement vectors are (0, 0), (+3 µm, 0), and (-3 µm, +3 µm). For every location, calculate mean and average friction.",
```
MrtinoRG, line 4: The description ends with "calculate mean and average friction," which is redundant and doesn't mention roughness or friction SD.

tasks/afm_2.0/environments/level_2/task_1.json
MrtinoRG, line 1: I think the increase in complexity from Level 1 could be stronger here. At the moment, Level 2 mainly asks the agent to repeat the acquisition three times while changing P and I, but it does not require any comparison of the resulting images. Could we add a metric that measures the effect of the feedback changes, for example the mismatch between forward and backward topography for each acquisition? This would make Level 2 a controlled experiment where the agent needs to acquire, analyze, and compare the scans rather than mainly repeat the Level 1 workflow three times.

tasks/afm_2.0/environments/level_2/task_2.json
MrtinoRG, line 1: Similar to task 1, I think the Level 2 increase here is currently mostly operational: the agent performs three acquisitions at different setpoints, but does not need to analyze how the measurements change. Could we require Ra/Rq for each acquisition and, for example, the relative change between the different setpoints? That would give a clearer progression from one correctly configured acquisition in Level 1 to a small controlled experiment and comparison in Level 2.

tasks/afm_2.0/environments/level_2/task_3.json
MrtinoRG, line 1: Could we also add an image comparison to the Level 2 version here? Right now, the main extra difficulty is acquiring three scans at 80%, 70%, and 60% amplitude setpoints. Requiring Ra/Rq for each scan and the change relative to the middle acquisition would make the task require both experimental control and analysis, and would give a clearer complexity increase over Level 1.

MrtinoRG (commented):
Hi @imandal98, 

Thanks a lot for pushing this. I really like how the level 2 adds on top of level 1. 

I have added some comments. Please let me know what do you think

tasks/afm_2.0/environments/level_1/task_1.json
```diff
@@ -0,0 +1,103 @@
...
+            "dgain": 10,
+            "times_per_line": 0.075,
+            "points_per_line": 512,
+            "lines_per_frame": 512,
+            "centre_x": 0,
+            "centre_y": 0,
+            "rotation": 0,
+            "image_height": 5000,
+            "image_width": 5000,
+            "mode": 4
```
n0w0f: While solving task these decision are taken by the agent ? the parameters above these

tasks/afm_2.0/environments/level_1/task_1.json
```diff
@@ -0,0 +1,103 @@
...
+          "nominal_duration_seconds": 38.4,
+          "nominal_sampling_interval_nm": 9.765625,
+          "required_channels": ["topography"]
+        }
+      ],
+      "question": 1,
+      "level": 1,
+      "tolerance": 0.05
+    },
+    "submission_format": "{\"images\": [{\"sequence_number\": 1, \"path\": \"<absolute raw .nid path>\", \"recorded_params\": {\"pgain\": 100, \"igain\": 50, \"dgain\": 10, \"times_per_line\": 0.075, \"points_per_line\": 512, \"lines_per_frame\": 512, \"centre_x\": 0, \"centre_y\": 0, \"rotation\": 0, \"image_height\": 5000, \"image_width\": 5000, \"mode\": 4}, \"physical_settings\": {\"amplitude_setpoint_percent\": 70, \"drive_amplitude_mv\": 200}, \"channels\": {\"topography\": \"Z-Axis\"}}]}",
```
n0w0f: I think the submission format here leaks solutions or decisions that the agent has to make ? Instead of the actual values can we use example values or just syntax ?
Because there is a tendency sometime to submit whatever is in the submisison_format

tasks/afm_2.0/environments/level_1/task_1.json
```diff
@@ -0,0 +1,103 @@
...
+        "dgain": 10,
+        "times_per_line": 0.075,
+        "points_per_line": 512,
+        "lines_per_frame": 512,
+        "centre_x": 0,
+        "centre_y": 0,
+        "rotation": 0,
+        "image_height": 5000,
+        "image_width": 5000,
+        "mode": 4
```
n0w0f: If this values are here why do you need them from agent via submission ? or am i missing something ?

tasks/afm_2.0/environments/level_1/task_1.json
```diff
@@ -0,0 +1,103 @@
...
+              "drive_amplitude_mv": 200
+            },
+            "nominal_duration_seconds": 38.4,
+            "nominal_sampling_interval_nm": 9.765625,
+            "required_channels": ["topography"]
+          }
+        ],
+        "expected_image_count": 1,
+        "nominal_total_duration_seconds": 38.4,
+        "scoring_scope": "Checks submitted per-image parameter records, distinct readable raw files, channel dimensions and recomputed metrics. Records are self-reported; independent acquisition history and calibration verification require instrument telemetry."
```
n0w0f: How is this used ? or is this for user ?

tasks/afm_2.0/environments/level_1/task_10.json
```diff
@@ -0,0 +1,109 @@
...
+            "lateral_trace",
+            "lateral_retrace"
+          ]
+        }
+      ],
+      "question": 10,
+      "level": 1,
+      "tolerance": 0.05
+    },
+    "submission_format": "{\"images\": [{\"sequence_number\": 1, \"path\": \"<absolute raw .nid path>\", \"recorded_params\": {\"pgain\": 100, \"igain\": 50, \"dgain\": 0, \"times_per_line\": 0.075, \"points_per_line\": 256, \"lines_per_frame\": 256, \"centre_x\": 0, \"centre_y\": 0, \"rotation\": 0, \"image_height\": 2000, \"image_width\": 2000, \"mode\": 2}, \"physical_settings\": {\"normal_force_setpoint_nn\": 5}, \"channels\": {\"topography\": \"Z-Axis\", \"lateral_trace\": \"<lateral channel name>\", \"lateral_retrace\": \"<lateral channel name>\"}, \"ra_nm\": \"<number>\", \"rq_nm\": \"<number>\", \"friction_mean_v\": \"<number>\", \"friction_std_v\": \"<number>\", \"retrace_reversed\": true, \"commanded_center_nm\": [0, 0], \"displacement_nm\": [0, 0]}]}",
```
n0w0f: Again do you need all these number to come via agent ?
some of them are already here as input to task. I think it is better to ask for only the minimal amount of information with which we can score.

n0w0f: Do you want to use the same or similar variable name you asked in the task description in the submission format also ?
`calculate rms and mean roughness from the height channel and average friction magnitude and spatial standard deviation from the lateral channels.`
It is not clear to me maybe not be clear to the model as well `\"ra_nm\": \"<number>\", \"rq_nm\": \"<number>\", \"friction_mean_v\": \"<number>\", \"friction_std_v\": \"<number>\", \"retrace_reversed\":` 
meaning of these variable names

tasks/afm_2.0/environments/level_1/task_3.json
```diff
@@ -0,0 +1,107 @@
...
+          "nominal_duration_seconds": 38.4,
+          "nominal_sampling_interval_nm": 9.765625,
+          "required_channels": [
+            "topography"
+          ]
+        }
+      ],
+      "question": 3,
+      "level": 1,
+      "tolerance": 0.05
```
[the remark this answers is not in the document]
n0w0f: @imandal98  this seems to be relevant, are those tasks duplicates ?

tasks/afm_2.0/environments/level_1/task_4.json
```diff
@@ -0,0 +1,103 @@
...
+          "nominal_duration_seconds": 38.4,
+          "nominal_sampling_interval_nm": 9.765625,
+          "required_channels": ["topography"]
+        }
+      ],
+      "question": 4,
+      "level": 1,
+      "tolerance": 0.05
+    },
+    "submission_format": "{\"images\": [{\"sequence_number\": 1, \"path\": \"<absolute raw .nid path>\", \"recorded_params\": {\"pgain\": 100, \"igain\": 50, \"dgain\": 10, \"times_per_line\": 0.075, \"points_per_line\": 512, \"lines_per_frame\": 512, \"centre_x\": 0, \"centre_y\": 0, \"rotation\": 0, \"image_height\": 5000, \"image_width\": 5000, \"mode\": 4}, \"physical_settings\": {\"amplitude_setpoint_percent\": 70, \"drive_amplitude_mv\": 200}, \"channels\": {\"topography\": \"Z-Axis\"}, \"ra_nm\": \"<number>\", \"rq_nm\": \"<number>\"}]}",
```
n0w0f: again here the submission variables could use similar names to that of the task description. There is also additional things requested here

tasks/afm_2.0/environments/level_1/task_5.json
```diff
@@ -0,0 +1,103 @@
+[
+  {
+    "name": "AFM level 1 task 5: Scale-Dependent Topographic Roughness",
+    "description": "Acquire one tapping-mode topography image over a 5 µm × 5 µm area at 512 × 512 pixels and a time per line of 0.075 s. Set P/I/D gains to 100/50/10, and amplitude setpoint to 70%. Process the image and measure the rms roughness at multiple scales (1 x 1 µm, 2 x 2 µm).",
```
n0w0f: bit confused on this one. do we need to run separate for 1x1 and 2x2 ? or is this just computation ?
Also i would remove the brackets around (1 x 1 µm, 2 x 2 µm)

tasks/afm_2.0/environments/level_1/task_6.json
```diff
@@ -0,0 +1,101 @@
+[
+  {
+    "name": "AFM level 1 task 6: Scan-Time Effect On Contact-Mode Roughness",
+    "description": "Acquire one contact-mode topography image over a 5 µm × 5 µm area at 256 × 256 pixels. Use a time per line of 0.075 s, normal-force setpoint of 10 nN, and P/I/D gains of 100/50/0. Before acquisition, calculate the nominal acquisition duration as the number of scan lines multiplied by the time per line; the expected nominal duration is 19.2 s, excluding overhead. Calculate rms and mean roughness.",
```
n0w0f:
```suggestion
    "description": "Acquire one contact-mode topography image over a 5 µm × 5 µm area at 256 × 256 pixels. Use a time per line of 0.075 s, normal-force setpoint of 10 nN, and P/I/D gains of 100/50/0. Before acquisition, calculate the nominal acquisition duration as the number of scan lines multiplied by the time per line. Calculate rms and mean roughness.",
```
Is this information required here ?

tasks/afm_2.0/environments/level_1/task_8.json
n0w0f: there is also lateral_trace also asked there , also in other tasks as well

tasks/afm_2.0/environments/level_2/task_1.json
```diff
@@ -0,0 +1,199 @@
...
... 512, \"lines_per_frame\": 512, \"centre_x\": 0, \"centre_y\": 0, \"rotation\": 0, \"image_height\": 5000, \"image_width\": 5000, \"mode\": 4}, \"physical_settings\": {\"amplitude_setpoint_percent\": 70, \"drive_amplitude_mv\": 200}, \"channels\": {\"topography\": \"Z-Axis\"}}, {\"sequence_number\": 2, \"path\": \"<absolute raw .nid path>\", \"recorded_params\": {\"pgain\": 150, \"igain\": 50, \"dgain\": 10, \"times_per_line\": 0.075, \"points_per_line\": 512, \"lines_per_frame\": 512, \"centre_x\": 0, \"centre_y\": 0, \"rotation\": 0, \"image_height\": 5000, \"image_width\": 5000, \"mode\": 4}, \"physical_settings\": {\"amplitude_setpoint_percent\": 70, \"drive_amplitude_mv\": 200}, \"channels\": {\"topography\": \"Z-Axis\"}}, {\"sequence_number\": 3, \"path\": \"<absolute raw .nid path>\", \"recorded_params\": {\"pgain\": 150, \"igain\": 100, \"dgain\": 10, \"times_per_line\": 0.075, \"points_per_line\": 512, \"lines_per_frame\": 512, \"centre_x\": 0, \"centre_y\": 0, \"rotation\": 0, \"image_height\": 5000, \"image_width\": 5000, \"mode\": 4}, \"physical_settings\": {\"amplitude_setpoint_percent\": 70, \"drive_amplitude_mv\": 200}, \"channels\": {\"topography\": \"Z-Axis\"}}]}",
```
n0w0f: There is many variables asked here, I think many of them need not come from angent submission

tasks/afm_2.0/environments/level_2/task_10.json
n0w0f: `Before each move, `

what is meant by each move is not very clear to me

tasks/afm_2.0/environments/level_2/task_6.json
```diff
@@ -0,0 +1,193 @@
+[
+  {
+    "name": "AFM level 2 task 6: Scan-Time Effect On Contact-Mode Roughness",
+    "description": "Acquire exactly three contact-mode topography images of the same 5 µm × 5 µm region using times per line of 0.1, 0.075, and 0.05 s, in that order. Keep the resolution at 256 × 256 pixels, normal-force setpoint at 10 nN, and P/I/D gains at 100/50/0. Before acquisition, calculate the nominal durations; the expected values are 25.6, 19.2, and 12.8 s, respectively, excluding overhead. Calculate rms and mean roughness for every image and report the percentage change of each metric relative to the 0.075 s-per-line result.",
```
n0w0f: `calculate the nominal durations; the expected values are 25.6, 19.2, and 12.8 s, respectively, excluding overhead.`

Why do we give those expected values here ?

tasks/afm_2.0/environments/level_2/task_9.json
```diff
@@ -0,0 +1,217 @@
...
...retrace\": \"<lateral channel name>\"}, \"ra_nm\": \"<number>\", \"rq_nm\": \"<number>\", \"friction_mean_v\": \"<number>\", \"friction_std_v\": \"<number>\", \"retrace_reversed\": true}, {\"sequence_number\": 3, \"path\": \"<absolute raw .nid path>\", \"recorded_params\": {\"pgain\": 100, \"igain\": 50, \"dgain\": 0, \"times_per_line\": 0.075, \"points_per_line\": 256, \"lines_per_frame\": 256, \"centre_x\": 0, \"centre_y\": 0, \"rotation\": 0, \"image_height\": 5000, \"image_width\": 5000, \"mode\": 2}, \"physical_settings\": {\"normal_force_setpoint_nn\": 10}, \"channels\": {\"topography\": \"Z-Axis\", \"lateral_trace\": \"<lateral channel name>\", \"lateral_retrace\": \"<lateral channel name>\"}, \"ra_nm\": \"<number>\", \"rq_nm\": \"<number>\", \"friction_mean_v\": \"<number>\", \"friction_std_v\": \"<number>\", \"retrace_reversed\": true}], \"summary\": {\"ra_nm\": {\"mean\": \"<number>\", \"sample_std\": \"<number>\", \"cv_percent\": \"<number or null>\"}, \"rq_nm\": {\"mean\": \"<number>\", \"sample_std\": \"<number>\", \"cv_percent\": \"<number or null>\"}, \"friction_mean_v\": {\"mean\": \"<number>\", \"sample_std\": \"<number>\", \"cv_percent\": \"<number or null>\"}}}",
```
n0w0f: what is `cv_percent` ?

you do not ask for this in task prompt

tasks/afm_2.0/src/aila_image_process.py
```diff
@@ -0,0 +1,239 @@
...
+    y1 = Ymin + minr * y_scale
+    x2 = Xmin + maxc * x_scale
+    y2 = Ymin + maxr * y_scale
+
+    W1, H1 = x2 - x1, y2 - y1
+    xc_cur, yc_cur = (x1 + x2) / 2, (y1 + y2) / 2
+
+    # offset from current scan's center (in the scan's own LL origin coords)
+    delta_x = xc_cur - (W0 / 2)
+    delta_y = yc_cur - (H0 / 2)
```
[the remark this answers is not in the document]
n0w0f: Do we have cases where image extent does not start at (0, 0) ?

tasks/afm_2.0/src/aila_image_process.py
```diff
@@ -0,0 +1,239 @@
...
+
+    # Convert Z values to nanometers
+    Z = data["Image"]["Forward"]["Z-Axis"] * 1e9
+
+    # Initial flattening
+    Z_flat = flatten(Z, order=1)
+
+    # Extract spatial extents in micrometers
+    param = afm.param
+    extents = [param[i][j][0] * 1e6 for i in ["X", "Y"] for j in ["min", "range"]]
```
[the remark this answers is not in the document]
n0w0f: @imandal98 this is probably relevant

tasks/afm_2.0/src/env.py
n0w0f: maybe expose as cli ?

tasks/afm_2.0/src/env.py
```diff
@@ -0,0 +1,378 @@
...
+
+        prompt += "\nAvailable input data:\n"
+
+        # Display resolved inputs from dependencies
+        resolved = self.resolve_inputs(state)
+        for input_name, ref in self.current_task.input_map.items():
+            prompt += f"- {input_name} (from {ref.task_id}): {resolved[input_name]}\n"
+
+        # Display initial input data
+        for key, value in self.current_task.initial_input.items():
```
n0w0f: I think if you give all the input information like this it might not be required to give those in the task description. Right now you have those details in both places

tasks/afm_2.0/src/score.py
```diff
@@ -0,0 +1,647 @@
...
+            if isinstance(result, str):
+                result = result.strip()
+                try:
+                    parsed_result = json.loads(result)
+                    if isinstance(parsed_result, dict) and "answer" in parsed_result:
+                        answer = float(parsed_result["answer"])
+                    else:
+                        answer = float(result)
+                except json.JSONDecodeError:
+                    answer = float(result)
```
n0w0f: if there is json error wont `float(result )` break here ?

tasks/afm_2.0/src/score.py
```diff
@@ -0,0 +1,647 @@
...
+        try:
+            # Fix common LLM formatting mistakes
+            cleaned = result.replace(";", ",").strip()
+
+            # Ensure JSON braces are balanced
+            if not cleaned.startswith("{"):
+                cleaned = "{" + cleaned
+            if not cleaned.endswith("}"):
+                cleaned = cleaned + "}"
+
```
n0w0f: This cleaning is only in some scoring functions ? and why only on those ?

tasks/afm_2.0/src/score.py
```diff
@@ -0,0 +1,647 @@
...
+        except Exception as e:
+            logger.warning(f"Error: {e}")
+            return 0
+
+    return score_fn
+
+
+def auto_match_unit(rms_meters: float, llm_value: float) -> float:
+    """
+    Scale rms_meters to match the order-of-magnitude of llm_value.
```
n0w0f: mm, I am not sure if we should make this automatic scaling . It could be a genuine mistake by the agetn as well. Also you now give units in task definition

tasks/afm_2.0/src/score.py
```diff
@@ -0,0 +1,647 @@
...
+            logger.info(f"Raw submission {result}")
+            logger.info(f"Parsed data: {data['equation']}, {data['Rb']}, {data['A']}")
+            # Fit power law
+            C, k = fit_power_law(data["A"], data["Rb"])
+            fitted_eq = f"Rb = {C:.4f} * A**{k:.4f}"
+            logger.info(f"Fitted power law: {fitted_eq}")
+            match = re.search(
+                r"Rb\s*=\s*(?:([0-9.]+)\s*\*\s*)?A\s*\*\*\s*([0-9.]+)", data["equation"]
+            )
+            if match:
```
n0w0f: We need test for this probably to make sure this regex is robust

tasks/afm_2.0/pyproject.toml
```diff
@@ -0,0 +1,173 @@
...
+    "Intended Audience :: Science/Research",
+    "Topic :: Scientific/Engineering",
+    "Operating System :: Microsoft :: Windows",
+    "Operating System :: Unix",
+    "Operating System :: MacOS",
+]
+requires-python = ">=3.11"
+dependencies = [
+    # Removing the external GitHub dependency
+    # "corral@git+https://github.com/lamalab-org/mat-agent-bench.git",
```
n0w0f: why is this commented ?

n0w0f (changes requested):

n0w0f:
@imandal98 and @sajidmannan 
These are are some summary points from review, 

- Right now we ask for many parameters in submission format, which i think need not come from the agent, those are anyway present in initial input
- The variable names in submissions are different from the task description
- All the variables neeeded in submission are sometime not mentioned in task prompt, for example often you request for std deviation in submission but not mentioned in prompt and vice versa.
- In some scoring functions there is cleaning of jsons, but in some this is not present.
- there is some function for autoscaling of units, do we need that, in task jsons we have units.

tasks/afm_2.0/environments/level_1/task_10.json
MrtinoRG: For this, perhaps one can do as in MD and ask the agent to fill a JSON file instead of answering with the JSON itself

tasks/afm_2.0/environments/level_1/tasks_json/task_4.json
```diff
@@ -0,0 +1,50 @@
+[
+  {
+    "name": "AFM level 1 task 4: Tapping-Mode Roughness Repeatability",
+    "description": "Acquire one tapping-mode topography image over a $\\pu{5 µm} \\times \\pu{5 µm}$ area at $512 \\times 512$ pixels and a time per line of 0.075 s. Set P/I/D gains to 100/50/10, and amplitude setpoint to 70%. Process the image and measure the rms roughness. Report all roughness values in nanometres ($\\pu{nm}$).",
```
MrtinoRG: You request nm as units here, but I think in the scoring you are comparing them with meters without doing the conversion.

MrtinoRG: And this applies also to other tasks in this level

tasks/afm_2.0/environments/level_1/tasks_json/task_10.json
```diff
@@ -0,0 +1,52 @@
+[
+  {
+    "name": "AFM level 1 task 10: Spatially Resolved Roughness And Friction",
+    "description": "Treat the initial scan center as ($\\pu{0 µm}$, $\\pu{0 µm}$). Acquire one contact-mode lateral-force image centered at the origin using a scan size of $\\pu{2 µm} \\times \\pu{2 µm}$, resolution of $256 \\times 256$ pixels, time per line of 0.075 s, setpoint of 0.1 V, and P/I/D gains of 100/50/0. From this single acquisition, calculate rms and mean roughness from the height channel and average friction magnitude. Report all roughness values in nanometres ($\\pu{nm}$). Report all friction-signal values, in volts ($\\pu{V}$) from the lateral channel.",
```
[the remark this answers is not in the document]
MrtinoRG: @imandal98 I think that the bot is fully correct here

tasks/afm_2.0/src/env.py
```diff
@@ -0,0 +1,398 @@
...
+    "check_file_exists": check_file_exists,
+    "check_roughness_function": check_roughness_function,
+    "score_single_topography": score_single_topography,
+    "score_single_rms_roughness": score_single_rms_roughness,
+    "score_single_mean_roughness": score_single_mean_roughness,
+    "score_single_topography_roughness": score_single_topography_roughness,
+    "score_single_average_friction": score_single_average_friction,
+    "score_single_rms_friction": score_single_rms_friction,
+    "score_single_lateral_roughness": score_single_lateral_roughness,
+    "score_single_roughness_and_friction": score_single_roughness_and_friction,
```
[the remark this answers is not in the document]
MrtinoRG: @imandal98 I think this is also a valid concern. Right now, it is not possible to run these tasks

tasks/afm_2.0/src/env.py
```diff
@@ -0,0 +1,398 @@
...
+LLM_MODEL = os.environ.get("LLM_MODEL", "gpt_4o").strip()
+event(
+    "DEBUG",
+    "environment.configuration",
+    subsystem="runtime",
+    benchmark="afm",
+    llm_model=LLM_MODEL,
+)
+ENVIRONMENT = "enviroment"
+TASK_TYPE = "subtasks_1"  # "single_task" or "subtasks"
```
MrtinoRG: I would suggest removing the subtasks from the env.py files, as right now the env does not load because of this, and we are not going to need the subtasks anymore.

MrtinoRG: I mean it does not load because it defaults to subtasks and the subtask path is not valid anymore :)

tasks/afm_2.0/src/score.py
```diff
@@ -0,0 +1,668 @@
...
+                friction = _friction_signal(image, shape)
+                measured["average_friction"] = float(np.mean(friction))
+                measured["rms_friction"] = float(np.sqrt(np.mean(friction**2)))
+
+            for metric, answer in submitted.items():
+                expected = _finite_number(measured[metric])
+                # A relative comparison keeps small, unconverted signals meaningful.
+                if not math.isclose(answer, expected, rel_tol=tolerance, abs_tol=0.0):
+                    logger.warning(f"{metric}: submitted {answer}, measured {expected}")
+                    return 0.0
```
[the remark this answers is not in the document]
MrtinoRG: Yes, this is the issue I was pointing to in th task

tasks/afm_2.0/src/score.py
```diff
@@ -0,0 +1,668 @@
...
+            return float(check_params(final_params))
```
MrtinoRG: This will only work if the agent did not change the params between images, right? Why do you need to compare with the system anyway?

MrtinoRG: I think that if you compare it with what you get from `_single_image_scorer`, it is good enough. Like comparing the `final_params` with what you get from `_single_image_scorer`. Perhaps I am missing something, but to me this is how I would do it

MrtinoRG (commented):
I think the changes needed right now are very minimal. If you need help implementing them I can also help. But the env looks really good , I really like the separation between envs. Thanks for adding it!

PR #402, opened 2026-09-17
feat: rm retired environments

n0w0f:
Removed old ml and catalyst environments


MrtinoRG (commented):
I am not sure what models we are planning to run but perhaps these envs would still produce some signal on weaker models so I would not remove as also they do not hurt

n0w0f:
> I am not sure what models we are planning to run but perhaps these envs would still produce some signal on weaker models so I would not remove as also they do not hurt

do we plan to also maintain them ?

MrtinoRG:
Is it much work expected? I think both are simple envs that should not create much overhead

MrtinoRG (approved):

PR #408, opened 2026-09-21
feat: add common tool execution to common working path plus stronger permission control

### LINKED ISSUE #406: Route all the tool calls through a common method to strengthen permissions plus workspace conventions across envs

MrtinoRG:


MrtinoRG:
- What does this PR address? 

In the current `main` branch, there are several issues with how the permissions and tools are called. First, there are different routes that the tool execution can take: `environment.submit_job` is the default for the tools that can run in the background, while `execute_action` works for the others. The main issue about this is related to the permissions as the routes are different, the permissions can be different.

By looking at the code, I thought that the wetlab implementation was very clean, as it routes all tool execution through a single, centralized, and controlled route, ensuring the permissions are consistent across tools. Something that later I found is that for wetlab tools this was relatively straightforward as they all are very controlled tools while for the other envs tools this is not the case.

Also, in the code in `main`, the different envs could write to different paths in the filesystem. Mostly, it will depend on the tool and agent where to write, which is not ideal, I would say. I thought that perhaps it would be good to have the convention that all the environments write to the same path: `/workspace`.

- Solutions that I implemented.

For the tool permissions and route, this PR adds a centralized route through a tool executor that mostly ensures that the tool execution is "prepared". This validates the permissions and ensures that the tool does not bypass the permissions that were established before execution. As a result of this, the tools need to define their permissions when created (none, read, or read_write).

This adds a bit of redundancy, especially with the `execute_action` route, but I would say that this is better than having permission inconsistencies and not following convention. Also, with redundancy, we also make the permissions more explicit and controlled, which is good. 

For the filesystem, the PR simply enforces the `/workspace` convention for all the environments. Something that we should discuss is if we want to have some central prompt detailing the workspace structure and other possible conventions from the framework. I am not sure if this is necessary at this point, as the environment that is heavier on IO stuff is MD, but perhaps in the future we will have other envs for which this would apply as well. This will also help with the checkpointing and restoring of the workspace.

- Where to start the review

	- First of all, while this PR was quite big already, it seems that some pre-commit hooks applied to other files, making it bigger. Sorry for that.
	- Then I would suggest that one would start in the `/runtime/tool_execution.py` as it contains the layer that was added to route all tool executions and validate the permissions. But then, in that same dir, the worker files also contain some modifications not only about the tool permissions but also related to the workspace handling.
	- `core/tool.py` and `core/tool_catalog.py` define the permissions, so they are also important.
	- Obviously, `runtime/permissions.py` as it contains a lot of changes around the permissions system.
	- And then `workspace/filesystem.py` and `core/environment.py`, as they contain the machinery related to making `workspace` the default location. 
	- It would also be nice to have a second view on the permissions assigned to the tools of spectra, retro, wetlab, and stargazer.

